PR #4, opened 2019-08-01
PLANET-3759 - Convert Cookies block to Gutenberg

comzeradd:
This also contains some minor cleanup fixes:
* Added .git extension to submodule. It was removed at c207eb275813b7f48fc7d57d1fed74abeaef8946 but without it the submodule didn't work for me.
* Add `parserOptions` on eslint, so it doesn't complain for import/export.
* Added production mode to build command, since we practically only use this for production. We have the `start` command if we don't want minification.
* Removed unnecessary `.gitignore` files from a couple of folders.


kirdia:
Looks good!
It needs this functionality though for the frontend.
[https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/admin/js/blocks/CookiesBlock.js](url)

comzeradd:
@kirdia PR amended.

classes/blocks/class-cookies.php
```diff
@@ -0,0 +1,123 @@
...
+				$post = get_post( $post_id );
+			}
+		} else {
+			$post = get_queried_object();
+		}
+
+		// Enqueue js for the frontend.
+		if ( ! $this->is_rest_request() ) {
+			wp_enqueue_script( 'cookies', P4GBKS_PLUGIN_URL . 'public/js/cookies.js', [ 'jquery' ], '0.1', true );
+			wp_localize_script( 'cookies', 'cookies', $menu );
```
kirdia: @comzeradd please remove this, is unneeded

comzeradd:
@kirdia thanks! PR fixed

react-blocks/src/blocks/Cookies/CookiesBlock.js
```diff
@@ -0,0 +1,131 @@
...
+          type: 'string',
+        },
+        all_cookies_name: {
+          type: 'string',
+        },
+        all_cookies_description: {
+          type: 'string',
+        },
+      },
+      edit: withSelect((select) => {
```
sagarsdeshmukh: As we are not using `withSelect` HOC here, we could remove it.

comzeradd:
@sagarsdeshmukh PR updated

PR #10, opened 2019-08-06
PLANET-3815: Show empty message if the block has no content.

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3815

Notes:
As @kirdia  mentioned, this should happen "natively" (so to speak) via this Wordpress code:

https://github.com/WordPress/gutenberg/blob/master/packages/server-side-render/src/server-side-render.js#L87-L93

However, I could never get that condition to match. No matter what I returned from the block. So for the moment, this is just an style improvement over the previous message, to show something nicer, and supporting translations and custom per-block messages.


react-blocks/src/blocks/Submenu/Submenu.js
```diff
@@ -105,7 +105,7 @@ export class Submenu extends Component {
             attributes={{
               submenu_style: this.props.submenu_style,
               title: this.props.title,
-              levels: this.props.levels,
+              levels: this.props.levels
```
pablocubico: Oops, this wasn't intended, the trailing comma is supposed to be better for version control, but I just hate seeing it!

react-blocks/build/editorStyle.min.css
```diff
@@ -1 +1,59 @@
-.submenu-block-attribute-wrapper,.submenu-level-link{display:inline-block;margin-right:20px}.submenu-level-link .components-base-control__field{margin-bottom:13px}.LayoutSelector{display:flex;width:100%;margin-bottom:30px;cursor:pointer}.LayoutSelector .LayoutOption{padding:20px;border-right:1px solid #ccc;width:100%}.LayoutSelector .LayoutOption:last-child{border-right:0}.LayoutSelector .help{font-family:Roboto,sans-serif;font-size:10pt;margin-top:10px}.Preview .PreviewBar{border-bottom:1px solid #8bcc49;font-style:italic;padding:12px}.Preview .submenu-sidebar{float:unset!important}
\ No newline at end of file
+/**
```
comzeradd: These don't seem minified. `npm run build` should minify them.

PR #13, opened 2019-08-20
PLANET-3914 Convert Columns block to Gutenberg

sagarsdeshmukh:
[JIRA 3914](https://jira.greenpeace.org/browse/PLANET-3914)


react-blocks/src/blocks/Columns/Columns.js
```diff
@@ -0,0 +1,193 @@
...
+              </Tooltip>
+            )}
+          </div>
+        </Fragment>
+      );
+    }
+
+    render() {
+
+      // Temporary fix to refresh serversiderender component on columns fields changes.
```
sagarsdeshmukh: To fix the columns fields not updating onchange in `ServerSideRender ` component, I added below temporary fix.

classes/blocks/class-columns.php
```diff
@@ -0,0 +1,135 @@
...
+				list( $src ) = wp_get_attachment_image_src( $attributes[ "attachment_$i" ], $image_size );
+				if ( $src ) {
+					$attributes[ "attachment_$i" ] = $src;
+				}
+			}
+		}
+
+		$block_data = [
+			'fields'              => $attributes,
+			'available_languages' => P4BKS_LANGUAGES,
```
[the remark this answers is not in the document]
sagarsdeshmukh: It's `P4GBKS_LANGUAGES`. Good catch 👍

react-blocks/src/blocks/Columns/Columns.js
sagarsdeshmukh: Yes, The random number trick works. 👍

comzeradd: Wouldn't [componentDidUpdate](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/react-blocks/src/blocks/Timeline/Timeline.js#L21) also be used? Or am I missing something?

react-blocks/src/blocks/Columns/ColumnsBlock.js
```diff
@@ -0,0 +1,221 @@
...
+                  shortcode: ({named: {columns_block_style = ''}}) => columns_block_style,
+                },
+                columns_title: {
+                  type: 'string',
+                  shortcode: ({named: {columns_title = ''}}) => columns_title,
+                },
+                columns_description: {
+                  type: 'string',
+                  shortcode: ({named: {columns_description = ''}}) => columns_description,
+                },
```
kirdia: here missing transform functions for columns attribute

react-blocks/src/blocks/Columns/ColumnsBlock.js
```diff
@@ -0,0 +1,221 @@
...
+                }
+                setAttributes({columns: new_columns});
+              }
+            }
+          }
+
+          function addColumn() {
+            const {columns} = attributes;
+
+            if ( 4 > columns.length ) {
```
kirdia: this check should be reversed probably

react-blocks/src/blocks/Columns/ColumnsBlock.js
```diff
@@ -0,0 +1,221 @@
...
+                }]
+              });
+            }
+          }
+
+          function removeColumn() {
+            const {columns}  = attributes;
+            const column_count = columns.length - 1;
+
+            if (0 <= column_count) {
```
kirdia: also here should this check be reversed?

react-blocks/src/blocks/Columns/ColumnsBlock.js
```diff
@@ -0,0 +1,221 @@
...
+              setAttributes({
+                columns: [
+                  ...columns.slice(0, column_count),
+                  ...columns.slice(column_count + 1)
```
kirdia: you can use slice with a negative index here to remove the last item

react-blocks/src/blocks/Columns/Columns.js
kirdia: this is happening because changes on a sub-attribute of the columns attribute will not trigger component render in this case (but it works when a new element is removed/added to the array).
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/13/files#diff-2c222971ee3d2fcc0a3c7ce63a33916dR105

check here for a workaround. you'd have to clone the current array, set the sub-attribute and reset the columns attribute
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/react-blocks/src/blocks/Submenu/SubmenuBlock.js#L129

pablocubico: All of the options still feel like workarounds and "unreacty", I think we are missing something very simple here. I think we could revisit this after, maybe dispatching an action to WP's store... 🤔

react-blocks/src/blocks/Columns/Columns.js
```diff
@@ -0,0 +1,188 @@
...
+              placeholder={__('Enter block description', 'p4ge')}
+              value={columns_description}
+              onChange={this.props.onDescriptionChange}
+            />
+          </div>
+          <div>
+
+            {columns.map((item, index) => {
+              return (
+                <div>
```
pablocubico: Hey S! This `<div>` should include a `key` property as: `<div key={index}>`, to avoid a typical "no key in iteration" error, all iterations should include this `key` in the root element.

pablocubico:
@sagarsdeshmukh looks good! Just added a simple iterator issue. For the `tempField` thing, I think the best is @kirdia 's suggestion on deep cloning an object using `JSON.parse/stringify`, seems like an usual choice in the React universe... which is so full of quirks and antipatterns!

sagarsdeshmukh:
PR updated! please have a look 👍

PR #15, opened 2019-08-22
PLANET-3905 Register/Hide blocks per post type

dpivo:
- Filter added for blocks allowed in post types


comzeradd:
@dpivo something went wrong with rebasing. can you squash the commits to one?

PR #16, opened 2019-08-22
PLANET-3905 Register/Hide blocks per post type

dpivo:
- Filter allowed blocks per post type


comzeradd:
Check phpcs/tests for the warnings.

PR #22, opened 2019-08-29
Add Media block styles file

sagarsdeshmukh:
Added missing scss files of Media block to Gutenberg repo.


comzeradd:
I see you also added `SplitTwoColumns.scss` here. Is this intended?

sagarsdeshmukh:
> I see you also added `SplitTwoColumns.scss` here. Is this intended?

It was by mistake, thanks for flag it.

PR #25, opened 2019-08-30
PLANET-3929 Video: not aligned correctly on S and L screens

dpivo:
- Applied changes to make the video responsive and take up width of screen.


comzeradd:
Looks good :100: 

Gutenberg has its own Video Block (actually the ticket seems to refer to that one). Can you check if it's easy to enqueue the media script there too?

public/js/media.js
```diff
@@ -0,0 +1,10 @@
+jQuery(function ($) {
+
+  let $all_oembed_videos = $( 'iframe[src*="youtube"]' );
+
+  //Remove height and with from imbed iframes and wrap im container
```
comzeradd: you have a few typos here on the comment (width, embed, in)

PR #26, opened 2019-08-30
PLANET 3919: Implement 'transforms' for each new block

pablocubico:
WIP!

This includes 2 approaches:

**1.** Added a POC for converting the blocks using PHP only, as we do with the "convert old blocks to new blocks" feature. String replacement via `do_shortcode`. 

This is less aggressive to the post's content than using `rawHandler`. 

Converting a post with this method will produce a Gutenberg post with a mix of `Classic` blocks (basically a rich text editor) and P4 blocks.

**2.** For the `rawHandler` implementation, I continued @kirdia 's work on https://github.com/greenpeace/planet4-convert-to-gutenberg , and copied it to this repo so we can use the block definitions included here.

It uses a `wpAlias` script since there was two limits for using the WP CLI:
- I can't connect to the container DB [on a Mac](https://docs.docker.com/docker-for-mac/networking/) as easy as on Linux
- the `wpCli` node wrapper uses `child_process.spawn` internally which caused problems with symbolic links, shell scripts or apparently anything other than a binary causing `NOENT` errors, so I changed it to just using exec.

My `wpAlias` lives in `/usr/local/bin` and its content is:
```bash
cd /Users/pablocubico/dev/greenpeace/planet4-docker-compose
export COMPOSE_INTERACTIVE_NO_CLI=1
docker-compose -p planet4dockercompose exec php-fpm wp $@
```

I had to also reference `zsh` [in the script](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/26/files#diff-914b649674b03fe0315c3237a73fdd8dR262), if you're not using ZSH, replace that with your preferred shell.

You can run: 

`node BlockConversion.js --debug` 

This method will produce Gutenberg posts with many block types, not just `Classic`, it tries to convert HTML to its corresponding blocks like `Paragraph` for `<p>` and so on.

To test it without doing any real changes to the DB. It will convert the last 20 posts (the 20 is hardcoded..., you can change it in the code).

I separated the blocks config (see the new `...Config.js` files) to make it "cross-runtime" since we need them both in the browser and in the node environment.

I created a separate `node` target for Webpack to convert the script to ES6 and use the same configuration, but I'm not sure there is any value to it since this is only intended to be run on the first migration, we could leave it as it is (in which case, I'll roll back the Webpack changes and the `commandsIndex` files). So ignore that for now.

After doing this, I'm leaning towards the PHP-only conversion (see `class-shortcodetogutenberg.php`). I guess `rawHandler` is a bit buggy, even using the "Convert to Blocks" function in the editor. I'll check if there is a new version of `@wordpress/scripts` or any workarounds, but it fails to catch some shortcodes, for example I found this reproducible bug where it fails to convert the first block of some given content:

convert-blocks-bug

Next steps:

- Find a workaround for undetected shortcodes, for example, using the PHP conversion as a second pass after the `rawHandler` one. Or finding out if there's an updated for the `@wordpress/scripts` package (where rawHandler is included).
- Probably making the script fully synchronous (right now is an `async/await` soup  😛).
- Error handling, logging, and resume from last row converted?
- Documenting and preparing for running this on the servers.

Ref: https://jira.greenpeace.org/browse/PLANET-3919


pablocubico:
Closing as this is outdated.

PR #28, opened 2019-09-02
PLANET-3930: Submenu: long full width not aligned on S screens

pablocubico:
This replicates the same change made to the original blocks plugin.

Ref: https://jira.greenpeace.org/browse/PLANET-3930


kirdia:
Looks good! Needs rebasing

PR #30, opened 2019-09-06
PLANET-3983 Fix persistence of tags, post_types and posts tokens in covers block

kirdia:
Populate tags, post types and posts tokens on initial render so selected values of these attributes are visible to the editor.
Make posts selections dynamic based on block's attributes and input search.
Enhance covers transforms function to include all old block's attributes
Reset some block's attributes when cover_type is changed
Pass planet4_options to js using localize_script
Few minor changes in php to accomodate attribute type changes from old block.


pablocubico:
Hey @kirdia looks like the `LayoutSelector` is not highlighting the selected radio, because of the type issue it has (Number vs String vs etc...), can you check it? We removed the type cast it used to had because sometimes we use it with numbers and sometimes with strings... 😬

kirdia:
@pablocubico thanx for noticing. I updated the PR

react-blocks/src/blocks/Covers/Covers.js
```diff
@@ -104,44 +241,50 @@ export class Covers extends Component {
...
-            this.props.cover_type === 3
-            ? <FormTokenField
-              value={ this.state.postTypeTokens }
-              suggestions={ postTypeSuggestions }
-              label='Post Types'
-              onChange={ tokens => this.onSelectedPostTypesChange(tokens) }
-              placeholder="Select Tags"
-            />
-            : null
+            this.props.cover_type === 3 && this.props.posts.length === 0
```
pablocubico: Hey @kirdia looks like these conditions are false because `cover_type` is a String now, it `=== '3'`

classes/blocks/class-covers.php
```diff
@@ -188,7 +193,7 @@ private function filter_posts_by_ids( $fields ) {
 			];
 
 			// If cover type is take action pages set post_type to page.
-			if ( isset( $fields['cover_type'] ) && '1' === $fields['cover_type'] ) {
+			if ( isset( $fields['cover_type'] ) && 1 === $fields['cover_type'] ) {
```
pablocubico: note: this one looks suspicious, should it be a string too? 🤔

classes/blocks/class-covers.php
```diff
@@ -44,7 +44,6 @@ public function add_block_shortcode( $attributes, $content ) {
 		);
 
 		$attributes['tags']       = explode( ',', $attributes['tags'] );
```
comzeradd: Could you also adjust space here?
phpcs: `Equals sign not aligned correctly; expected 1 space but found 7 spaces`

PR #31, opened 2019-09-06
PLANET-4013 Use default articles title and read more button text values

kirdia:
Use default articles title and read more button text values for articles block if their values are empty.
Removed also setting the default values in register_block_type, because they add no value if wp_block_type is not used.
https://developer.wordpress.org/reference/classes/wp_block_type/

### DIFF
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -64,10 +64,6 @@ class Articles extends Base_Block {
 	public function __construct() {
 		add_shortcode( 'shortcake_articles', [ $this, 'add_block_shortcode' ] );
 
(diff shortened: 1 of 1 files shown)


comzeradd:
Seems to work fine. But if I fill-in the Title on the backend and then delete it, it doesn't rollback to the default option. Is this something we can fix?

kirdia:
Fixed conflict.
I did not understand that. Can you elaborate?
If title and planet4 setting are empty it should show 'Related Articles'
articles_title

kirdia:
Yeap, it does not work on the frontend. There are some changed attributes on the twig template

kirdia:
Updated PR, reused old fields names for title, description and count

PR #37, opened 2019-09-17
PLANET-4054: Gutenberg Compatibility: Separate Bootstrap, layout styles and block styles in master theme

pablocubico:
(WIP)

Ref: https://jira.greenpeace.org/browse/PLANET-4054


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "react-blocks/src/styles/styleguide"]
 	path = react-blocks/src/styles/styleguide
 	url = https://github.com/greenpeace/planet4-styleguide.git
+[submodule "assets/src/styles/styleguide"]
+	path = assets/src/styles/styleguide
+	url = https://github.com/greenpeace/planet4-styleguide
```
comzeradd: I guess we can remove the old submodule description here.

comzeradd:
Can you also fix [stylelint path](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/.circleci/config.yml#L50)?

comzeradd:
Tested it locally and it seems to work fine. Do you also want to squash some or all the commits before merge?

comzeradd:
Also if `style.css` on the root folder is not used, we can delete it.

comzeradd:
This looks good, but needs rebase. Not sure if you want to remove the lodash/babel fix after all, since this wasn't the root cause of the bug.

pablocubico:
@comzeradd rebased, lodash commit removed!

PR #42, opened 2019-09-24
PLANET-4014 Add load_more functionality to Articles block

comzeradd:


classes/blocks/class-articles.php
```diff
@@ -179,6 +181,12 @@ public function prepare_data( $fields ): array {
 			$recent_posts = $this->populate_post_items( $sliced_posts );
 		}
 
+		// Enqueue js for the frontend.
+		if ( ! $this->is_rest_request() ) {
+			wp_enqueue_script( 'load-more', P4GBKS_PLUGIN_URL . 'public/js/load_more.js', [ 'jquery' ], '0.3', true );
+			wp_localize_script( 'load-more', 'more_url', '/wp-admin/admin-ajax.php' );
```
sagarsdeshmukh: Here the 3rd argument of `wp_localize_script ()` function expects array.

sagarsdeshmukh:
Overall PR looks good to me 👍

classes/blocks/class-articles.php
comzeradd: @sagarsdeshmukh good catch! I amended the commit.

PR #44, opened 2019-09-27
PLANET-4138 Cover-Content block - Fix load more button issue on 1 raw…

sagarsdeshmukh:
… selected

[JIRA 4138](https://jira.greenpeace.org/browse/PLANET-4138)

### DIFF
--- a/templates/blocks/content_four_column.twig
+++ b/templates/blocks/content_four_column.twig
@@ -55,9 +55,9 @@
 
 
 				{% if ( posts|length <= 4
-					or ( posts|length > 4 and posts|length <= 8 and posts_view == '3' )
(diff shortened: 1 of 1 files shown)


templates/blocks/content_four_column.twig
```diff
@@ -57,7 +57,7 @@
 				{% if ( posts|length <= 4
 					or ( posts|length > 4 and posts|length <= 8 and posts_view == '3' )
```
kirdia: Looks good. can you change these checks also (for 2 rows)? they seems wrong
`posts|length > 4 and posts|length <= 8 and posts_view == '3' `

It should be posts_view=='2', since the values have different meaning between the old blocks and the new.

PR #45, opened 2019-10-01
Fix typo error and move Planet4 Blocks menu entry into 3rd row.

galousis:
When adding a new post and clicking on + icon to add blocks


galousis:
@comzeradd thanks for the tip on commit msgs

PR #47, opened 2019-10-01
[No ticket] - Minor block fixes

pablocubico:
I needed to change a few null coalescing operators (`??`) to simple ternaries (`? :`) because in the Shortcake versions of `prepare_data`, the unset `fields` were not present in the array, in the Gutenberg side, those are empty strings, which are taken as valid by `??`.

It has three commits, I could squash them but I already merged the first two into the PLANET-4054 branch to be able to test the blocks correctly 😬

### DIFF
--- a/classes/blocks/class-counter.php
+++ b/classes/blocks/class-counter.php
@@ -105,7 +105,7 @@ class Counter extends Base_Block {
 		}
 		$target = floatval( $fields['target'] );
 
-		if ( array_key_exists( 'completed_api', $fields ) ) {
+		if ( array_key_exists( 'completed_api', $fields ) && $fields['completed_api'] ) {
 			$response_api  = wp_safe_remote_get( $fields['completed_api'] );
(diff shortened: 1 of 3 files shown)


classes/blocks/class-counter.php
```diff
@@ -105,7 +105,7 @@ public function prepare_data( $fields ) : array {
 		}
 		$target = floatval( $fields['target'] );
 
-		if ( array_key_exists( 'completed_api', $fields ) ) {
+		if ( array_key_exists( 'completed_api', $fields ) && $fields['completed_api'] ) {
```
kirdia: I've added a similar change to conversions branch.
You could check  `! empty( $fields['completed_api'])` which will check both if the key exists and it does not have an empty value

pablocubico:
Closing this as it was fixed by a push by Kyriakos and merged into the Webpack branch.

PR #49, opened 2019-10-02
PLANET-4070 Convert auto-generated shortcakes to Gutenberg related ch…

sagarsdeshmukh:
…anges

`exclude_post_id` param is pass from autogenerated tag page article block.

Related [PR](https://github.com/greenpeace/planet4-master-theme/pull/925)


react-blocks/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -118,7 +118,11 @@ export class ArticlesBlock {
         button_link_new_tab: {
```
kirdia: Looks good! Can you add the field in transforms function also?

react-blocks/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -80,6 +80,12 @@ export class ArticlesBlock {
                   return attributes.named.tags ? attributes.named.posts.split(',') : [];
                 }
               },
+              exclude_post_id: {
+                type: 'string',
```
kirdia: can you change the type to integer and cast the value?

PR #52, opened 2019-10-03
PLANET-4149: Apply GP CH custom "Subscript / Superscript" commands to all blocks allowing in-page text editors

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4149


pablocubico:
@comzeradd aaah sure, that happens when I build and push using `npm start` instead of `npm run build`. 👍

PR #54, opened 2019-10-04
PLANET-4161: Image Block: Modify frontend rendering & commands

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4161

I found this solution in this thread: 
https://github.com/WordPress/gutenberg/issues/6499


assets/src/styles/editorStyle.scss
```diff
@@ -29,12 +30,15 @@
 @import "blocks/Timeline";
 @import "blocks/Counter";
 
+@import "blocks/core-overrides/Image";
```
kirdia: this is imported twice

kirdia:
Needs a rebase

Also not sure if I missed something but it does not look good on the frontend for me.

image block

PR #53, opened 2019-10-04
PLANET-4174 Implement conversion functionality - from shortcake blocks to gutenberg blocks

kirdia:
WIP

Add test cases for each block

- [x] Articles (pretty complete test cases, needs another final check) 🔵🔵🔵🔵🔵
- [ ] Carousel Header  ⚪️⚪️⚪️⚪️⚪️
- [ ] Columns  ⚪️⚪️⚪️⚪️⚪️
- [ ] Counter  🔵🔵🔵🔵⚪️
- [ ] Covers/Newcovers  ⚪️⚪️⚪️⚪️⚪️
- [ ] Gallery (Will need block code adjustments probably) ⚪️⚪️⚪️⚪️⚪️
- [ ] Happypoint (Will need block code adjustments probably)  ⚪️⚪️⚪️⚪️⚪️
- [ ] Media  ⚪️⚪️⚪️⚪️⚪️
- [ ] Socialmedia  ⚪️⚪️⚪️⚪️⚪️
- [ ] SplitTwoColumns  ⚪️⚪️⚪️⚪️⚪️
- [ ] Submenu  🔵⚪️⚪️⚪️⚪️
- [ ] Timeline  ⚪️⚪️⚪️⚪️⚪️

PR #56, opened 2019-10-10
Small fixes for various blocks

kirdia:
Break down PR #53 which includes various small fixes for the blocks attributes shemas and transforms functions.

Small fixes for various blocks attributes schema definitions
Add missing attributes in attributes schema definitions for some blocks
Fixes in transforms functions for various blocks
Fix covers_view in old covers twig template


sagarsdeshmukh:
PR Looks good to me, While testing on local and `sagardev` instance, I found that `Columns block, Split 2 columns block, Gallery block, Happy point block` has JS error, which breaks these blocks functionality in Gutenberg editor. IMO it is not related to this PR, so we can merge this PR.

PR #62, opened 2019-10-16
Add missing functions for cookies

comzeradd:
Cookie function don't seem to be inherited from master-theme. You can check that on the inspector on the [Privacy page](https://k8s.p4.greenpeace.org/gutenberg/privacy-and-cookies/) by (un-)clickinh the cookies checkbox.

### DIFF
--- a/public/js/cookies.js
+++ b/public/js/cookies.js
@@ -1,9 +1,29 @@
-/* global createCookie, readCookie */
+function createCookie(name, value, days) {
+  let date = new Date();
+  date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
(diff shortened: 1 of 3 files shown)


public/js/cookies.js
```diff
@@ -1,9 +1,29 @@
-/* global createCookie, readCookie */
+function createCookie(name, value, days) {
```
kirdia: These two (createCookie, readCookie) are declared in master-theme in global namespace

pablocubico:
@comzeradd @kirdia Didn't realize the scope was enclosed by the Webpack refactor, I globalized the functions here: https://github.com/greenpeace/planet4-master-theme/pull/936

PR #69, opened 2019-10-18
Fix carousel header warnings

kirdia:


classes/blocks/class-carouselheader.php
```diff
@@ -107,8 +107,11 @@ public function prepare_data( $fields ): array {
 			}
 
 			if ( isset( $slide['focal_points'] ) ) {
-				$focus_image          = round( $slide['focal_points']['x'] * 100, 0 ) . '% '
-										. round( $slide['focal_points']['y'] * 100, 0 ) . '%';
+
+				$x = isset( $slide['focal_points']['x'] ) ? round( $slide['focal_points']['x'] * 100, 0 ) . '% ' : '50%';
+				$y = isset( $slide['focal_points']['y'] ) ? round( $slide['focal_points']['y'] * 100, 0 ) . '% ' : '50%';
```
sagarsdeshmukh: Don't need space after percentage here.

comzeradd: ah missed your comment and merged already

PR #85, opened 2019-10-24
PLANET-4252 EMBEDS: set max width for all media types

dpivo:
- Widths ad heights set for video media types
- widths set for other media types
- unable to set for Facebook and height not properly set for vimeo (Investigation ticket to follow)


classes/class-loader.php
```diff
@@ -346,6 +346,8 @@ public function enqueue_public_assets() {
 			],
 			$css_blocks_creation
 		);
+
+		wp_enqueue_script( 'media', P4GBKS_PLUGIN_URL . 'public/js/media.js', [ 'jquery' ], '0.1', true );
```
comzeradd: You are deleting this script in this PR. So do we still need it?

dpivo: @comzeradd Good point. I removed it

PR #88, opened 2019-10-24
PLANET-4320 Counter block- progress bar style: counter not added to form

dpivo:
- Added missing JS to append counter block inside en-form block

### DIFF
--- a/classes/blocks/class-counter.php
+++ b/classes/blocks/class-counter.php
@@ -122,6 +122,11 @@ class Counter extends Base_Block {
 			$fields['text']
 		);
 
+		// Enqueue js for the frontend.
+		if ( ! $this->is_rest_request() ) {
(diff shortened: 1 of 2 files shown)


classes/class-loader.php
```diff
@@ -346,6 +346,8 @@ public function enqueue_public_assets() {
 			],
 			$css_blocks_creation
 		);
+
+		wp_enqueue_script( 'counter', P4GBKS_PLUGIN_URL . 'public/js/counter.js', [ 'jquery' ], '0.1', true );
```
sagarsdeshmukh: Could you enqueue this js only for counter block. [eg](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/classes/blocks/class-cookies.php#L106)

sagarsdeshmukh:
Tested locally, Looks good 👍  need rebase and requested small change.

dpivo:
> Tested locally, Looks good 👍 need rebase and requested small change.

@sagarsdeshmukh Change done and rebased, Please have a look.

PR #92, opened 2019-10-28
PLANET-3928: Image: caption is wider than photo itself, no credit in the front end

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3928


galousis:
Seems ok

PR #93, opened 2019-10-28
PLANET-4293 add wpautop filter to make paragraphs out of line breaks …

Inwerpsel:
…in block's main description fields

### DIFF
--- a/templates/blocks/articles.twig
+++ b/templates/blocks/articles.twig
@@ -9,7 +9,7 @@
 					</header>
 				{% endif %}
 				{% if ( fields.articles_description ) %}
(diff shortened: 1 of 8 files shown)


pablocubico:
Hey @Inwerpsel @kirdia @comzeradd I added a HOLD label as I think this might be superseded by: 

https://github.com/greenpeace/planet4-master-theme/pull/943

Which adds `wpautop` to everything, using the `the_content` filter, as some random dudes in the internet suggest. Just kidding, it seems to work (I need to test it with more content).

If it works, we should also revert this one I did before: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/57

pablocubico:
@galousis hmmm not so sure, not so sure... not as magic as I hoped it would be...

PR #94, opened 2019-10-28
PLANET-4282 use null coalescing operator to avoid undefined indices

Inwerpsel:


classes/blocks/class-splittwocolumns.php
```diff
@@ -124,10 +124,10 @@ public function prepare_data( $fields ): array {
 		$issue_id        = absint( $fields['select_issue'] ?? '' );
 		$issue_meta_data = get_post_meta( $issue_id );
 
-		$tag_id            = absint( $fields['select_tag'] ? $fields['select_tag'] : '' );
+		$tag_id            = absint( $fields['select_tag'] ?? '' );
 		$tag               = get_term( $tag_id );
-		$campaign_image_id = $fields['tag_image'] ? $fields['tag_image'] : get_term_meta( $tag_id, 'tag_attachment_id', true );
-		$issue_image_id    = $fields['issue_image'] ? $fields['issue_image'] : get_post_thumbnail_id( $issue_id );
+		$campaign_image_id = $fields['tag_image'] ?? get_term_meta( $tag_id, 'tag_attachment_id', true );
+		$issue_image_id    = $fields['issue_image'] ?? get_post_thumbnail_id( $issue_id );
```
pablocubico: There is an issue we faced before with null coalescing operators, for example, when rendering a shortcode such as:

```
[shortcake_split_two_columns select_issue="60" focus_issue_image="left top" select_tag="--Select Tag--" focus_tag_image="left top" /]
```

The `$fields['issue_image']`  value is an empty string, not null, so the null comparison fails and the thumbnail is never displayed. As a workaround, we were using this (as @kirdia suggested): 

`$issue_image_id  = !empty( $fields['issue_image'] ) ? $fields['issue_image'] : get_post_thumbnail_id( $issue_id );`

Another way around would be to initialize the values as `null` in the `add_block_shortcode` method of this file, but I don't remember now if that can cause side effects on other places... 🤔 

So I guess I would recommend going for the `!empty(...)` solution.

pablocubico: Now that I mention it, I realized we _ideally_ won't have shortcodes after the Gutenberg migration, but I guess I would use `empty` just in case... I guess it's a bit more defensive, also in case a block passes down some empty-ish value which is not null.

PR #96, opened 2019-10-28
PLANET-4303 Add script that equalizes column headings height like in …

Inwerpsel:
…planet4-plugin-blocks

### DIFF
--- a/classes/blocks/class-columns.php
+++ b/classes/blocks/class-columns.php
@@ -165,7 +165,7 @@ class Columns extends Base_Block {
 		$attributes_normalized['no_of_columns'] = $columns_set;
 
 		// Define the image size that will be used, based on layout chosen and number of columns.
-		$columns_block_style = $attributes['columns_block_style'];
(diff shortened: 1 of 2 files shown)


classes/blocks/class-columns.php
```diff
@@ -185,6 +185,11 @@ public function prepare_data( $attributes ): array {
 			}
 		}
 
+		// enqueue script that equalizes the heights of the titles of the blocks
+		if ( ! $this->is_rest_request() ) {
+			wp_enqueue_script( 'submenu', P4GBKS_PLUGIN_URL . 'public/js/columns.js', [ 'jquery' ], '0.1', true );
```
pablocubico: Hey @Inwerpsel I guess this "submenu" should be "column-headers" or something like that.

Inwerpsel: Good catch!

pablocubico:
Merged!

PR #100, opened 2019-10-31
PLANET-4227 Gallery and Submenu blocks: selection not visible when switching style

dpivo:
- Fixed layout selector not preserving number types

Ref: https://jira.greenpeace.org/browse/PLANET-4227


assets/src/components/LayoutSelector/LayoutSelector.js
```diff
@@ -19,12 +19,17 @@ export class LayoutSelector extends Component {
     return <div className='LayoutSelector'>
       {
         this.props.options.map((layoutOption, i) => {
+
+          if ( 'number' == typeof layoutOption.value ) {
```
pablocubico: Hey @dpivo I think we should rather avoid doing type-checking and/or casting in the LayoutSelector (as we did in the past!) could you just pass the values as strings as in the Covers block: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/assets/src/blocks/Covers/Covers.js#L194 ?

The value is already being casted to number in the Gallery block so I guess it should work right away:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/assets/src/blocks/Gallery/GalleryBlock.js#L160

Thanks!

dpivo: Hey @pablocubico I did it like this because even though the 'gallery_block_style' is being cast to a number, when it is passed to the LayoutSelector it for some reason get's cast to a string. but this is only when changing the style, on editing of the block if a style has previously been selected, it pulls through as a number when the block is opened for editing.

galousis: @pablocubico @dpivo Tested locally and seems okay

galousis:
@pablocubico @dpivo Tested locally and seems okay

assets/src/components/LayoutSelector/LayoutSelector.js
pablocubico: @dpivo I hear ya, maybe I'm being too defensive, let's go with your solution and make the LayoutSelector smarter then, it's not inside a loop running 10 million times anyway. Let's just test this doesn't break the selector on any of the other blocks where it's used.

PR #108, opened 2019-11-01
PLANET-4340 add component derived from wordpress core FlatTermSel…

Inwerpsel:
…ector (https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/components/post-taxonomies/flat-term-selector.js). The new component has the same functionality, except when entering non-existing terms it does not instantly creates these.

The original component is being replaced using a hook provided by wordpress (see https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/components/post-taxonomies/flat-term-selector.js#L264). I have considered modifying the component in the hook, but as functionality is removed and not added it seemed more straightforward to replace it with a simpler form of the component.


Inwerpsel:
Perhaps it is easier to review this by starting with a diff with the wordpress component this is based on  https://github.com/WordPress/gutenberg/blob/master/packages/editor/src/components/post-taxonomies/flat-term-selector.js .

PR #113, opened 2019-11-06
[No ticket yet] Fix media block

pablocubico:
The videos are not expanding full-width because the new plugin is missing the `media.js` script from the old plugin.

Example:
https://release.k8s.p4.greenpeace.org/international/explore/ships/


comzeradd:
I remember @dpivo removed those at 89665b871ad48a186805f4ed1cde86cd95dee101

comzeradd:
This is the ticket, if you want to update it: https://jira.greenpeace.org/browse/PLANET-4384

PR #114, opened 2019-11-07
PLANET-4154 set download button toggle to false by default and style

dpivo:
Ref: https://jira.greenpeace.org/browse/PLANET-4154

- Added filter to set download button attribute to false by default.
- Added script to change class of button to our button secondary class.


public/js/file.js
```diff
@@ -0,0 +1,8 @@
+//Set download button class to our class
```
comzeradd: To avoid having a js running for overwriting each of these buttons, maybe it's easier to add `wp-block-file__button` to the button component on the styleguide. Like we did for native button classes. See here: https://github.com/greenpeace/planet4-styleguide/blob/3987dab893b8ca0797785fa8eb1e7ed912ee3eee/src/components/_buttons.scss#L80

I haven't tested it, but this maybe removes the need for the extra `File.scss` as well.

PR #123, opened 2019-11-11
PLANET-4387 Fix articles block ignore categories setting and selected…

sagarsdeshmukh:
… tags issue

[JIRA 4387](https://jira.greenpeace.org/browse/PLANET-4387)

Observations :
1. Issue found for non-existing tags - added validation on tag ids
    1. In case of non-existing tags/no tags selected - show articles post on the basis of current page/post tags (same as old articles block)
2. Issue found for ignore category setting(The ignore category setting not working as expected) - issue fixed

Issue tested on local env -
Before changes -
image

After changes -
image

### DIFF
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -366,7 +366,6 @@ class Articles extends Base_Block {
 
 		$exclude_post_id   = (int) ( $fields['exclude_post_id'] ?? '' );
 		$ignore_categories = $fields['ignore_categories'];
-		$options           = get_option( 'planet4_options' );
 
 		// Get page categories.
 		$post_categories   = get_the_category();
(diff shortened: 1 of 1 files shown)


classes/blocks/class-articles.php
```diff
@@ -381,8 +380,21 @@ private function filter_posts_by_page_types_or_tags( &$fields ) {
 		// Get user defined tags from backend.
 		$tags = $fields['tags'] ?? [];
 
+		// Validate tag ids.
+		if ( ! empty( $tags ) ) {
+			$validated_tags = [];
```
Inwerpsel: Small suggestion: `array_filter` is a bit easier to read, and you would not need the empty check and inner assignments
```$tags = array_filter($tags, function($tag_id) { return get_tag( $tag_id ) instanceof \WP_Term; }```

PR #129, opened 2019-11-14
PLANET-4360 split campaign assets into separate themes

dpivo:
Ref: https://jira.greenpeace.org/browse/PLANET-4360


dpivo:
> LGTM! Tested locally, the size warning is no longer showing and the Campaign pages load faster.
> 
> One minor detail: could you remove `campaign.min.css` from the repo since it's no longer used?

@pablocubico Sorted

PR #139, opened 2019-11-19
PLANET-4401 : Disable Take action and Split block for Campaigns posts.

galousis:
[Ref](https://jira.greenpeace.org/browse/PLANET-4401)

If any improvements please


assets/src/blocks/Covers/Covers.js
```diff
@@ -22,6 +22,41 @@ export class Covers extends Component {
...
+          label: __('Content Covers', 'p4ge'),
+          image: window.p4ge_vars.home + 'images/content_covers.png',
+          value: '3',
+          help: __('Content covers pull the image from the post.')
+        }];
+
+      // Get current post type to filter LayoutSelector options in the case of "campaigns" post type.
+      let currentPostType = wp.data.select('core/editor').getCurrentPostType();
+
+      if ('campaign' === currentPostType)
```
Inwerpsel: You can do the assignment inside of this if (and add an else) so that it doesn't get assigned and overwritten immediately for campaigns.

Inwerpsel: Or maybe better: create the array with the common element first, then add the others if we're not on a campaign. That way you don't have to duplicate the definition of the common option.

comzeradd:
Closing in favor of #154

PR #141, opened 2019-11-20
Planet 4438 Add NRO custom blocks

Inwerpsel:
I initially started for a broader reworking of the report page, so I started with a cleanup commit to remove some boilerplate code. The scope of the ticket then narrowed to only include making the report also include NRO developed blocks. However it makes sense to already include the cleanup commits as it only removes things (+ small change to use ::class instead of string class names). I kept both commits separate as they do different things.

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -234,11 +234,9 @@ P4GBKS\Loader::get_instance(
 		// --- Add here your own Block Controller ---
 		// DEPRECATED: Blocks could be registered inside Loader class
 		// 'P4GBKS\Controllers\Blocks\NewCovers_Controller'
-		'P4GBKS\Controllers\Menu\Settings_Controller',
-		'P4GBKS\Controllers\Menu\Blocks_Usage_Controller',
-		'P4GBKS\Controllers\Menu\Reusable_Blocks_Controller',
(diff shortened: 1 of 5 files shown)


planet4-gutenberg-blocks.php
```diff
@@ -234,11 +234,9 @@ function set_allowed_block_types( $allowed_block_types, $post ) {
...
-		'P4GBKS\Controllers\Menu\Blocks_Usage_Controller',
-		'P4GBKS\Controllers\Menu\Reusable_Blocks_Controller',
+		\P4GBKS\Controllers\Menu\Settings_Controller::class,
+		\P4GBKS\Controllers\Menu\Blocks_Usage_Controller::class,
+		\P4GBKS\Controllers\Menu\Reusable_Blocks_Controller::class,
 	],
-	'P4GBKS\Views\View',
-	'P4GBKS\Command\Shortcode_To_Gutenberg',
-	'P4GBKS\Command\Controller'
+	\P4GBKS\Views\View::class
```
Inwerpsel: The lines below this, being parameter 3 and 4 of the function call, had no effect as the function only has 2 parameters

Inwerpsel:
phpunit test is only failling because of an issue while fetching test data repository of wordpress using svn. Probably a temporary network error (also happens on other branches), so will try again later :crossed_fingers: .

PR #144, opened 2019-11-21
PLANET-4415 CPP - Columns block - No Image style - Climate Emergency template - Description width is smaller than on other templates

dpivo:
Removed formatting columns no image description in climate theme.

Ref: https://jira.greenpeace.org/browse/PLANET-4415


dpivo:
> Looks good 👍
> 
> Just needs a rebase

@comzeradd Rebased

PR #146, opened 2019-11-21
PLANET-4425 Fix columns block open link in new tab

Inwerpsel:
NOTE: I had started with some necessary refactors before we noticed this needed to be deployed in a hotfix. However I suggest we keep these changes as merging them quickly will reduce the chance of merge conflicts.

* Change type of columns.link_new_tab to boolean (a filter will be added
later to deal with existing blocks that have an empty string or
something else in that field)
* Remove the transformation to flat array with index postfixes
* Move the logic of filtering columns (based on attributes and total
number of columns) to a single place in the block class instead of in
many places in the templates
* Use a hook that is called before the blocks attributes are validated.
Unfortunately the block renderer validates the request prior to when the
block is rendered, and that hook is not called there.
* Furthermore there doesn't seem to be a hook in the REST API to modify
the block to be rendered before it is validated, so the only option was
the modify this on the frontend just before it calls the API.


templates/blocks/columns.twig
```diff
@@ -17,44 +17,43 @@
...
-							{% endif %}
+								<h3>
+									{% if col.cta_link and not is_campaign %}<a href="{{ col.cta_link|e('esc_url') }}">{% endif %}
+										{{ col.title|e('wp_kses_post')|raw }}
+										{% if col.cta_link and not is_campaign %}</a>{% endif %}
+								</h3>
+								{% if col.description %}
+									<p>{{ col.description|e('wp_kses_post')|raw|nl2br }}</p>
+								{% endif %}
+								{% if col.cta_text and link %}
```
sagarsdeshmukh: It would be `col.cta_link` instead of `link`.

templates/blocks/columns.twig
```diff
@@ -17,44 +17,43 @@
...
+									<a href="{{ col.cta_link|e('esc_url') }}"
+										{{  col.link_new_tab ? 'target="_blank"' }}
+										{% if fields.columns_block_style == 'no_image' %}
+										class="btn btn-secondary"
+									{% else %}
```
sagarsdeshmukh: The else block needs indentation.

templates/blocks/columns_tasks.twig
```diff
@@ -15,70 +15,60 @@
...
-									<div class="col" data-id="{{ i }}">
-										{% if ( fields['description_'~i] ) %}
-											<p {% if ( fields['cta_text_'~i] and fields['link_'~i] ) %}class="has-cta"{% endif %}>
-												{{ fields['description_'~i]|e('wp_kses_post')|raw }}
-											</p>
-										{% endif %}
-									</div>
-								{% endif %}
+							{% for col in fields.columns %}
+								<div class="col" data-id="{{ i }}">
```
sagarsdeshmukh: The `i`  will replace by `loop.index`.

sagarsdeshmukh:
The `Icon` column style not respect the open in new tab admin setting. it always opens in new tab.
https://k8s.p4.greenpeace.org/sagardev/test-column-icon-style-block/

sagarsdeshmukh:
On edit page, I am getting error for column block -

image

https://k8s.p4.greenpeace.org/sagardev/wp-admin/post.php?post=999&action=edit

@Inwerpsel - could you please check and confirm on it.

Inwerpsel:
> On edit page, I am getting error for column block -
> 
> image
> 
> https://k8s.p4.greenpeace.org/sagardev/wp-admin/post.php?post=999&action=edit
> 
> @Inwerpsel - could you please check and confirm on it.



> The `Icon` column style not respect the open in new tab admin setting. it always opens in new tab.
> https://k8s.p4.greenpeace.org/sagardev/test-column-icon-style-block/

There apparently is a script that makes all non-internal links open in a new tab. So this does work as intended when the link is internal, i.e. it starts with "/". However this script doesn't run in the block preview, which causes the preview to not reflect the actual frontend appearance in case of an external url.

I suggest we take that up in a follow up ticket, as this is already the behavior on production.

Inwerpsel:
> On edit page, I am getting error for column block -
> 
> image
> 
> https://k8s.p4.greenpeace.org/sagardev/wp-admin/post.php?post=999&action=edit
> 
> @Inwerpsel - could you please check and confirm on it.

This was because I didn't version bump editorIndex.js

PR #149, opened 2019-11-22
PLANET-4428: Webpack: Source maps don't work

pablocubico:
The source maps were broken because the minified files are being symlinked from the root folder, and the `sourceMap` option was missing in some of the steps of the SCSS chain.

So the CSS should live inside `build` and the image paths should be relative to its location. We can keep the symlink version of `style.min.css` in the root  if it's needed by WP, but we should load the one inside `assets/build` in the frontend.

Note: I tried using the `SourceMapDevToolPlugin` plugin  and setting the `append` or `publicPath` properties to alter the `sourceMappingURL` line in the minified files, but it didn't seem to work properly, also, adding the plugin would affect the JS files too.


comzeradd:
For the plugins, I think it's not needed to have the styles on the root folder (only for theme). So we could completely remove the symlinks.

pablocubico:
> For the plugins, I think it's not needed to have the styles on the root folder (only for theme). So we could completely remove the symlinks.

Ugh, I forgot to do this on the master-theme! Duly noted.

PR #151, opened 2019-11-25
PLANET-4448 Implemet Social Share block

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-4448


comzeradd:
Thanks! I added the style to `editorStyle.scss`

PR #73, opened 2019-10-22
WYSIWYG Counter example

pablocubico:
For more context, see the docs in this ticket's comments: https://jira.greenpeace.org/browse/PLANET-4205

This is an example for improving the UX of a block's editing UI by converting it to a WYSIWYG approach, the effort is:

- Copying the Twig template into JSX, converting `class` to `className` and copying the Twig logic into JS.
- Convert some elements to RichText editor
- Show the WYSIWYG editor on focus and the ServerSideRender on blur.
- Move some elements from the UI to the Sidebar using `InspectorControls`

counter-wysiwyg

PR #153, opened 2019-11-26
PLANET-4449: Carousel Header: review initialization code

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4449

Replaced the ServerSideRender reference for a DOM-based check on an interval (which gets cleared on success), as the previous approach was unreliable as far as I tested. 

Surprisingly, there is no success Callback on the ServerSideRender component (as noted in [this blog post](https://www.ibenic.com/gutenberg-components-server-side-render/)). The blog post mentions that the feature is on its way to be added but the docs for the ServerSideRender component itself are pretty non-existent, and they actually say: 

"This package is meant to be used only with WordPress core. Feel free to use it in your own project but please keep in mind that it might never get fully documented."
https://developer.wordpress.org/block-editor/packages/packages-server-side-render/

So I guess we may change the rendering into a WYSIWYG thing on the React side as I sampled in this PR and eventually stop using the ServerSideRender component:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/73

And maybe as a workaround in the meantime, we could roll our own copy of ServerSideRender adding a success callback to it.


classes/blocks/class-carouselheader.php
```diff
@@ -83,6 +83,12 @@ public function __construct() {
 				],
 			]
 		);
+
+		add_action( 'enqueue_block_editor_assets', [ $this, 'enqueue_editor_scripts' ] );
+	}
+
+	public function enqueue_editor_scripts() {
```
comzeradd: Can you add a doc comment for this function? (see phpcs error)

pablocubico:
@comzeradd updated!

PR #157, opened 2019-11-27
PLANET-4251 Add option to remove single image from Gallery block

sagarsdeshmukh:
[JIRA 4251](https://jira.greenpeace.org/browse/PLANET-4251)

### DIFF
--- a/assets/src/blocks/Gallery/Gallery.js
+++ b/assets/src/blocks/Gallery/Gallery.js
@@ -6,7 +6,7 @@ import {
   ServerSideRender,
   Toolbar,
   IconButton,
-  Button
+  Button, Tooltip
 } from '@wordpress/components';
 import {BlockControls,MediaUpload,MediaUploadCheck} from "@wordpress/editor";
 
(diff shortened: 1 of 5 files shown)


pablocubico:
Hey @sagarsdeshmukh this looks and works great! I only see one issue: the Tooltip is not discoverable on tablets or phones, because there is no pointer, so it's always tricky to rely on hover on those devices. So I would suggest just making the (X) always visible in all thumbnails, as it was in the shortcode version (it's on the ticket's attachments).

sagarsdeshmukh:
> Hey @sagarsdeshmukh this looks and works great! I only see one issue: the Tooltip is not discoverable on tablets or phones, because there is no pointer, so it's always tricky to rely on hover on those devices. So I would suggest just making the (X) always visible in all thumbnails, as it was in the shortcode version (it's on the ticket's attachments).

Hi @pablocubico ,
Thanks for the review, I updated the PR as per suggestion. Please have a look.
image

PR #159, opened 2019-11-28
PLANET-4427 remove built assets

Inwerpsel:
These can be removed as the builder now builds these whenever it installs a branch of a plugin/theme.

Master branch will still have these files when changes are merged.


comzeradd:
I think it's best to target the develop branch first.

Can you also add `assets/build` to `.gitignore`?

PR #169, opened 2019-12-05
PLANET-4538 Force SocialMediaCards images to download and PLANET-4539 improve twitter

Inwerpsel:
* Add a title on all buttons (not in ticket but discussed with Will and can be included here)

https://jira.greenpeace.org/browse/PLANET-4538
* Force SocialMediaCards image download button to always download instead of showing in browser

https://jira.greenpeace.org/browse/PLANET-4539
* Make tweet open in popup instead of new tab
* keep the original href as fallback, though it's not used by js
* Url encode to make hashtags work

### DIFF
--- a/classes/blocks/class-socialmediacards.php
+++ b/classes/blocks/class-socialmediacards.php
@@ -66,6 +66,10 @@ class SocialMediaCards extends Base_Block {
 	 * @return array The data to be passed in the View.
 	 */
 	public function prepare_data( $fields ): array {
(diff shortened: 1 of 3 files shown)


Inwerpsel:
> Looks Good!
> But, Is the PR supposed to be against PLANET-4531 and not the release?

Ah thanks, still needed to change the target branch to release, did that now. I made it against the other branch previously to better see the difference, but that one was merged already.

PR #170, opened 2019-12-05
PLANET-4527 Add API endpoint for Blocks Usage

sagarsdeshmukh:
[JIRA 4527](https://jira.greenpeace.org/browse/PLANET-4527)

Block Usage API endpoint example: https://k8s.p4.greenpeace.org/sagardev/wp-json/plugin_blocks/v2/plugin_blocks_report/

### DIFF
--- a/classes/controller/menu/class-blocks-usage-controller.php
+++ b/classes/controller/menu/class-blocks-usage-controller.php
@@ -25,6 +25,51 @@ if ( ! class_exists( 'Blocks_Usage_Controller' ) ) {
 		];
 
 		/**
+		 * Blocks_Usage_Controller constructor.
+		 *
+		 * @param View $view The view object.
+		 */
+		public function __construct( $view ) {
+			parent::__construct( $view );
+			$this->hooks();
+		}
+
+		/**
+		 * Class hooks.
+		 */
+		private function hooks() {
+			add_action( 'rest_api_init', [ $this, 'plugin_blocks_report_register_rest_route' ] );
+		}
+
+		/**
+		 * Register API route for report of blocks usage in pages/posts.
+		 */
+		public function plugin_blocks_report_register_rest_route() {
+			register_rest_route(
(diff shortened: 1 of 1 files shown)


classes/controller/menu/class-blocks-usage-controller.php
```diff
@@ -73,19 +125,24 @@ public function plugin_blocks_report() {
...
+							<th>' . __( 'ID', 'planet4-blocks-backend' ) . '</th>
+							<th>' . __( 'Title', 'planet4-blocks-backend' ) . '</th>
+					</tr>';
+					foreach ( $results as $result ) {
+						echo '<tr><td>'.$result->ID .'</td>';
+						echo  '<td><a href="post.php?post=' . $result->ID . '&action=edit" >' . $result->post_title . '</a></td></tr>';
+					}
+					echo '</table>';
+				} else {
+					$report[ ucfirst( str_replace( '_', ' ', $block_type ) ) ] = count($results);
```
Inwerpsel: What is the reason for using ucfirst on the block type here?

classes/controller/menu/class-blocks-usage-controller.php
```diff
@@ -199,18 +276,31 @@ public function plugin_blocks_report() {
...
-			echo '</table>';
 			// phpcs:enable
+
+			if ( 'json' === $type ) {
+				return $report;
+			}
 		}
 
 		/**
+		 * Return All WP block types(P4+Core).
```
Inwerpsel: Perhaps I should have added a comment here when I wrote this function. But maybe you can adjust it to say that it only returns some core block types, not all of them.

classes/controller/menu/class-blocks-usage-controller.php
sagarsdeshmukh: The API fields(block type name) display as a header title in Spreadsheet report.

PR #182, opened 2019-12-16
PLANET-4540 Fix facebook share and remove unused code

Inwerpsel:
* Facebook share now opens in new tab, providing the
social_media_message as quoted text
* All SocialMediaCards code was initially copied from the Gallery
block as a starting point, however this was never cleaned up. This
removes some of it, except for the focal points, which don't work but it
is not clear whether it should be removed or the functionality should be
implemented.
* Remove the conversion from shortcode.
* Prevent notices for the link_new_tab fix

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -243,7 +243,7 @@ function empty_string_to_false_in_link_new_tab_in_columns_blocks( $block ): arra
 	// Yes, that's right, WordPress doesn't follow its own rules here so we have a camel among snakes.
 	if ( 'planet4-blocks/columns' === $block['blockName'] ?? null ) {
 		foreach ( $block['attrs']['columns'] ?? [] as $key => $column ) {
-			if ( true !== $column['link_new_tab'] ) {
+			if ( isset( $column['link_new_tab'] ) && true !== $column['link_new_tab'] ) {
 				$block['attrs']['columns'][ $key ]['link_new_tab'] = false;
 			}
 		}
--- a/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
+++ b/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
@@ -3,8 +3,6 @@ import { Preview } from '../../components/Preview';
 
 
 import {
-  MediaPlaceholder,
-  InspectorControls,
   BlockControls,
   MediaUpload,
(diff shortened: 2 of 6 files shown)


planet4-gutenberg-blocks.php
```diff
@@ -243,7 +243,7 @@ function empty_string_to_false_in_link_new_tab_in_columns_blocks( $block ): arra
 	// Yes, that's right, WordPress doesn't follow its own rules here so we have a camel among snakes.
 	if ( 'planet4-blocks/columns' === $block['blockName'] ?? null ) {
 		foreach ( $block['attrs']['columns'] ?? [] as $key => $column ) {
-			if ( true !== $column['link_new_tab'] ) {
+			if ( isset( $column['link_new_tab'] ) && true !== $column['link_new_tab'] ) {
```
sagarsdeshmukh: Good catch 👍

sagarsdeshmukh:
Hi @Inwerpsel - could you please changes the get `source_url` to -
`image_urls_array[img_id] = img_url.source_url;`
at below 
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/PLANET-4540/assets/src/blocks/SocialMediaCards/SocialMediaCardsBlock.js#L53
Reference : https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/PLANET-4540/assets/src/blocks/Gallery/GalleryBlock.js#L95

sagarsdeshmukh:
PR looks good to me, the changes works as per JIRA ticket req.

General finding -
- The remove selected image option not visible.
- The title and desc not display in admin editor preview
- The textdomain value is different at some places
- The label of field are "Select Gallery Images" , "Select gallery image focal point" which not relevant to 'Social media card' block(its mainly because of copy paste from gallery block)

Inwerpsel:
> PR looks good to me, the changes works as per JIRA ticket req.
> 
> General finding -
> 
> * The remove selected image option not visible.
> * The title and desc not display in admin editor preview
> * The textdomain value is different at some places
> * The label of field are "Select Gallery Images" , "Select gallery image focal point" which not relevant to 'Social media card' block(its mainly because of copy paste from gallery block)

Yeah I saw some of those issues as well, but it's so much that it deserves its own ticket. Perhaps we can take those remarks along when we fix the focal point behavior.

PR #183, opened 2019-12-16
PLANET-4559 - Feature/spreadsheet table

alagregor:


classes/blocks/class-spreadsheettable.php
```diff
@@ -0,0 +1,91 @@
...
+	public function prepare_data( $fields ): array {
+		$fields['url']       = '' !== $fields['url'] ? $fields['url'] : '';
+
+		// Enqueue script for table filter
+		wp_enqueue_script( 'spreadsheet-table', P4GBKS_PLUGIN_URL . 'public/js/filter_spreadsheet_table.js', [ 'jquery' ], '0.2', true );
+
+		$spreadsheet_data = [];
+		// Firstly check if url is correct
+		if (strpos($fields['url'], 'output=csv') !== false) {
+			if (($handle = fopen($fields['url'], "r")) !== FALSE) {
```
Inwerpsel: Please always use lowercase `true` and `false`

classes/blocks/class-spreadsheettable.php
```diff
@@ -0,0 +1,91 @@
...
+				while (($data = fgetcsv($handle, 1000, ",")) !== FALSE) {
```
Inwerpsel: Perhaps it would be prudent to have an upper limit to the amount of rows that will be read, or the size of the file to be loaded. This can be relatively high, but no upper limit here could cause unexpected issues.

classes/blocks/class-spreadsheettable.php
```diff
@@ -0,0 +1,91 @@
...
+					$spreadsheet_data[] = $data;
+				}
+				fclose($handle);
+			}
+		}
+		$fields['spreadsheet_data'] = $spreadsheet_data;
+
+		$data = [
```
Inwerpsel: I know this is currently not the case in most other places, but `$data` is redundant and you can just `return [ ... ]`

classes/class-loader.php
```diff
@@ -108,7 +108,12 @@ private function __construct( $services, $view_class ) {
 			new Blocks\SubMenu(),
 			new Blocks\TakeActionBoxout(),
 			new Blocks\Timeline(),
+<<<<<<< HEAD
```
Inwerpsel: Please properly resolve this merge conflict.

gulpfile.js
```diff
@@ -0,0 +1,124 @@
+/* global require, exports */
```
Inwerpsel: These files were removed by us, but mistakenly re-added during conflict resolution.

package-lock.json
```diff
@@ -861,6 +860,43 @@
 				}
 			}
 		},
+		"@gulp-sourcemaps/identity-map": {
```
Inwerpsel: Same here, I don't think you intended to change this file

package.json
```diff
@@ -14,21 +14,39 @@
...
 	"devDependencies": {
-		"@wordpress/components": "^8.3.2",
-		"@wordpress/data": "^4.9.2",
 		"@wordpress/scripts": "3.3.0",
-		"autoprefixer": "^9.7.0",
-		"bootstrap": "^4.4.1",
-		"css-loader": "^3.2.0",
-		"lodash": "^4.17.15",
-		"lodash.assign": "^4.2.0",
+		"autoprefixer": "^9.7.3",
```
Inwerpsel: Also changed by conflict resolution

planet4-gutenberg-blocks.php
```diff
@@ -118,6 +118,7 @@
 	'planet4-blocks/submenu',
 	'planet4-blocks/take-action-boxout',
 	'planet4-blocks/timeline',
+	'planet4-blocks/spreadsheettable',
```
Inwerpsel: Words in the block name should be separated by a dash, so that would make `spreadsheet-table` here.

public/js/filter_spreadsheet_table.js
```diff
@@ -0,0 +1,11 @@
+jQuery(function ($) {
+  'use strict';
+
+  	// Filter Spreadsheet table by entered value in search field
+	$('.spreadsheet-table-search').keyup(function() {
```
Inwerpsel: Please use a different event than `keyup`. Keyup triggers at literally any keyup, also when the values don't change (like arrow keys, tabbing to the field, ...). Keyup also doesn't trigger when pasting using the mouse. Probably the `change` event would be better here.

public/js/filter_spreadsheet_table.js
```diff
@@ -0,0 +1,11 @@
...
+		var value = $(this).val().toLowerCase();
+		$('.spreadsheet-table .table-row:not(:eq(0))').filter(function() {
+			$(this).toggle($(this).text().toLowerCase().indexOf(value) > -1);
```
Inwerpsel: You can use the more readable `$(this).text().toLowerCase().includes(value)`

public/js/filter_spreadsheet_table.js
```diff
@@ -0,0 +1,11 @@
+jQuery(function ($) {
```
Inwerpsel: This could be written without jQuery, which will be an advantage when we decide to remove it, as javascript by now supports  pretty much everything.

templates/blocks/spreadsheettable.twig
```diff
@@ -0,0 +1,31 @@
+{% block spreadsheettable %}
+
```
Inwerpsel: The block shouldn't display anything if no matching file was found or if there is no data. Currently the search box is always displayed.

classes/blocks/class-spreadsheettable.php
```diff
@@ -0,0 +1,91 @@
...
+	 *
+	 * @return array The data to be passed in the View.
+	 */
+	public function prepare_data( $fields ): array {
+		$fields['url']       = '' !== $fields['url'] ? $fields['url'] : '';
+
+		// Enqueue script for table filter
+		wp_enqueue_script( 'spreadsheet-table', P4GBKS_PLUGIN_URL . 'public/js/filter_spreadsheet_table.js', [ 'jquery' ], '0.2', true );
+
+		$spreadsheet_data = [];
```
Inwerpsel: Maybe it would make sense to call this array `$rows`? It will make the twig template more readable.

templates/blocks/spreadsheettable.twig
```diff
@@ -0,0 +1,31 @@
...
+	<input class="spreadsheet-table-search" type="text" placeholder="Search by Name, Position or Country" />
+	<div class="spreadsheet-table">
+		{% for key, item in fields.spreadsheet_data %}
+
+			<div class="table-row">
```
Inwerpsel: What is the reason for not using the html `<table>` element here to display a table?

classes/blocks/class-spreadsheettable.php
```diff
@@ -0,0 +1,91 @@
...
+	const BLOCK_NAME = 'spreadsheettable';
+	/**
+	 * Register shortcake shortcode.
+	 *
+	 * @param array  $attributes Shortcode attributes.
+	 * @param string $content   Content.
+	 *
+	 * @return mixed
+	 */
+	public function add_block_shortcode( $attributes, $content ) {
```
Inwerpsel: Same here, everything shortcode related is not needed anymore.

classes/blocks/class-spreadsheettable.php
```diff
@@ -0,0 +1,91 @@
...
+	 */
+	public function prepare_data( $fields ): array {
+		$fields['url']       = '' !== $fields['url'] ? $fields['url'] : '';
+
+		// Enqueue script for table filter
+		wp_enqueue_script( 'spreadsheet-table', P4GBKS_PLUGIN_URL . 'public/js/filter_spreadsheet_table.js', [ 'jquery' ], '0.2', true );
+
+		$spreadsheet_data = [];
+		// Firstly check if url is correct
+		if (strpos($fields['url'], 'output=csv') !== false) {
```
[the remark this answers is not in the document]
Inwerpsel: Also in the current form the server will load any file from any url, which is a security risk. If the block would use a regex instead to get the id from the entered url, then use that id to format a google sheets url, then that cannot happen.

templates/blocks/spreadsheettable.twig
```diff
@@ -0,0 +1,31 @@
...
+					{% for keyField, itemField in item %}
+
+						{% if ( keyField == 1 ) %}
+							<div class="col-6 pr-0 pl-0">
+						{% else %}
+							<div class="col-3 pr-0 pl-0">
+						{% endif %}
+
+							<div class="item">
+								{{itemField}}
```
Inwerpsel: The contents of the field should be escaped here

planet4-gutenberg-blocks.php
comzeradd: Please don't enabled it for pages post_type, only for campaigns.

templates/blocks/spreadsheettable.twig
alagregor: Because of handling responsive table, it is easier.

alagregor:
Most of the things are fixed now, pushed to GIT. Please check if it is ok now.

Inwerpsel:
> Most of the things are fixed now, pushed to GIT. Please check if it is ok now.

Hey Alan, thanks for the great work! We decided today that we will be including the block in the release on Monday, so to make sure we make that release we will take over the PR for the fine touches.

I created a new PR based on this one https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/185/files

PR #184, opened 2019-12-17
[PLANET-4624]: Carousel Header: flash of white background in RTL sites

pablocubico:
Blocks using the `blocks_wide` script as the Carousel Header show a "flash of unstyled content", so for a few milliseconds you can see the block displaced a bit to the right when loading a page.

According to [caniuse](https://caniuse.com/#search=calc) it looks like we can use `calc()` to achieve the same result.

Looks like it's been around for a while, so I think it's safe to remove the JS solution. Otherwise, we could still use it as a fallback.

There is also an EN Blocks plugin PR replicating this: https://github.com/greenpeace/planet4-plugin-gutenberg-engagingnetworks/pull/36

This PR also fixes a MENA issue about slides going in the wrong direction, see: https://jira.greenpeace.org/browse/PLANET-4221

The last commit fixes an issue with the Autoplay feature, see:
https://jira.greenpeace.org/browse/PLANET-4654


comzeradd:
Do we also use this on Carousel Header and 2-Split Column blocks?

pablocubico:
@sagarsdeshmukh EN PR added to the description! Some commits added here also.

sagarsdeshmukh:
The autoplay setting tested on `sagardev` instance (https://k8s.p4.greenpeace.org/sagardev/test-carousal-header-3/) looks good 👍

pablocubico:
> Do we also use this on Carousel Header and 2-Split Column blocks?
@comzeradd Yes, tested with Split 2-columns too. 👍

PR #186, opened 2019-12-19
PLANET-4532: Perf: Investigate lazy-loading solution across P4 (incl. all browsers)

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4532

Notes: 

- I included a vanilla JS lazy-loading lib, from a CDN, here and in the `master-theme`. This is what we currently do with things like jQuery. This same lib could be installed using `npm install`, but I couldn't find a way to install it only on the `master-theme`, and get the plugin's JS to load after the `master-theme`'s JS. We can address that separately in the future.

- These changes basically seem to improve some critical path metrics: DOM Content Loaded, First Paint, First meaningful paint..., running on my local instance, these are the results on `develop`:

Captura de pantalla 2019-12-19 a la(s) 11 00 42

Captura de pantalla 2019-12-19 a la(s) 11 01 49

Same audit in this branch:

Captura de pantalla 2019-12-19 a la(s) 10 58 47

Captura de pantalla 2019-12-19 a la(s) 10 59 28

These results can vary a little, as the total number of milliseconds shown below (which are bigger in this branch), but it looks like the FMP (First meaningful paint) event is now happening before the DCL (DOMContentLoaded).

A few more audits with more content will get better results.

### DIFF
--- a/classes/blocks/class-carouselheader.php
+++ b/classes/blocks/class-carouselheader.php
@@ -136,12 +136,14 @@ class CarouselHeader extends Base_Block {
 		// Enqueue js for the frontend.
 		if ( ! $this->is_rest_request() ) {
 			wp_enqueue_script( 'hammer', 'https://cdnjs.cloudflare.com/ajax/libs/hammer.js/2.0.8/hammer.min.js', [], '2.0.8', true );
+			wp_register_script( 'lazyload', 'https://cdnjs.cloudflare.com/ajax/libs/vanilla-lazyload/12.3.0/lazyload.min.js', [], '12.3.0', true );
 			wp_enqueue_script(
 				'carousel-header',
 				P4GBKS_PLUGIN_URL . 'assets/build/carouselHeaderFrontIndex.js',
 				[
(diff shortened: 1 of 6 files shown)


classes/blocks/class-carouselheader.php
```diff
@@ -136,12 +136,14 @@ public function prepare_data( $fields ): array {
 		// Enqueue js for the frontend.
 		if ( ! $this->is_rest_request() ) {
 			wp_enqueue_script( 'hammer', 'https://cdnjs.cloudflare.com/ajax/libs/hammer.js/2.0.8/hammer.min.js', [], '2.0.8', true );
+			wp_register_script( 'lazyload', 'https://cdn.jsdelivr.net/npm/vanilla-lazyload@11.0.4/dist/lazyload.min.js', [], '1.0.4', true );
```
comzeradd: can you do the same CDN change here as in master-theme?

pablocubico: Done!

PR #188, opened 2019-12-20
Planet 4609

Inwerpsel:
* Use a single nested "cards" attribute instead of separate flat attributes.
* Remove the interface for focal points as they don't work yet. Use
plain img element instead.
* Use bootstrap cols to dramatically decrease editor height by placing
fields next to image.
* Use textarea instead for social media message.
* Add button to remove image.
* Use p4ge translation domain everywhere.
* Add some actual help to the help texts.
* Pass title and description to server side renderer.
* Remove gallery naming leftovers from copy.

### DIFF
--- a/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
+++ b/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
@@ -1,9 +1,6 @@
 import { Component, Fragment } from '@wordpress/element';
 import { Preview } from '../../components/Preview';
-
-
 import {
-  BlockControls,
   MediaUpload,
   MediaUploadCheck
 } from '@wordpress/editor';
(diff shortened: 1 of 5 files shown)


assets/src/blocks/SocialMediaCards/SocialMediaCards.js
```diff
@@ -74,52 +79,25 @@ export class SocialMediaCards extends Component {
       <Fragment>
         <div>
           <TextControl
-            label={__('Title', 'planet4-blocks-backend')}
-            placeholder={__('Enter title', 'planet4-blocks-backend')}
-            help={__('Optional', 'planet4-blocks-backend')}
+            label={__('Title', 'p4ge')}
```
sagarsdeshmukh: Hi @Inwerpsel , The `planet4-blocks-backend` is the right text-domain.

sagarsdeshmukh:
⚠️Headsup: If anyone already using the `social-media-cards` block, the current PR changes will overwrite the existing data (we change the block level data structure)

sagarsdeshmukh:
Please remove below focal point instances as well -
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/planet-4609/assets/src/blocks/SocialMediaCards/SocialMediaCardsBlock.js#L73-L77
and 
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/planet-4609/assets/src/blocks/SocialMediaCards/SocialMediaCardsBlock.js#L125

sagarsdeshmukh:
PR looks good to me 💯 , added few small comment 👍

assets/src/blocks/SocialMediaCards/SocialMediaCards.js
Inwerpsel: Also for the frontend? If I search for occurrences of `'p4ge'` in `*.js` files I find over 200 results in the block files, while for only 27 occurrences of `'planet4-blocks-backend'`.

sagarsdeshmukh: For frontend, we have `planet4-blocks`. Yes, we have mix of text-domains in repo. We will fix the remaining in separate PR.

comzeradd: And keep in mind, that "frontend" in this context is definend from a user perspective, not development. It's for the strings that the end users see on the frontend.

Inwerpsel:
> Headsup: If anyone already using the `social-media-cards` block, the current PR changes will overwrite the existing data (we change the block level data structure)

Checked with Luca and the block is not used yet.

templates/blocks/social_media_cards.twig
```diff
@@ -16,31 +16,31 @@
...
+							<img src="{{ card.image_src }}"
+								srcset="{{ card.image_srcset }}" sizes="{{ card.image_sizes }}"
+								alt="{{ card.alt_text }}">
 
 							<div class="share-strip">
-								<a href="https://www.facebook.com/share.php?u={{ single_image.social_url|url_encode|default(post_url) }}&quote={{ single_image.message|url_encode }}"
-								   title="{{ __('Share on Facebook') }}"
-									onclick="dataLayer.push({'event' : 'uaevent', 'eventCategory' : 'Social Share', 'eventAction': 'Facebook', 'eventLabel': '{{ single_image.social_url|default(post_url) }}'});"
+								<a href="https://www.facebook.com/share.php?u={{ card.social_url|default(post_url)|url_encode }}&quote={{ card.message|url_encode }}"
+								   title="{{ __('Share on Facebook', 'planet4-blocks-backend') }}"
```
sagarsdeshmukh: Here, the frontend text-domain(`planet4-blocks`) is applicable. The strings which display on frontend interface to user are mark as frontend text-domain.

classes/blocks/class-socialmediacards.php
```diff
@@ -30,22 +30,34 @@ public function __construct() {
...
+								'image_id'     => [
+									'type' => 'int',
+								],
+								'message'      => [
+									'type' => 'string',
+								],
+								'social_url'   => [
+									'type' => 'string',
+								],
+								'focal_points' => [
```
sagarsdeshmukh: Looks like it left back...

Inwerpsel: I kept it here intentionally so that the backend already regards blocks having that attribute as valid, since we will probably be adding it later.

If it is not specified here and there is a block that has a single card with the attribute, the behavior of wordpress is pretty weird, in that it will not fail but just discard the entire "cards" attribute but preserve the rest of the block.

So if we would at a later point add the focal points but have to revert the code to this state for some reason, then the cards with the `focal_points` attribute will still have a "valid" definition and keep on displaying.

sagarsdeshmukh:
The semicolons are missing at the end of statements in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/188/files#diff-55e26e20bc569dfb36757cb6f5546ca6R86 function. though its optional, but we have it most of our codebase, so I think we should use them to keep consistency.
cc: @comzeradd

Inwerpsel:
> The semicolons are missing at the end of statements in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/188/files#diff-55e26e20bc569dfb36757cb6f5546ca6R86 function. though its optional, but we have it most of our codebase, so I think we should use them to keep consistency.
> cc: @comzeradd

I can follow the consistency argument, but perhaps, as we don't have a hard rule on this yet, it would make sense to prefer not having semicolons? Most widely adopted javascript rulesets prefer not using semicolons (see https://standardjs.com/rules-en.html#semicolons).

comzeradd:
Yes, we use semicolons everywhere. See [eslintrc](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/.eslintrc.json#L21-L23). CI doesn't test for our Blocks JS code at the moment, so maybe we have missed it a few places.

PR #190, opened 2019-12-27
PLANET-4603 Introduce campaign sidebar

Inwerpsel:
This requires the following changes in master-theme: https://github.com/greenpeace/planet4-master-theme/pull/984

* Register a sidebar with campaign settings.
* Add HOC function for creating post-meta controls.
* Use new meta field without underscore for campaign_page_template, as
the underscore prevents API usage. In usages fall back to old field if
the new doesn't exist yet.
* Add ColorPaletteControl that wraps around ColorPalette and functions
similarly to other input controls. This is currently missing in
wordpress, but should it be added there then this can go.

### DIFF
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -1,0 +1,225 @@
+import { PluginSidebar, PluginSidebarMoreMenuItem } from "@wordpress/edit-post";
+import { Component } from '@wordpress/element';
+import { PanelBody, RadioControl, SelectControl } from '@wordpress/components';
+import { compose } from '@wordpress/compose';
+import ColorPaletteControl from '../ColorPaletteControl/ColorPaletteControl';
+import { withPostMeta } from '../PostMeta/withPostMeta';
+import { __ } from '@wordpress/i18n';
+import { fromThemeOptions } from '../fromThemeOptions/fromThemeOptions';
+
+const themeOptions = [
+  {
+    value: '',
+    label: 'Default',
+  },
+  {
+    value: 'antarctic',
+    label: 'Antarctic',
+  },
+  {
+    value: 'arctic',
+    label: 'Arctic',
+  },
+  {
+    value: 'climate',
+    label: 'Climate Emergency',
+  },
+  {
+    value: 'forest',
+    label: 'Forest',
+  },
+  {
+    value: 'oceans',
+    label: 'Oceans',
+  },
+  {
+    value: 'oil',
+    label: 'Oil',
+  },
+  {
+    value: 'plastic',
+    label: 'Plastics',
+  },
+
+];
+
+const ThemeSelect = withPostMeta( SelectControl );
+
+const Select = compose(
+  fromThemeOptions,
+  withPostMeta,
+)( SelectControl );
+
+const Radio = compose(
+  fromThemeOptions,
+  withPostMeta,
+)( RadioControl );
+
+const ColorPalette = compose(
+  fromThemeOptions,
+  withPostMeta,
+)( ColorPaletteControl );
+
+
+export class CampaignSidebar extends Component {
+  static getId() {
+    return 'planet4-campaign-sidebar';
+  }
+
+  static getIcon() {
+    return 'admin-appearance';
+  }
+
+  constructor( props ) {
+    super( props );
+    this.state = {
+      theme: null,
+    };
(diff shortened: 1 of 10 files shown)


pablocubico:
Hey @Inwerpsel this looks great!

I accidentally clicked "Unpin from toolbar" and the sidebar icon disappeared, apparently you have to register a "MoreMenuItem" so it appears in the three-dots "More" menu in the toolbar and you can "pin" it again. Could you add that? I found an example here: https://github.com/WordPress/gutenberg/issues/14211

I would suggest two design tweaks:
1 - just adding 15px padding to the `.components-base-control` wrapper in the template selector, so it matches the rest of the controls there, maybe you can add a specific class name to that element to avoid overriding other controls using that class.
2 - the radio buttons are misaligned, we are probably overriding those elsewhere

Looks amazing! Can't wait to do more changes in this direction.

Inwerpsel:
> Hey @Inwerpsel this looks great!
> 
> I accidentally clicked "Unpin from toolbar" and the sidebar icon disappeared, apparently you have to register a "MoreMenuItem" so it appears in the three-dots "More" menu in the toolbar and you can "pin" it again. Could you add that? I found an example here: [WordPress/gutenberg#14211](https://github.com/WordPress/gutenberg/issues/14211)
> 
> I would suggest two design tweaks:
> 1 - just adding 15px padding to the `.components-base-control` wrapper in the template selector, so it matches the rest of the controls there, maybe you can add a specific class name to that element to avoid overriding other controls using that class.
> 2 - the radio buttons are misaligned, we are probably overriding those elsewhere
> 
> Looks amazing! Can't wait to do more changes in this direction.

Thanks for the feeback @pablocubico ! Maybe you want to check again as I removed the options from withPostMeta and moved them to a separate function that gets these from a template file instead of from an endpoint.

I added the menu link you proposed, indeed we need to have it or there is no way for a user to get the menu back after unpinning. The `PluginSidebar` component also has `isPinned` and `isPinnable` props, unfortunately it is not possible to to set `isPinned` to true and `isPinnable` to false (probably they want to avoid plugins cluttering the editor).

For the 15px margin, does that still apply? Currently it looks like it matches the rest. Maybe your comment is on an earlier version. I had added the `<div class="components-panel__body is-opened">` there manually as the defaults styles of the sidebar only work when putting things in a collapsible panel, but it's a bit tedious for the end user to have everything inside a panel. But maybe we need to write some css so that it always looks good when you include controls outside a panel without having to provide a wrapper diff.

The aligment of the button and the text for the radio buttons looks like a general issue. For example in renderEdit of the SocialMedia block it is also off by the same amount.

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -0,0 +1,225 @@
...
+      } );
+  }
+
+  render() {
+    return (
+      <>
+        <PluginSidebarMoreMenuItem
+          target={ CampaignSidebar.getId() }
+          icon={ CampaignSidebar.getIcon()}>
+          Your Sidebar Plugin
```
pablocubico: I guess this could be "Campaign Options", same as below.

Inwerpsel: Maybe I'll just remove `PluginSidebarMoreMenuItem`, as I added a css fix that removes the unpin button. So if we remove this one too then the user can't unpin it in the first place.

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -0,0 +1,220 @@
+import { PluginSidebar, PluginSidebarMoreMenuItem } from "@wordpress/edit-post";
```
pablocubico: Hey @Inwerpsel this "PluginSidebarMoreMenuItem" is no longer used, BUT, I missed your comment about this, I guess you may think about re-adding it just in case, since the CSS fix for "Unpin from toolbar" could get translated if you set the WP admin to other languages, also it might be a bit too generic (what if other buttons on other toolbars have that label?). I think you can keep both fixes, adding a bit more scope to the CSS thing.

Inwerpsel: Re-added it :+1:

assets/src/components/Sidebar/CampaignSidebar.js
Inwerpsel: In the end I added it again, just in case the css fix would fail as suggested by Pablo

PR #194, opened 2020-01-06
PLANET-4560 Move rules to campaign style / fix spacing

Inwerpsel:
* Remove unneeded button height
* Move background-color to oceans stylesheets
* Remove spacing around cell contents

### DIFF
--- a/templates/blocks/spreadsheet.twig
+++ b/templates/blocks/spreadsheet.twig
@@ -3,27 +3,23 @@
 	{% if ( fields.rows ) %}
 		<div class="block-spreadsheet">
 			<div class="form-inline">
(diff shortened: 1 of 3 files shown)


templates/blocks/spreadsheet.twig
```diff
@@ -11,13 +11,9 @@
 					<tr data-order="{{ rowNumber }}">
 						{% for cell in row %}
 							{% if rowNumber == 0 %}
-								<th title="{{ __('Sort by') }}">
-									{{ cell|escape }}
-								</th>
+								<th title="{{ __('Sort by') }}">{{ cell|escape }}</th>
```
pablocubico: Looks good! I think we should add the `planet4-blocks` domain here in the translator function.

PR #198, opened 2020-01-13
PLANET-3870 Add character counter

Inwerpsel:


assets/src/blocks/Articles/Articles.js
```diff
@@ -3,14 +3,17 @@ import {Preview} from '../../components/Preview';
...
+  TextControl as BaseTextControl,
+  TextareaControl as BaseTextareaControl,
   ServerSideRender
 } from '@wordpress/components';
+import withCharacterCounter from '../../components/withCharacterCounter/withCharacterCounter';
 
 
 const {apiFetch} = wp;
 const {addQueryArgs} = wp.url;
+const TextControl = withCharacterCounter( BaseTextControl );
```
Inwerpsel: I'm wondering if it might be nicer if we move this part to a separate component, so that we can just import that instead of having to call this everywhere. Any thoughts?

PR #204, opened 2020-01-22
PLANET-4561: Make sure P4 Spreadsheet embed (New Block) respects all Campaign Styles

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4561

### DIFF
--- a/assets/src/styles/blocks/Spreadsheet.scss
+++ b/assets/src/styles/blocks/Spreadsheet.scss
@@ -27,9 +27,18 @@
 .spreadsheet-search {
   width: 350px;
   max-width: 100%;
-  margin: 10px 0;
-  padding: 3px 8px;
+  margin: 10px 10px 10px 0;
+  padding: 0 8px;
+  height: 40px;
   border: 1px solid;
+}
+
+.spreadsheet-default-sort {
+  height: 40px !important;
+  box-sizing: border-box;
+  line-height: 1rem !important;
+  width: auto;
+  margin-bottom: 0;
 }
 
 .spreadsheet-sorted-by {
@@ -40,14 +49,6 @@
   &[data-sort-dir="desc"]:after {
     content: "\25B2";
   }
-}
-
-.spreadsheet-default-sort {
-  display: none;
-  border: 1px solid;
-  letter-spacing: .06rem;
-  line-height: 38px;
-  margin-left: 5px;
 }
 
 .spreadsheet-empty-message {
--- a/assets/src/styles/campaigns/themes/_theme_antarctic.scss
+++ b/assets/src/styles/campaigns/themes/_theme_antarctic.scss
@@ -266,4 +266,31 @@ body.theme-antarctic {
   .video-block {
     @include campaign_video($antarctic-gradient);
   }
+
+  .page-template table.spreadsheet-table {
+    color: #fff;
+
+    tr {
+      background: #0A4461;
+      color: #FFFFFF;
+
+      &:nth-child(odd) {
+        background: #52C9F4;
+        color: #0A4461;
+      }
+
+      &:hover {
+        background-color: #074365;
+      }
+
+      &:first-child {
+        background: #06293a;
+        color: #FFFFFF;
+      }
+    }
+
+    .spreadsheet-table-search {
+      border-color: #407ea2;
+    }
+  }
 }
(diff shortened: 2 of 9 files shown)


assets/src/styles/campaigns/themes/_theme_forest.scss
```diff
@@ -519,4 +519,49 @@ body.theme-forest {
...
+      }
+    }
+
+    .spreadsheet-table-search {
+      border-color: #407ea2;
+    }
+  }
+
+  .spreadsheet-default-sort {
+    @include forest-button($white, $forest-dark-green);
```
Inwerpsel: Just wondering, is there a specific reason this is not used on all themes? arctic and oceans are not using is.

pablocubico: Removed these overrides in favor of using `btn-secondary` along with a single override for this class.

assets/src/styles/blocks/Spreadsheet.scss
```diff
@@ -28,10 +28,18 @@
...
   max-width: 100%;
   margin: 10px 0;
-  padding: 3px 8px;
+  padding: 0 8px;
+  height: 40px;
   border: 1px solid;
 }
 
+.spreadsheet-default-sort {
+  height: 40px !important;
```
pablocubico: Not ideal, but this will be removed after a tiny refactor documented in this ticket: https://jira.greenpeace.org/browse/PLANET-4740

Inwerpsel: Will this work on all screen sizes?

pablocubico: Hmmm I'll double check, this should be the same height as the input

pablocubico:
Updated! This is yet another override... I think we should have a better spec of Buttons (if we're developing a design system). It could describe them better, e.g.: https://medium.com/codyhouse/create-your-design-system-part-6-buttons-58e2eda2173e

assets/src/styles/blocks/Spreadsheet.scss
Inwerpsel: Did you verify that this works?

pablocubico: Yes, I actually did a few changes to make it look good on all sizes.

assets/src/styles/campaigns/themes/_theme_antarctic.scss
```diff
@@ -266,4 +266,31 @@ body.theme-antarctic {
   .video-block {
     @include campaign_video($antarctic-gradient);
   }
+
```
Inwerpsel: If we merge https://github.com/greenpeace/planet4-master-theme/pull/1012 first, maybe we can apply this already to these new styles, and use variables in Spreadsheet.scss instead of these theme specific files? Or do you think it's too early for that?

pablocubico: Hmmm I don't think I follow, but I think we can keep the scope of this ticket as is and do incremental refactors as PRs of the refactor ticket.

Inwerpsel: I agree with your point as far as refactors go. But in this case it's not a refactor but new code. So I was just wondering if we could avoid introducing any more campaign theme specific css rules if we already have the css variables code in place. Or is there something else that we don't have yet that blocks us from already doing that?

pablocubico: Sorry, I thought you were talking about the `.video-block` code above this, because that was what Github was showing on the "Conversation" tab.

I guess we could add these color values somewhere in the JSON files and use CSS variables in Spreadsheet.scss.

pablocubico: Nothing is blocking us, is just that this PR is from January 22nd, and the vars weren't there yet.

pablocubico: @Inwerpsel Actually this needs this to happen too: https://github.com/greenpeace/planet4-master-theme/pull/1012/files

pablocubico:
Closing this in favor a fresh PR to merge to rebase and merge the changes from the refactor branch.

New PR: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/236

PR #207, opened 2020-01-24
PLANET-4439 Populate Articles override suggestions with all posts

Inwerpsel:
* Add an endpoint that just returns all published posts and their ids.
If wpml is active this endpoint requires the language of the posts.
* Use that endpoint instead of the `/wp/v2/posts`, which was a huge
overkill in response time and payload size for this use case.
* Fetch the suggestions once when the override field is focused for the
first time, instead of on every input. There is absolutely no need to
perform an API request as this behavior is already handled by the
FormToken component if you provide it with all the options.

### DIFF
--- a/assets/src/blocks/Articles/ArticlesBlock.js
+++ b/assets/src/blocks/Articles/ArticlesBlock.js
@@ -169,7 +169,7 @@ export class ArticlesBlock {
         }
 
         // TO-DO: Check for posts types and posts too...
-        if ((tagsList && tagsList.length === 0) || (postTypesList && postTypesList.length === 0)) {
+        if ((tagsList && tagsList.length === 0) && (postTypesList && postTypesList.length === 0)) {
           return "Populating block's fields...";
         }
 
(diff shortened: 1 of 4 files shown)


assets/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -169,7 +169,7 @@ export class ArticlesBlock {
         }
 
         // TO-DO: Check for posts types and posts too...
-        if ((tagsList && tagsList.length === 0) || (postTypesList && postTypesList.length === 0)) {
+        if ((tagsList && tagsList.length === 0) && (postTypesList && postTypesList.length === 0)) {
```
Inwerpsel: I needed to change this locally or the block would never get passed here, not sure if it could cause any issues in some cases. I couldn't find a case but maybe something to keep an eye on. Could be that this is not needed anymore because `populatePostsToken` is moved to `componentDidMount` instead of the constructor (which react was giving warnings about in the browser console.

PR #208, opened 2020-02-03
PLANET-4710 Use flexbox to align column CTAs

Inwerpsel:
Flex display and justify-content:space-between on the column wrapper,
combined with flex-grow: 1 on only the p tag results in only the p tag
growing to provide the additional spacing and the buttons lining up
neatly at the bottom of the wrapper.

This might need some additional css changes elsewhere, though until now I haven't seen anything breaking because of it.

### DIFF
--- a/assets/src/styles/blocks/Columns.scss
+++ b/assets/src/styles/blocks/Columns.scss
@@ -59,16 +59,13 @@
     }
 
     .column-wrap {
+      display: flex;
+      flex-direction: column;
+      justify-content: space-between;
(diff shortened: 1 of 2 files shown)


pablocubico:
Using flex for both Icons and Image column types (not the Tasks type because it's a pretty different layout).

comzeradd:
Testing on "No Image" style, there is a glitch on resolutions between 768-992.

Screenshot from 2020-03-16 14-03-55

PR #210, opened 2020-02-05
PLANET-4439 Fix endpoint url

Inwerpsel:


Inwerpsel:
Gonna merge this one quickly as the URL used before was non-existent, so this shouldn't be able to break anything.

PR #217, opened 2020-02-14
PLANET-4735 Add dependencies on other fields in theme

Inwerpsel:
* Add a way to specifiy the values of a field based on the current value
of another field in the json file.


assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -74,16 +75,34 @@ export class CampaignSidebar extends Component {
...
+      if (meta) {
+        let theme = meta[ 'theme' ];
+        if ( theme === '' ) {
+          theme = 'default';
+        }
+        if (!isShallowEqual(this.state.meta, meta)) {
+          this.setState( prevState => {
+            return {
+              meta: meta,
+              theme: prevState.theme,
```
Inwerpsel: Normally I would have used `...prevState` here instead of repeating `theme`, however this caused an error which I think is a bug with the `...` operator.

comzeradd:
If this is needed for the next release, let's change the target branch before merge.

PR #229, opened 2020-02-28
PLANET-3427 Save any changes to post meta to the autosave.

Inwerpsel:
* Add custom endpoint that saves meta for the autosave revision of the current
post and user.
* On change of a withPostMeta component send a request to that endpoint.

### DIFF
--- a/classes/rest/class-rest-api.php
+++ b/classes/rest/class-rest-api.php
@@ -5,6 +5,7 @@
 
 namespace P4GBKS\Rest;
 
+use WP_REST_Request;
 use WP_REST_Server;
 
 /**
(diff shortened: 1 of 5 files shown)


classes/rest/class-rest-api.php
```diff
@@ -17,46 +18,97 @@ class Rest_Api {
...
-					]
-				);
-			}
+		add_action( 'rest_api_init', [ __CLASS__, 'endpoints']);
+
+	}
+
+	public static function endpoints() {
+		/**
+		 * A lightweight endpoint to get all posts with only id and title.
```
pablocubico: Is this comment correct? Looks vestigial.

Inwerpsel: It's lightweight in the sense that getting the same 2 fields for all posts from the wordpress API would mean more overhead. This is guaranteed to be fast even when there is a large amount of posts.

PR #234, opened 2020-03-05
PLANET-4804 Make Articles respect post selection order and...

Inwerpsel:
* Pass setAttributes to the component instead of wrapping this
in a function for each inidividual attribute, which caused logic to be
duplicated and split across multiple files.
* Pass attributes as a whole to the component instead of destructuring
it. That way we can just call the server side renderer with `attributes`
instead of having to construct these again from each inidividual
attribute.

### DIFF
--- a/assets/src/blocks/Articles/Articles.js
+++ b/assets/src/blocks/Articles/Articles.js
@@ -20,9 +20,9 @@ export class Articles extends Component {
     super(props);
 
     // Populate tag tokens for saved tags.
-    let tagTokens = props.tagsList.filter(tag => props.tags.includes(tag.id)).map(tag => tag.name);
+    let tagTokens = props.tagsList.filter( tag => props.attributes.tags.includes( tag.id ) ).map( tag => tag.name );
     // Populate post types tokens for saved post types.
-    let postTypeTokens = props.postTypesList.filter(post_type => props.post_types.includes(post_type.id)).map(post_type => post_type.name);
+    let postTypeTokens = props.postTypesList.filter( post_type => props.attributes.post_types.includes( post_type.id ) ).map( post_type => post_type.name );
 
     this.state = {
       tagTokens: tagTokens,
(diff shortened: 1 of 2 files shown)


assets/src/blocks/Articles/Articles.js
```diff
@@ -195,64 +203,70 @@ export class Articles extends Component {
...
                 />
               </div>
               <div>
                 <FormTokenField
-                  value={this.state.tagTokens}
                   suggestions={tagSuggestions}
                   label={__('Select Tags', 'p4ge')}
-                  onChange={tokens => this.onSelectedTagsChange(tokens)}
+                  value={this.state.tagTokens}
+                  onChange={this.onSelectedTagsChange.bind(this)}
```
pablocubico: Please move these bindings to the constructor (already discussed via chat). We can tweak Babel to allow arrow bindings later.

assets/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -173,67 +172,16 @@ export class ArticlesBlock {
...
-
-        function onSelectedPostTypesChange(postTypeIds) {
-          setAttributes({post_types: postTypeIds});
-        }
-
-        function onIgnoreCategoriesChange(value) {
-          setAttributes({ignore_categories: value});
-        }
-
         // We pass down all the attributes to Covers as props using
```
pablocubico: If you see this on time, please remove these old comments!

assets/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -131,7 +131,6 @@ export class ArticlesBlock {
         },
         exclude_post_id: {
           type: 'integer',
-          default: ''
         },
       },
       // withSelect is a "Higher Order Component", it works as
```
pablocubico: This comment is also from and old sample, please remove it while you update this.

PR #239, opened 2020-03-11
PLANET-4827 Add page to import archive files

Inwerpsel:
This should probably be in master-theme instead, but started off here
and it takes a while to move this to master-theme as the it doesn't have
the same controllers. So pushing this one to have it on the server
already.

### DIFF
--- a/classes/controller/menu/class-archive-import.php
+++ b/classes/controller/menu/class-archive-import.php
@@ -1,0 +1,246 @@
+<?php
+/**
+ * An importer for archived content from an XML file.
+ *
+ * @package P4GBKS\Controllers\Menu
+ */
+
+namespace P4GBKS\Controllers\Menu;
+
+use WP_Query;
+
+/**
+ * An importer for archived content from an XML file.
+ *
+ * @package P4GBKS\Controllers\Menu
+ */
+class Archive_Import extends Controller {
+
+	private const WAY_BACK_URL = 'https://wayback.archive-it.org/9650/';
+
+	private const P3_BASE_URL = 'http://p3-raw.greenpeace.org/';
+
+	/**
+	 * Get all urls of P3 that were already imported.
+	 *
+	 * @return string[] All urls of P3 that were already imported.
+	 */
+	private static function get_imported_p3_urls(): array {
+		$query = new WP_Query(
+			[
+				'meta_key'       => 'p3_url',
+				'posts_per_page' => -1,
+			]
+		);
+
+		$urls = [];
+		foreach ( $query->posts as $post ) {
+			$urls[] = get_post_meta( $post->ID, 'p3_url' )[0];
+		}
+
+		return $urls;
+	}
+
+	/**
+	 * Add the admin menu item.
+	 *
+	 * @return void
+	 */
+	public function create_admin_menu() {
+		if ( ! current_user_can( 'manage_options' ) ) {
+			return;
+		}
+		add_submenu_page(
+			'edit.php?post_type=archive',
+			__( 'Archive Import', 'planet4-blocks-backend' ),
+			__( 'Archive Import', 'planet4-blocks-backend' ),
+			'edit_posts',
+			'archive-import',
+			[ $this, 'get_page' ]
+		);
+	}
+
+	/**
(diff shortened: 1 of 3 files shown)


classes/controller/menu/class-archive-import.php
```diff
@@ -0,0 +1,250 @@
...
+		$items = $data->item;
+
+		$total    = count( $items );
+		$errors   = [];
+		$existing = [];
+		$created  = [];
+
+		foreach ( $items as $item ) {
+			$url = self::WAY_BACK_URL . $item->url;
+			if ( $this->does_post_exist( $url, $urls_imported_as_p4_content ) ) {
```
sagarsdeshmukh, line 109: Looks like the duplicate posts(previously imported) are not filtering/skipping with this check -
https://k8s.p4.greenpeace.org/international/?s=jailed+in+Russia+to+take+case+to+European+Court&orderby=_score
https://k8s.p4.greenpeace.org/international/?s=6+ways+corporate+lawsuits+kill+free+speech&orderby=_score
https://k8s.p4.greenpeace.org/international/?s=Global+movement+unites+against+Norwegian+oil+in+the+Arctic&orderby=_score

classes/controller/menu/class-archive-import.php
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * An importer for archived content from an XML file.
+ *
+ * @package P4GBKS\Controllers\Menu
+ */
+class Archive_Import extends Controller {
+
+	private const WAY_BACK_URL = 'https://wayback.archive-it.org/9650/';
+
+	private const P3_BASE_URL = 'http://p3-raw.greenpeace.org/';
```
sagarsdeshmukh, line 21: The fixed p3 link has a excuse for NL(https://github.com/greenpeace/planet4-gpi-export/blob/master/P3%20Archive%20Elastic%20Search%20data/data/netherlands/netherlands-nl_post_list_with_date_final-clean.csv)
We need to handle it with some hardcoded condition.

classes/controller/menu/class-archive-import.php
```diff
@@ -0,0 +1,250 @@
...
+					'numberposts' => -1,
+				]
+			);
+			foreach ( $archives as $archive ) {
+				wp_delete_post( $archive->ID );
+			}
+			return 'Deleted all archived posts';
+		}
+
+		if ( isset( $_GET['done'] ) ) { //phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
sagarsdeshmukh, line 83: I am not sure, when this condition executes? is it a part of XML import?

classes/controller/menu/class-archive-import.php
```diff
@@ -0,0 +1,250 @@
...
+			if ( is_wp_error( $result ) ) {
+				$errors[] = $result;
+			} else {
+				$created[] = $result;
+			}
+		}
+
+		//phpcs:disable
+		echo 'Peak memory usage: ' . $this->readable_bytes( memory_get_peak_usage() );
+		echo "Total: $total<br/>";
```
sagarsdeshmukh, line 152: for consistency, could you please use single quotes for string.

classes/controller/menu/class-archive-import.php
Inwerpsel: Ah you're right, this isn't used anymore, will remove it.

classes/controller/menu/class-archive-import.php
Inwerpsel: Are you ok with taking this up later? It's only for a couple of NROs, so we can start with those where this is not the case.

classes/controller/menu/class-archive-import.php
Inwerpsel: This string is using interpolation so I can't replace it with single quotes.

sagarsdeshmukh:
Few minor comments, other than that PR looks good to me 👍

classes/controller/menu/class-archive-import.php
Inwerpsel: Could be fixed but we might have to follow up on this

PR #241, opened 2020-03-13
PLANET-4842 Carousel: Make swipe scroll more responsive

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-4842


comzeradd:
We already use `swipeup` and `swipedown`, that only capture [vertical scrolling](http://hammerjs.github.io/recognizer-swipe/), not horizontal. And the gestures are only captured inside the Carousel block. So if you swipe below that you should get the regular browser scroll behavior.

PR #247, opened 2020-03-27
PLANET-4820: P4CG - Footer color should not be applied when "Main website navigation" is selected

pablocubico:
Ensuring the right rules and styles are applied.

Master theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1048
Styleguide PR: https://github.com/greenpeace/planet4-styleguide/pull/42
Ref: https://jira.greenpeace.org/browse/PLANET-4820

### DIFF
--- a/assets/src/styles/campaigns/themes/_theme_antarctic.scss
+++ b/assets/src/styles/campaigns/themes/_theme_antarctic.scss
@@ -18,6 +18,7 @@ body.theme-antarctic {
     line-height: 50px;
   }
 
+  font-family: var(--campaign_body_font, $lora);
   background: white;
 
   .btn-primary {
--- a/assets/src/styles/campaigns/themes/_theme_arctic.scss
+++ b/assets/src/styles/campaigns/themes/_theme_arctic.scss
@@ -18,6 +18,7 @@ body.theme-arctic {
     line-height: 50px;
   }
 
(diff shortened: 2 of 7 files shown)


assets/src/styles/campaigns/themes/_theme_antarctic.scss
```diff
@@ -18,6 +18,7 @@ body.theme-antarctic {
     line-height: 50px;
   }
 
+  font-family: var(--campaign_body_font, $lora);
```
Inwerpsel, line 21: Can't we set `lora` through a css variable for the theme instead of in these files? Then the css declaration could be shared among all themes.

Inwerpsel: We could do that here https://github.com/greenpeace/planet4-styleguide/blob/d41bfb0987ba4206129a0609a58deb5afef4d118/src/base/_body.scss#L8

pablocubico: Hey @Inwerpsel  that was my original intention, but as some themes have a different *default* body font (e.g.: Climate has Jost, Forest has Roboto), I want to see the outcome of the P4CG body font ticket first (can't find it right now...).

PR #249, opened 2020-04-08
PLANET-4919: Ensure automatically selected options are correctly shown in the Preview when switching themes

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4919

This check was not working for Palette options, because the value is stored as `color`, and not as `value` in the `options` array (check the comment a few lines above):
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/develop/assets/src/components/PostMeta/withPostMeta.js#L56

Master theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1053

### DIFF
--- a/assets/src/components/ColorPaletteControl/ColorPaletteControl.js
+++ b/assets/src/components/ColorPaletteControl/ColorPaletteControl.js
@@ -7,7 +7,7 @@ import { ColorPalette } from '@wordpress/components';
 function ColorPaletteControl( { label, className, value, help, instanceId, onChange, options = [], ...passThroughProps } ) {
   const id = `inspector-color-palette-control-${ instanceId }`;
 
-  const optionsAsColors = options.map( option => ( { color: option.value } ) );
+  const optionsAsColors = options.map( ( { value, ...props } ) => ( { color: value, ...props } ) );
 
   return !isEmpty( options ) && (
     <BaseControl label={ label } id={ id } help={ help }
--- a/assets/src/components/ColorPaletteControl/ColorPaletteControl.js
+++ b/assets/src/components/ColorPaletteControl/ColorPaletteControl.js
@@ -7,7 +7,7 @@ import { ColorPalette } from '@wordpress/components';
 function ColorPaletteControl( { label, className, value, help, instanceId, onChange, options = [], ...passThroughProps } ) {
   const id = `inspector-color-palette-control-${ instanceId }`;
 
-  const optionsAsColors = options.map( option => ( { color: option.value, name: option.label } ) );
+  const optionsAsColors = options.map( option => ( { color: option.value } ) );
 
   return !isEmpty( options ) && (
     <BaseControl label={ label } id={ id } help={ help }
--- a/assets/src/components/ColorPaletteControl/ColorPaletteControl.js
+++ b/assets/src/components/ColorPaletteControl/ColorPaletteControl.js
@@ -7,13 +7,15 @@ import { ColorPalette } from '@wordpress/components';
 function ColorPaletteControl( { label, className, value, help, instanceId, onChange, options = [], ...passThroughProps } ) {
   const id = `inspector-color-palette-control-${ instanceId }`;
 
+  const optionsAsColors = options.map( option => ( { color: option.value, name: option.label } ) );
+
   return !isEmpty( options ) && (
     <BaseControl label={ label } id={ id } help={ help }
                  className={ classnames( className, 'components-color-palette-control' ) }>
       <ColorPalette
         value={ value }
         onChange={ onChange }
-        colors={ options }
+        colors={ optionsAsColors }
         { ...passThroughProps }
       />
     </BaseControl>
--- a/assets/src/components/PostMeta/withPostMeta.js
+++ b/assets/src/components/PostMeta/withPostMeta.js
@@ -57,6 +57,7 @@ export function withPostMeta( WrappedComponent ) {
         )
       ) {
         value = this.props.defaultValue;
+        this.handleChange( metaKey, value );
       } else {
         value = metaValue;
       }
--- a/assets/src/components/PostMeta/withPostMeta.js
+++ b/assets/src/components/PostMeta/withPostMeta.js
@@ -33,9 +33,6 @@ export function withPostMeta( WrappedComponent ) {
       super( props );
       this.handleChange = this.handleChange.bind( this );
       this.valuePropName = getValuePropName( WrappedComponent );
-      this.state = {
-        value: null
-      };
     }
 
     async handleChange( metaKey, value ) {
@@ -46,19 +43,13 @@ export function withPostMeta( WrappedComponent ) {
       this.props.writeMeta( meta );
     }
 
-    static getDerivedStateFromProps(props, state) {
-      const value = getValueFromProps( props );
-      return {
-        value
-      }
-    }
-
     render() {
       const { metaKey, postMeta, writeMeta, getNewMeta, onChange, defaultValue, ...ownProps } = this.props;
+      const value = getValueFromProps( this.props );
 
       return <WrappedComponent
         { ...{
-          [ this.valuePropName ]: this.state.value,
+          [ this.valuePropName ]: value,
           onChange: ( value ) => {
             this.handleChange( metaKey, value || '' );
             // Fire any onchange event if passed by wrapped component.
--- a/assets/src/components/PostMeta/withPostMeta.js
+++ b/assets/src/components/PostMeta/withPostMeta.js
@@ -38,9 +38,11 @@ export function withPostMeta( WrappedComponent ) {
       };
     }
 
-    handleChange( metaKey, value ) {
-      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => ({ [ metaKey ]: value }));
-      const meta = getNewMeta(metaKey, value, this.props.postMeta );
+    async handleChange( metaKey, value ) {
+      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => {
+        return { [ metaKey ]: value };
+      });
+      const meta = await getNewMeta( metaKey, value, this.props.postMeta );
       this.props.writeMeta( meta );
     }
 
@@ -59,6 +61,7 @@ export function withPostMeta( WrappedComponent ) {
           [ this.valuePropName ]: this.state.value,
           onChange: ( value ) => {
             this.handleChange( metaKey, value || '' );
+            // Fire any onchange event if passed by wrapped component.
             if ( onChange ) {
               onChange( value );
             }
--- a/assets/src/components/PostMeta/withPostMeta.js
+++ b/assets/src/components/PostMeta/withPostMeta.js
@@ -14,6 +14,28 @@ const getValuePropName = ( Component ) => {
 
 const { apiFetch } = wp;
 
+const getValueFromProps = ( props ) => {
+  const { metaKey, postMeta } = props;
+
+  const metaValue = postMeta[ metaKey ];
+
+  let value;
+  // Use the default value if what is stored in the post meta is empty, or if it isn't one of the listed options.
+  if (
+    props.defaultValue
+    && (
+      !metaValue
+      || !props.options.some( option => option.value === metaValue )
+    )
+  ) {
+    value = props.defaultValue;
+  } else {
+    value = metaValue;
+  }
+
+  return value;
+}
+
 export function withPostMeta( WrappedComponent ) {
 
   class WrappingComponent extends Component {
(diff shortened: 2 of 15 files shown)


assets/src/components/ColorPaletteControl/ColorPaletteControl.js
```diff
@@ -7,13 +7,15 @@ import { ColorPalette } from '@wordpress/components';
 function ColorPaletteControl( { label, className, value, help, instanceId, onChange, options = [], ...passThroughProps } ) {
   const id = `inspector-color-palette-control-${ instanceId }`;
 
+  const optionsAsColors = options.map( option => ( { color: option.value } ) );
```
Inwerpsel, line 10: The colors passed to `ColorPalette` can have a label. I hadn't added that to the config files, as I didn't have the names for the colors we use. We may want to use these labels at some point as they improve Accessibility, so maybe you can already pass on this label from the theme here?

pablocubico: So I did this change, but now I'm starting to regret it, as we introduced another thing to maintain (checking that color names match the actual colors, which is quite impossible to test automatically) and this whole UI is far from accessible... also is quite weird to imagine a visually impaired person choosing color palettes (although it's possible). Also, I found that none of the colors in the campaign templates are present in `_colors.scss`, not even the GP greens 🤦 , so I'm making a script to actually collect the colors we are actually using in the whole site. Anyway, it's done, but let's try to keep the scope limited to what the tickets need.

pablocubico:
Hey @Inwerpsel I added this line to the current commit, I think that's all it takes to update the meta on fields that depend on another: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/249/files#diff-d9bfee53567c8981876ce5b3aade0e38R60

@comzeradd watch out: this is not approved yet but Github says it's ok to merge 🤔 

image

assets/src/components/ColorPaletteControl/ColorPaletteControl.js
Inwerpsel: The accessibility concerns are more broad in the sense that any user can benefit from it: likely we will have campaign themes with slight variations of the same color. Including a textual name will make it a lot easier to communicate about them, compared to having only the visual appearance to go by.

Maybe we could also set the text to the hex code if none is present.

assets/src/components/ColorPaletteControl/ColorPaletteControl.js
```diff
@@ -7,13 +7,15 @@ import { ColorPalette } from '@wordpress/components';
 function ColorPaletteControl( { label, className, value, help, instanceId, onChange, options = [], ...passThroughProps } ) {
   const id = `inspector-color-palette-control-${ instanceId }`;
 
+  const optionsAsColors = options.map( option => ( { color: option.value, name: option.label } ) );
```
Inwerpsel, line 10: Maybe an easier solution here is to spread so that we don't need to re-specify the `name` key and are future proof for other fields.
```
const { value, ...otherFields  } = option;
return {
    color: value,
    ...otherFields,
}
```

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -57,6 +57,7 @@ export function withPostMeta( WrappedComponent ) {
         )
       ) {
         value = this.props.defaultValue;
+        this.handleChange( metaKey, value );
```
Inwerpsel, line 60: This definitely should not happen in `render`, as it will result in another render, and in general `render` should not have any side effects, even if they would not trigger a render.

Did you do this to trigger `this.props.writeMeta`or the preview update, or both?

assets/src/components/ColorPaletteControl/ColorPaletteControl.js
pablocubico: I think we are overthinking this a bit, there is no standard for color names, so we'll have to start naming things as "Navy Blue", or "Lime Green". If we don't pass a name, the swatch actually gets the Hex value in the tooltip, so maybe we should just remove the labels, or I could add the Hex color to the label, as in: `Red (#FF0000)`. 🤔

comzeradd: Keep in mind also, that there is some design work pending here too, to vet our full color palette and keep the ones that we actually want to use. So I wouldn't spend much time trying to figure out names for colors that maybe shouldn't even be used.

Inwerpsel: Indeed we should not be setting up the names right now, I was just suggesting to pass on the label already so that adding the labels at a later point becomes an easy configuration only task.

Yeah also noticed that it already shows the hex code. We can re-iterate later if we want it to display differently.

pablocubico:
Hey @Inwerpsel I did some changes in the last commit, mainly:
- removed the call to `handleThemeChange` in the `ThemeSelect` compo, because the themes were being loaded twice, because the `withPostMeta` fields are dispatching `editPost` actions to the store (via `writeMeta`), so the changes are catched by the `wp.data` subscriber on `CampaignSidebar`.
- in the `withPostMeta` HOC, I cached the value of the components in the HOC's state, using `getDerivedStateFromProps` which happens before render, and now I'm calling the `handleChange` method using `componentDidUpdate`, after render, which is a typical combination for handling side effects (lately replaced by the `useEffect` hook. 

I did a few tests and it looks like the preview is catching the right values now, while the saved post stays untouched. Let me know how it looks!

assets/src/components/ColorPaletteControl/ColorPaletteControl.js
pablocubico: Resolving this as we decided to remove the labels from the JSON files.

assets/src/components/PostMeta/withPostMeta.js
pablocubico: Resolving this as it changed and it's outdated now.

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -14,13 +14,38 @@ const getValuePropName = ( Component ) => {
...
+
 export function withPostMeta( WrappedComponent ) {
 
   class WrappingComponent extends Component {
     constructor( props ) {
       super( props );
       this.handleChange = this.handleChange.bind( this );
       this.valuePropName = getValuePropName( WrappedComponent );
+      this.state = {
+        value: getValuePropName( props )
```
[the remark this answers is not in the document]
pablocubico: You're right, the call is wrong, and I think this is actually being overwritten by `getDerivedStateFromProps`. I'll test.

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -14,13 +14,38 @@ const getValuePropName = ( Component ) => {
...
+
 export function withPostMeta( WrappedComponent ) {
 
   class WrappingComponent extends Component {
     constructor( props ) {
       super( props );
       this.handleChange = this.handleChange.bind( this );
       this.valuePropName = getValuePropName( WrappedComponent );
+      this.state = {
+        value: getValueFromProps( props )
```
pablocubico, line 47: Hey @Inwerpsel I think your previous comment here got lost, I updated this to `getValueFromProps` but I'm keeping this in the constructor since otherwise it would trigger a warning: `react-dom.js:539 Warning: `WrappingComponent` uses `getDerivedStateFromProps` but its initial state is undefined. This is not recommended. Instead, define the initial state by assigning an object to `this.state` in the constructor of `WrappingComponent`. This ensures that `getDerivedStateFromProps` arguments have a consistent shape.`

Inwerpsel: You could still assign `null` to `value` so that `getValueFromProps` isn't executed twice?

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -14,13 +14,38 @@ const getValuePropName = ( Component ) => {
...
+      };
     }
 
     handleChange( metaKey, value ) {
```
Inwerpsel, line 51: We can actually adapt the code below that updates the preview meta to use the function I introduced to do the same thing on page load: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/0c211c994b7626a37b5c8b3f886181e24791a84a/assets/src/saveMetaToPreview.js

But maybe it actually would make more sense if we set this up as a listener to the `wp/data` store in some way? Then we wouldn't even need to call `savePreviewMeta` more than once. Though a challenge with that approach is that you need to listen to any event in the `wp/data` store and then only fire if the meta has changed. Or maybe there is a way after all to just listen post meta changes.

Inwerpsel: It would also be an improvement over the current call to `savePreviewMeta` happening on the `DOMContentLoaded` event.

pablocubico:
Hey @Inwerpsel news about this feature:

- I removed the `DOMContentLoaded` subscriber from the `savePreviewMeta` method
- Removed the duplicated coda (the one POSTing to `/save-preview-meta`
- Added the  `savePreviewMeta` here: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/249/files#diff-0fe7b281274ce46f740cd4828efc6356R102 , where we are setting the `meta` for the CampaignSidebar.

The call to `getDerivedStateFromProps` is still there as I'm still relying on the cached state value to call `handleChange` which writes the meta. But so far it looks good, maybe we can wrap it up if all these works, I'm currently testing it.

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -14,56 +14,66 @@ const getValuePropName = ( Component ) => {
...
-        this.props.defaultValue
-        && (
-          !metaValue
-          || !this.props.options.some( option => option.value === metaValue )
-        )
-      ) {
-        value = this.props.defaultValue;
-      } else {
-        value = metaValue;
+      if (metaValue != this.state.value) {
```
Inwerpsel, line 66: Can you make this a strict comparison? `!==`

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -85,19 +86,20 @@ export class CampaignSidebar extends Component {
...
       const meta = wp.data.select( 'core/editor' ).getEditedPostAttribute( 'meta' );
+
       if ( meta ) {
         let theme = meta[ 'theme' ];
         if ( theme === '' ) {
           theme = 'default';
         }
         if ( !isShallowEqual( this.state.meta, meta ) ) {
           this.setState( { meta: meta } );
+          savePreviewMeta();
```
Inwerpsel, line 102: Be aware that `CampaignSidebar` will only mount if the icon is clicked! I think we might also need to update this preview if the sidebar isn't opened.

pablocubico: Hmmm good point, so maybe we need to call it on `DOMContentLoaded` after all. Or maybe there is a better editor hook for that. I'll dig into it.

pablocubico: I just checked: looks like the component is mounted even if the Sidebar is not displayed, so `savePreviewMeta` is called on load.

assets/src/components/PostMeta/withPostMeta.js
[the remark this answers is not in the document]
pablocubico: Resolving as this is outdated.

assets/src/components/PostMeta/withPostMeta.js
pablocubico: This got outdated too.

assets/src/components/Sidebar/CampaignSidebar.js
pablocubico: This is not relevant anymore.

pablocubico:
Hey @Inwerpsel , I merged your branch, squashed some of the commits and rebased to master.

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -21,51 +33,35 @@ export function withPostMeta( WrappedComponent ) {
       super( props );
       this.handleChange = this.handleChange.bind( this );
       this.valuePropName = getValuePropName( WrappedComponent );
+      this.state = {
+        value: null
```
Inwerpsel, line 37: Still not sure if we need to introduce state and `getDerivedStateFromProps` in the component to do handle translating the props into a value. What is the benefit over calling `getValueFromProps` in render?

pablocubico: This is not needed anymore as you removed the `componentDidUpdate` method in your previous commit.

PR #251, opened 2020-04-10
PLANET-4924 Re-enable the native buttons block

sagarsdeshmukh:
The button block is updated in WP-5.4 release with "buttons" block. The new buttons block offers additional options to editors -
- Allow to set custom button border radius (finding a way to disable this option...)
- New button style options(we disabled it and allow P4 buttons styles only)
- Color gradient option for button background
- Toggle option to "Open in new tab" (this options seems useful)

Task -
- Disable border radius option
- Allow only the colours mentioned in Will's comment for colour palette( [JIRA 4924](https://jira.greenpeace.org/browse/PLANET-4924) )
- Ensure that it still follows our custom styles
- Whitelist buttons block
- Disable the gradient presets & custom gradients

The WP core buttons block,wont have a build in setting to disable the  border radius option and update custom text and background color. The issue regarding build in support to control button block is open( [Issue #19796](https://github.com/WordPress/gutenberg/issues/19796)  ) 

The approach I followed here to fix button block is, Override core button block edit method with our requirement. (ref. [Issue #19796 comment](https://github.com/WordPress/gutenberg/issues/19796#issuecomment-608062092))

Eg. https://k8s.p4.greenpeace.org/test-titan/test-button-block/

Styleguide PR: [greenpeace/planet4-styleguide#47](https://github.com/greenpeace/planet4-styleguide/pull/47)

### DIFF
--- a/assets/src/components/p4_button/edit.js
+++ b/assets/src/components/p4_button/edit.js
@@ -1,0 +1,282 @@
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+/**
+ * WordPress dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { useCallback, useState } from '@wordpress/element';
+import { compose } from '@wordpress/compose';
+import {
+	KeyboardShortcuts,
+	PanelBody,
+	RangeControl,
+	TextControl,
+	ToggleControl,
+	withFallbackStyles,
+	ToolbarButton,
+	ToolbarGroup,
+	Popover,
+} from '@wordpress/components';
+import {
+	BlockControls,
+	__experimentalUseGradient,
+	ContrastChecker,
+	InspectorControls,
+	__experimentalPanelColorGradientSettings as PanelColorGradientSettings,
+	RichText,
+	withColors,
+	__experimentalLinkControl as LinkControl,
+} from '@wordpress/block-editor';
+import { rawShortcut, displayShortcut } from '@wordpress/keycodes';
+import { link } from '@wordpress/icons';
+
+const { getComputedStyle } = window;
+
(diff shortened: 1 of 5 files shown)


classes/class-loader.php
```diff
@@ -570,6 +570,10 @@ public function set_color_palette() {
 
 		// Disable custom color option.
 		add_theme_support( 'disable-custom-colors' );
+
+		// Disable gradient presets & custom gradients.
+		add_theme_support( 'editor-gradient-presets', [] );
```
sagarsdeshmukh, line 575: Note: Here I disabled the `gradient presets` & `custom gradients` globally.

assets/src/components/p4_button/edit.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * This file is copy of core button block edit.js (https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/button/edit.js), with customize changes.
```
Inwerpsel, line 2: Can you update this link to point to the commit of this file you started from? This currently points to that file in the latest commit of the gutenberg repo, and it looks like it is different from the one you based on (has newer changes I believe).

assets/src/components/p4_button/edit.js
```diff
@@ -57,31 +75,6 @@ const applyFallbackStyles = withFallbackStyles( ( node, ownProps ) => {
 } );
 
 const NEW_TAB_REL = 'noreferrer noopener';
-const MIN_BORDER_RADIUS_VALUE = 0;
```
Inwerpsel: Maybe it's better to keep this part in, as we already removed its usage. Keeping it will make it a lot easier to keep this file up to date with whatever changes happen in Gutenberg.

PR #254, opened 2020-04-14
PLANET-4913 Fix spreadsheet block size and scrolling

Inwerpsel:
* Apply sticky positioning to the header.
* Apply a max height for different screen sizes.
* Cache the sheet response for 5 minutes.
* Use `<thead>` and `<tbody>`. For now this assumes the first row is
always a header, we might want to make that configurable.
* Remove the default sort button.
* Fix alternating colors when the rows are filtered by using a different
technique. Instead of hiding the elements, remove them and add them back
when needed.
* Fix sorting not applied after filtering.

### DIFF
--- a/classes/blocks/class-spreadsheet.php
+++ b/classes/blocks/class-spreadsheet.php
@@ -17,7 +17,9 @@ class Spreadsheet extends Base_Block {
 	 */
 	public const BLOCK_NAME = 'spreadsheet';
 
-	private const MAX_ROWS = 1000;
+	private const MAX_ROWS = 10000;
+
+	private const CACHE_LIFETIME = 30;
 
 	/**
 	 * SpreadsheetTable constructor.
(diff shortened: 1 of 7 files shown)


public/js/spreadsheet.js
```diff
@@ -29,37 +51,32 @@ document.addEventListener( 'DOMContentLoaded', () => {
...
-        headerCell.dataset.sortDir = newDir;
-        if ( newDir === 'desc' ) {
-          sortedRows = sortedRows.reverse();
+        // If text is the same preserve the original sort order.
+        return rowA.dataset.order - rowB.dataset.order;
         }
-        table.append( ...sortedRows );
-      } );
-    } );
+      ); const newDir = prevDir !== 'asc' ? 'asc' : 'desc';
```
pablocubico, line 69: There seems to be a indentation issue in line 68 (ESLint complained), and I guess this line could probably be in a new line. This is totally a nit, feel free to ignore.

public/js/spreadsheet.js
```diff
@@ -1,26 +1,48 @@
...
+    let previousSearchTerm;
 
-    searchInput.addEventListener( 'input', () => {
-      let searchTerm = searchInput.value;
+    const filterRows = ( event ) => {
       let hasMatches = false;
+
+      const searchTerm = event.target.value.toLowerCase().trim();
+
+      previousSearchTerm = searchTerm;
```
pablocubico, line 30: It looks like this is not used anywhere, right? I didn't notice at first, ESLint complained about the definition in line 23 not being used.

pablocubico:
Hey @Inwerpsel  looks great! Love that you use vanilla code. Just added a few very minor comments.

public/js/spreadsheet.js
Inwerpsel: Yes, this was a leftover from a previous approach, removed it now

classes/blocks/class-spreadsheet.php
```diff
@@ -51,34 +51,13 @@ public function prepare_data( $fields ): array {
...
-					// phpcs:ignore Squiz.PHP.DisallowSizeFunctionsInLoops
-					&& count( $rows ) <= self::MAX_ROWS
-				) {
-					$rows[] = $data;
-				}
-				// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_fclose
-				fclose( $handle );
-				$fields['rows'] = $rows;
-			}
+			$sheet = $this->get_sheet( $id, $fields['skip_cache'] ?? true );
```
Inwerpsel, line 55: I had added the cache but then realized that if we don't also visualize in the editor that there is a 5 minute delay, or offer a way to invalidate the cache, then this will be confusing to editors setting up the block. So I made it default to true if the block doesn't have the `skip_cache` field so that we can still preserve the logic.

PR #258, opened 2020-04-22
PLANET-4878 Add 'pre-publish-checklist' panel on click of publish button

sagarsdeshmukh:
[JIRA 4878](https://jira.greenpeace.org/browse/PLANET-4878)

Added a [PluginPrePublishPanel](https://developer.wordpress.org/block-editor/developers/slotfills/plugin-pre-publish-panel/) panel with validation message of mandatory fields.

image

### DIFF
--- a/assets/src/BlockEditorValidation.js
+++ b/assets/src/BlockEditorValidation.js
@@ -1,0 +1,64 @@
+const { registerPlugin } = wp.plugins;
+const { PluginPrePublishPanel } = wp.editPost;
+const { dispatch } = wp.data;
+const {__} = wp.i18n;
+
+export const blockEditorValidation = () => {
+  registerPlugin( 'pre-publish-checklist', { render: PrePublishCheckList } );
+};
+
+const isValid = element => {
+  // Apply validation only for campaign post types.
+  if ('campaign' === $('#post_type').val() && 'required' === $(element).data('validation')) {
+    if (!$(element).val() || 'not set' === $(element).val()) {
+      return false;
+    }
+  }
+
+  return true;
+};
+
+const PrePublishCheckList = () => {
+  let checkListMsg = [];
+
+  // Filter CMB2 fields that have the 'data-validation' attribute set to 'required'.
+  const elements = Array.from( document.querySelectorAll( '[data-validation]' ) );
+  const invalidElements = elements.filter( element => !isValid( element ) );
+  elements.forEach( element => { element.classList.remove( 'cmb2_required_field_error') } );
+  invalidElements.forEach( element => { element.classList.toggle( 'cmb2_required_field_error') } );
+  const postIsValid = invalidElements.length > 0
+
+  if ( postIsValid ) {
+    // Open "Analytics & Tracking" fields metabox, if closed.
+    $('#p4_campaign_fields').removeClass('closed');
+    checkListMsg.push( __( 'Please check "Analytics & Tracking" section for required fields.', 'planet4-master-theme-backend' ) );
+    invalidElements.forEach( element => { checkListMsg.push( ' - ' + $(element).parent().find('label').text() + ' is a required field' ) } );
+  }
+
+  let classname = '';
+  if ( postIsValid ) {
+    dispatch( 'core/editor' ).lockPostSaving();
+    classname = 'p4-plugin-pre-publish-panel-error';
+  } else {
+    dispatch( 'core/editor' ).unlockPostSaving();
+    checkListMsg.push( __( 'All good.', 'planet4-master-theme-backend' ) );
+  }
+
+  return (
+    <PluginPrePublishPanel
+      title={ __( 'Publish Checklist', 'planet4-master-theme-backend' ) }
+      initialOpen="true"
+      className={ classname }
+      icon="none">
+        <ul>
+          {checkListMsg.map((msg, index) => {
+            return (
+              <li key={index} >
+                <p>{msg}</p>
+              </li>
+            );
+          })}
+        </ul>
+    </PluginPrePublishPanel>
+  )
+};
--- a/assets/src/editorIndex.js
+++ b/assets/src/editorIndex.js
@@ -22,6 +22,7 @@ import { addButtonLinkPasteWarning } from './addButtonLinkPasteWarning';
 import { setupCustomSidebar } from "./setupCustomSidebar";
 import { setUpCssVariables } from './connectCssVariables';
 import { SubPagesBlock } from './blocks/SubPages/SubPagesBlock';
+import { blockEditorValidation } from './BlockEditorValidation';
 
 new ArticlesBlock();
 new CarouselHeaderBlock();
@@ -48,3 +49,4 @@ addButtonLinkPasteWarning();
 replaceTaxonomyTermSelectors();
 setupCustomSidebar();
 setUpCssVariables();
+blockEditorValidation();
--- a/assets/src/styles/editorOverrides.scss
+++ b/assets/src/styles/editorOverrides.scss
@@ -202,3 +202,11 @@ input.describe[type=text][data-setting=caption] {
     }
   }
 }
+
+.p4-plugin-pre-publish-panel-error {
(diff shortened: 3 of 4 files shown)


assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,96 @@
...
+  // "Analytics & Tracking" fields validation.
+  let $htmlbody   = $( 'html, body' );
+  let $toValidate = $( '[data-validation]' );
+
+  let analyticsErrorFlag   = false;
+
+  if ( $toValidate.length ) {
+    let $row = null;
+
+    function add_required($row) {
```
Inwerpsel, line 25: Can you move these function declarations outside of `PrePublishCheckList`? That will make it easier to read `PrePublishCheckList`.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,96 @@
...
+      $row.find('select').css({'border-color': 'red'});
```
Inwerpsel, line 26: Can this style be set with a class instead? That way it's easier to add multiple css rules for invalid rows without having to also remove each rule individually.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,96 @@
...
+        if ($row.is('.cmb-type-file-list')) {
+
+          const has_LIs = $row.find('ul.cmb-attach-list li').length > 0;
+
+          if (!has_LIs) {
+            add_required($row);
+            analyticsErrorFlag = true;
+          } else {
+            remove_required($row);
+            analyticsErrorFlag = false;
```
Inwerpsel, line 53: Probably`analyticsErrorFlag` shouldn't be changed in this case? Else having 2 fields, where the first is invalid and the second is valid would result in the `analyticsErrorFlag` being false.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,96 @@
...
+          }
+        }
+      }
+    });
+
+    if ( analyticsErrorFlag ) {
+      lockPost = analyticsErrorFlag;
```
Inwerpsel, line 69: Do we need the `lockPost` variable? There's only one usage and looks like it could just as well use `analyticsErrorFlag` there instead.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,96 @@
...
+      const p4_postbox = $htmlbody.find('#p4_campaign_fields');
+      if (p4_postbox.hasClass('closed')) {
+        p4_postbox.removeClass('closed');
+      }
+      checkListMsg += __( 'Please check "Analytics & Tracking" section for required fields.', 'planet4-master-theme-backend' );
+      classname = 'p4-plugin-pre-publish-panel-error';
+    }
+  }
+
+  if ( lockPost === true ) {
```
Inwerpsel, line 80: If we use `analyticsErrorFlag`, then it's guaranteed to be a boolean, so it could just be `if ( analyticsErrorFlag ) {`

assets/src/BlockEditorValidation.js
sagarsdeshmukh: True, but my thought behind the `lockPost` variable is for adding a multiple validation rules and keep single lockPost flag. (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/258/files#diff-c391b7e90009d8476d32127fbf7c1ad8R71)

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,93 @@
...
+
+  let checkListMsg = '';
+  let classname    = '';
+
+  // "Analytics & Tracking" fields validation.
+  let $htmlbody   = $( 'html, body' );
+  // Filter CMB2 fields that have the 'data-validation' attribute set to 'required'.
+  let $toValidate = $( '[data-validation]' );
+
+  let analyticsErrorFlag   = false;
```
Inwerpsel, line 21: This should probably be something more generic than `analyticsErrorFlag` as other fields could be validated too?

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      // Apply validation only for campaign post types.
+      if ('campaign' === $('#post_type').val() && 'required' === $this.data('validation')) {
+        if ($row.is('.cmb-type-file-list')) {
+
+          const has_LIs = $row.find('ul.cmb-attach-list li').length > 0;
+
+          if (!has_LIs) {
+            $row.find('select').addClass('cmb2_required_field_error');
```
Inwerpsel, line 42: You can extract whether the row is valid into a const before changing the class. That way you can do that with a single call to `toggleClass` (passing the const) instead of 4 calls to `addClass` and `removeClass`.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,93 @@
...
+          }
+        }
+      }
+    });
+
+    if ( analyticsErrorFlag ) {
+      lockPost = analyticsErrorFlag;
+      // Open "Analytics & Tracking" fields postbox, if closed.
+      const p4_postbox = $htmlbody.find('#p4_campaign_fields');
+      if (p4_postbox.hasClass('closed')) {
```
Inwerpsel, line 63: We can call `removeClass` without checking if it already has the class.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,93 @@
...
+        p4_postbox.removeClass('closed');
+      }
+      checkListMsg += __( 'Please check "Analytics & Tracking" section for required fields.', 'planet4-master-theme-backend' );
```
Inwerpsel, line 66: Ah hadn't noticed it so this must be why it now has the specific `analyticsErrorFlag` variable. But maybe here it would make more sense to list all invalid fields instead of only the section? 

So that editors instantly see all that is wrong with the post, without having to navigate to the analytics section.

Then we wouldn't need to have a variable for each section and it will be easier to add other fields.

sagarsdeshmukh: Done.
image
Please have a look.

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,64 @@
...
+  const invalidElements = elements.filter( element => !isValid( element ) );
+  elements.forEach( element => { element.classList.remove( 'cmb2_required_field_error') } );
+  invalidElements.forEach( element => { element.classList.toggle( 'cmb2_required_field_error') } );
+  const postIsValid = invalidElements.length > 0
+
+  if ( postIsValid ) {
+    // Open "Analytics & Tracking" fields metabox, if closed.
+    $('#p4_campaign_fields').removeClass('closed');
+    checkListMsg.push( __( 'Please check "Analytics & Tracking" section for required fields.', 'planet4-master-theme-backend' ) );
+    invalidElements.forEach( element => { checkListMsg.push( ' - ' + $(element).parent().find('label').text() + ' is a required field' ) } );
```
Inwerpsel, line 35: Why is it prefixed with a dash?

sagarsdeshmukh: Just to display the field list like a bullet list, -
image

assets/src/BlockEditorValidation.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const PrePublishCheckList = () => {
+  let checkListMsg = '';
+
+  // Filter CMB2 fields that have the 'data-validation' attribute set to 'required'.
+  const elements = Array.from( document.querySelectorAll( '[data-validation]' ) );
+  const invalidElements = elements.filter( element => !isValid( element ) );
+  elements.forEach( element => { element.classList.remove( 'cmb2_required_field_error') } );
+  invalidElements.forEach( element => { element.classList.toggle( 'cmb2_required_field_error') } );
```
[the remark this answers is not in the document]
sagarsdeshmukh: here we need `invalidElements` to get the field names to show in error message.

PR #259, opened 2020-04-23
PLANET-4862 Remove nonce for Articles load more

Inwerpsel:
https://jira.greenpeace.org/browse/PLANET-4862

* The nonce-verification offers no additional security, as a new nonce
can be fetched from any page with an articles block without
authentication. Possibly it could also cause the load more button to
stop working.
* Remove some event handlers in search.js that are attached to classes
which don't occur in the code anymore.

### DIFF
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -216,55 +216,54 @@ class Articles extends Base_Block {
 	public function load_more() {
 
 		// If this is an ajax call.
-		if ( wp_doing_ajax() ) {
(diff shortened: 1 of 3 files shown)


sagarsdeshmukh:
The PHP side of code looks good to me, @pablocubico - could you please have a look on `load_more.js` file once.

pablocubico:
@sagarsdeshmukh @Inwerpsel Looks good to me, CI is complaining a bit but it seems to be unrelated.

PR #260, opened 2020-04-24
Find all fields that depend on a field and update them with defaults if

Inwerpsel:
their values are not allowed anymore.

### DIFF
--- a/assets/src/components/PostMeta/withPostMeta.js
+++ b/assets/src/components/PostMeta/withPostMeta.js
@@ -38,9 +38,11 @@ export function withPostMeta( WrappedComponent ) {
       };
     }
 
-    handleChange( metaKey, value ) {
-      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => ({ [ metaKey ]: value }));
-      const meta = getNewMeta(metaKey, value, this.props.postMeta );
+    async handleChange( metaKey, value ) {
+      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => {
+        return { [ metaKey ]: value };
+      });
+      const meta = await getNewMeta( metaKey, value, this.props.postMeta );
       this.props.writeMeta( meta );
     }
 
@@ -59,6 +61,7 @@ export function withPostMeta( WrappedComponent ) {
           [ this.valuePropName ]: this.state.value,
           onChange: ( value ) => {
             this.handleChange( metaKey, value || '' );
+            // Fire any onchange event if passed by wrapped component.
             if ( onChange ) {
               onChange( value );
             }
--- a/assets/src/components/PostMeta/withPostMeta.js
+++ b/assets/src/components/PostMeta/withPostMeta.js
@@ -12,28 +12,18 @@ const getValuePropName = ( Component ) => {
   return 'value';
 };
 
-const { apiFetch } = wp;
-
 const getValueFromProps = ( props ) => {
-  const { metaKey, postMeta } = props;
+  const { metaKey, postMeta, defaultValue } = props;
 
   const metaValue = postMeta[ metaKey ];
 
-  let value;
-  // Use the default value if what is stored in the post meta is empty, or if it isn't one of the listed options.
-  if (
-    props.defaultValue
+  const shouldUseDefault = defaultValue
     && (
       !metaValue
       || !props.options.some( option => option.value === metaValue )
-    )
-  ) {
-    value = props.defaultValue;
-  } else {
-    value = metaValue;
-  }
+    );
+  return shouldUseDefault ? defaultValue : metaValue;
 
-  return value;
 }
 
 export function withPostMeta( WrappedComponent ) {
@@ -49,7 +39,9 @@ export function withPostMeta( WrappedComponent ) {
     }
 
     handleChange( metaKey, value ) {
-      this.props.writeMeta( metaKey, value );
+      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => ({ [ metaKey ]: value }));
+      const meta = getNewMeta(metaKey, value, this.props.postMeta );
+      this.props.writeMeta( meta );
     }
 
     static getDerivedStateFromProps(props, state) {
@@ -59,17 +51,8 @@ export function withPostMeta( WrappedComponent ) {
       }
     }
 
-    componentDidUpdate() {
-      const { metaKey, postMeta } = this.props;
-      const metaValue = postMeta[ metaKey ];
-
-      if (metaValue !== this.state.value) {
-        this.handleChange( metaKey, this.state.value );
-      }
-    }
-
     render() {
-      const { metaKey, postMeta, writeMeta, onChange, ...ownProps } = this.props;
+      const { metaKey, postMeta, writeMeta, getNewMeta, onChange, defaultValue, ...ownProps } = this.props;
 
       return <WrappedComponent
         { ...{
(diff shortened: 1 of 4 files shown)


assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -59,17 +51,8 @@ export function withPostMeta( WrappedComponent ) {
...
-      const metaValue = postMeta[ metaKey ];
-
-      if (metaValue !== this.state.value) {
-        this.handleChange( metaKey, this.state.value );
-      }
-    }
-
     render() {
-      const { metaKey, postMeta, writeMeta, onChange, ...ownProps } = this.props;
+      const { metaKey, postMeta, writeMeta, getNewMeta, onChange, defaultValue, ...ownProps } = this.props;
```
Inwerpsel, line 55: Added this `defaultValue` props so that it's not in `ownProps`. That was causing issues as React uses the same prop name itself and was causing following warning: `Warning: Select elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both). Decide between using a controlled or uncontrolled select element and remove one of these props. More info: https://fb.me/react-controlled-components`

pablocubico: Right, I saw that warning, I think this change will cause the `withDefaultLabel` HOC to stop working as it relies on the `defaultValue` prop, maybe we can use a different name for the prop, I tried that but it causes an "Unknown Prop Warning", also quite common. So maybe we need to review how that works and make sure we don't use `defaultValue` at all anywhere.

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -86,6 +86,30 @@ export class CampaignSidebar extends Component {
...
       parent: null,
     };
+    this.handleThemeSwitch = this.handleThemeSwitch.bind( this );
+    this.loadTheme = this.loadTheme.bind( this );
+  }
+
+  async handleThemeSwitch( metaKey, value, meta ) {
+    await this.loadTheme( value )
+    const invalidatedFields = this.state.theme.fields.filter( field => {
+      const resolvedField = getFieldFromTheme(this.state.theme, field.id, meta);
```
Inwerpsel, line 96: We have a naming collision here, so for now called it `resolvedField`, but probably can be named better.

assets/src/components/PostMeta/withPostMeta.js
pablocubico: Possible hack: React has a feature where using a non-standard lowercase prop (e.g.: `themedefault`) won't trigger the "Unknown prop in DOM component" warning. But we probably want to avoid that also.

assets/src/components/fromThemeOptions/fromThemeOptions.js
```diff
@@ -97,7 +99,9 @@ export function fromThemeOptions( WrappedComponent ) {
         return <WrappedComponent { ...ownProps }/>
       }
 
-      const field = getFieldFromTheme( theme, ownProps.metaKey );
+      const meta = wp.data.select( 'core/editor' ).getEditedPostAttribute( 'meta' );
```
pablocubico, line 102: Is this needed here? I see the logic for merging the meta and the theme options is spread into a few parts, for example: we should possibly unify `withPostMeta` and `fromThemeOptions`, they look pretty dependant on each other.  

With this change we are retrieving the `meta` field in the `render` method of `fromThemeOptions`, a field should probably only need to know about its own value and not much else. So far, all the fields that use `fromThemeOptions` actually use `withPostMeta` in conjunction. Maybe you can use the `postMeta` prop in this HOC directly, since `compose` will pass the values right-to-left, and move the static methods from the outside to inside the classes, to avoid passing down the `meta` attribute to all of them?. I know `getFieldFromTheme` is exported separately and it calls `resolveDependency`, so I guess that will require some decoupling to do. Or some coupling actually, because now the `fromThemeOptions` file looks more like a set of utility function which rely heavily on the `meta` field. 

Anyway, I'm not sure, I think this probably needs to happen in some sort of `meta` handler on top of Campaign Sidebar and not in these HOCs which are used to decorate fields.

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -86,6 +86,30 @@ export class CampaignSidebar extends Component {
...
+
+      if ( !resolvedField || !resolvedField.options ) {
+        return !!meta[ field.id ];
+      }
+
+      return !(resolvedField.options.some( option => option.value === meta[ field.id ]) );
```
pablocubico, line 102: This logic seems to be repeated below, is there a way to encapsulate it better, or make sure it only happens in one place? There also a similar check in `withPostMeta` (not shown in this diff).

assets/src/components/PostMeta/withPostMeta.js
```diff
@@ -38,9 +38,11 @@ export function withPostMeta( WrappedComponent ) {
       };
     }
 
-    handleChange( metaKey, value ) {
-      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => ({ [ metaKey ]: value }));
-      const meta = getNewMeta(metaKey, value, this.props.postMeta );
+    async handleChange( metaKey, value ) {
+      const getNewMeta = this.props.getNewMeta || (( metaKey, value, meta ) => {
```
pablocubico, line 42: I found it a bit hard to guess what's happening with `getNewMeta`,  it seems that there are a few flavors of it: this "default" definition if the prop is not set, then there is a `provideNewMeta` def in `fromThemeOptions`, and it also seems to be overriden by `handleThemeSwitch` in the `ThemeSelect` compo, I think it makes it a bit hard to see which version the component is actually going to get at the end.

assets/src/components/fromThemeOptions/fromThemeOptions.js
Inwerpsel: > With this change we are retrieving the meta field in the render method of fromThemeOptions, a field should probably only need to know about its own value and not much else.

A field needs to know other values than its own from the meta, as it could depend on another field. The idea of the `fromThemeOptions` HOC is that it encapsulates this logic for reading from the theme. Technically would be cleaner to add `meta` as a prop, but that would involve having to provide the prop everywhere. We could instead use `withSelect` instead like in `withPostMeta`. I considered that, but it didn't seem worth the complexity at the moment, since we're only reading from the meta.

> Is this needed here? 
I just moved it up from here to have it called only once: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/260/commits/589dee85fca1b47e3866934aaf376b9d6a6e6815#diff-e9a3f238758823bef04f80b5c8f492c1L28

Yeah currently `fromThemeOptions` and `withPostMeta` are somewhat coupled, though it is just `withThemeOptions` that depends on `withPostMeta`, not the other way round. Currently it's already used on its own for the theme dropdown, but I can imagine re-using it for other UI where we store things in `postMeta`, not necessarily related to the theme.

Maybe we could already decouple the theme from the meta storage too, on the other hand we already have 2 HOCs, so it might be more straightforward to include the coupling in `fromThemeOptions`, if we're not planning to use `fromThemeOptions` with other types of storage.

assets/src/components/PostMeta/withPostMeta.js
Inwerpsel: Yeah I re-added it to not break `withDefaultLabel`. The "Unknown Prop Warning" is because the prop isn't being removed with destructuring, but called with `this.props.defaultValue`. So `ownProps` is passed with an unknown attribute to react. If we rename it both issues are fixed.

assets/src/components/fromThemeOptions/fromThemeOptions.js
pablocubico: > A field needs to know other values than its own from the meta, as it could depend on another field. The idea of the fromThemeOptions HOC is that it encapsulates this logic for reading from the theme.

Hmmm not sure about that, personally, I would make the CampaignSidebar smarter and the fields dumber, so maybe instead of supercharging the fields with `withPostMeta` and `fromThemeOptions`, I would move all the options & dependencies logic to the CampaignSidebar or even something on top of it.

Anyway, this looks good, I'm doing a round of manual tests locally just in case. 💯

PR #267, opened 2020-05-06
Planet 5037 add css variables attribute

Inwerpsel:
Adds the backend support for using css variables inside blocks. Also setup the Spreadsheet block edit component to put the color into the attribute that is used to store the variables.

### DIFF
--- a/assets/src/blocks/Spreadsheet/Spreadsheet.js
+++ b/assets/src/blocks/Spreadsheet/Spreadsheet.js
@@ -16,7 +16,11 @@ export class Spreadsheet extends Component {
   renderEdit() {
     const { __ } = wp.i18n;
 
-    const { color } = this.props;
+    const { attributes, setAttributes } = this.props;
+
+    const toAttribute = attributeName => value => {
+      setAttributes( { [ attributeName ]: value } );
+    };
 
     const colors = [
       { name: 'blue', color: '#C9E7FA' },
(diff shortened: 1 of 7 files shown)


assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -16,22 +16,38 @@ export class Spreadsheet extends Component {
...
+
     return (
       <Fragment>
         <InspectorControls>
           <PanelBody title={__('Setting', 'p4ge')}>
             <ColorPaletteControl
               label={__('Table Color', 'p4ge')}
-              value={ color }
-              onChange={ this.props.onTableColorChange }
+              value={ attributes.color }
```
Inwerpsel, line 49: Still need to make this read from the right variable, probably that will be easier to do if we make `CSS_VARIABLES_ATTRIBUTE` an object instead of an array. Then `value` would just be `attributes.css_variables['spreadsheet-row-background'`.

assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -16,22 +16,38 @@ export class Spreadsheet extends Component {
...
+      setAttributes( { [ attributeName ]: value } );
+    };
 
     const colors = [
       { name: 'blue', color: '#C9E7FA' },
       { name: 'green', color: '#D0FAC9' },
       { name: 'grey', color: '#DCDCDC' },
     ];
 
+    const toCssVariable = ( variableName ) => ( value ) => {
```
Inwerpsel, line 31: Probably this function can be extracted so it can be reused in other blocks. But maybe it makes more sense to do once we implement the same in another block?

assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -16,22 +16,38 @@ export class Spreadsheet extends Component {
   renderEdit() {
     const { __ } = wp.i18n;
 
-    const { color } = this.props;
+    const { attributes, setAttributes } = this.props;
+
+    const toAttribute = attributeName => value => {
```
pablocubico, line 21: (Non-blocking comment): This one is already defined in Covers and Articles, maybe we can move it somewhere like in some kind of "block helpers" folder/file?

Inwerpsel: We can, though the tricky bit is that it currently depends on `setAttributes` from the scope where it's defined. So if we isolate it then it would need to get that as an argument, like `setAttributes => attributeName => value => { setAttributes( { [ attributeName ]: value } ) }`. But I didn't do that yet because maybe we can look into a way to define this as a HOC or something else together with reading the attribute value, so we can avoid the following (having to provide the attribute name everywhere in 2 props, once as a string, once as a property):
```
            value={ attributes.url }
            onChange={ toAttribute( 'url' ) }
```

PR #268, opened 2020-05-07
PLANET-4143 All blocks: Add https:// validation error for url fields

dpivo:
- Added new control to handle the URL inputs.
- Added styling for the error mesasage

### DIFF
--- a/assets/src/blocks/Articles/Articles.js
+++ b/assets/src/blocks/Articles/Articles.js
@@ -10,6 +10,7 @@ import withCharacterCounter from '../../components/withCharacterCounter/withChar
 import TagSelector from '../../components/TagSelector/TagSelector';
 import PostSelector from '../../components/PostSelector/PostSelector';
 import PostTypeSelector from '../../components/PostTypeSelector/PostTypeSelector';
+import {URLInput} from "../../components/URLInput/URLInput";
 
 
 const TextControl = withCharacterCounter( BaseTextControl );
@@ -61,11 +62,10 @@ export class Articles extends Component {
 
 
         <div>
-          <TextControl
+          <URLInput
             label={__('Button Link', 'p4ge')}
-            placeholder={__('Add read more button link', 'p4ge')}
-            value={attributes.read_more_link}
-            onChange={ toAttribute('read_more_link')}
+            value={ attributes.read_more_link }
+            onChange={ toAttribute('read_more_link') }
           />
         </div>
 
--- a/assets/src/blocks/Counter/Counter.js
+++ b/assets/src/blocks/Counter/Counter.js
@@ -6,6 +6,7 @@ import {
 import { LayoutSelector } from '../../components/LayoutSelector/LayoutSelector';
 import { Preview } from '../../components/Preview';
 import withCharacterCounter from '../../components/withCharacterCounter/withCharacterCounter';
+import {URLInput} from "../../components/URLInput/URLInput";
 
 const TextControl = withCharacterCounter( BaseTextControl );
 const TextareaControl = withCharacterCounter( BaseTextareaControl );
@@ -86,9 +87,9 @@ export class Counter extends Component {
         </div>
 
         <div>
-          <TextControl
-            label= { __('Completed API URL', 'p4ge') }
-            placeholder= { __('API URL of completed number. If filled in will overide the \'Completed\' field', 'p4ge') }
+          <URLInput
+            label={ __('Completed API URL', 'p4ge') }
+            placeholder={ __('API URL of completed number. If filled in will overide the \'Completed\' field', 'p4ge') }
             value={ this.props.completed_api }
             onChange={ this.props.onCompletedAPIChange }
           />
--- a/assets/src/components/URLInput/URLInput.js
+++ b/assets/src/components/URLInput/URLInput.js
@@ -1,0 +1,31 @@
+import {Component} from '@wordpress/element';
+import {TextControl} from '@wordpress/components';
+import {URLValidationMessage} from "../../components/URLValidationMessage/URLValidationMessage";
+
+export class URLInput extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+
+    const {__} = wp.i18n;
+    const { label, placeholder, value, onChange, disabled, help } = this.props
+
+    return (
+      <div>
+          <TextControl
+            label={label}
+            placeholder={placeholder}
+            value={value}
+            onChange={onChange}
+            disabled={disabled}
+            help={help}
+          />
+          <URLValidationMessage
+            url={value}
+          />
+      </div>
+    )
+  }
+}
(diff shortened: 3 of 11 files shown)


assets/src/components/CustomLink/CustomLink.js
```diff
@@ -0,0 +1,48 @@
...
+const TextControl = withCharacterCounter( BaseTextControl );
+
+export class CustomLink extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+
+  isCustomLinkValid() {
+    return !!this.props.value ? this.props.value.substr(0, 8).toLowerCase() === 'https://' : true;
```
Inwerpsel, line 15: Maybe it can be a bit nicer if we use javascript's API for URLs? See this SO answer: https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url/49849482#answer-43467144

Maybe it's also more readable if you make a separate return statement before this one, if `this.props.value` is false. There's already a lot of things going on on this line :)

assets/src/components/CustomLink/CustomLink.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  render() {
+
+    const {__} = wp.i18n;
+    const allowSearch = !!this.props.allowSearch;
```
Inwerpsel, line 21: Why not `const { allowSearch } = this.props;`? You could then also add the other props to that.

assets/src/components/CustomLink/CustomLink.js
```diff
@@ -0,0 +1,48 @@
+import {Component} from '@wordpress/element';
+import {URLInput} from '@wordpress/block-editor'
+import {TextControl as BaseTextControl} from '@wordpress/components';
+import withCharacterCounter from '../../components/withCharacterCounter/withCharacterCounter';
+
+const TextControl = withCharacterCounter( BaseTextControl );
+
+export class CustomLink extends Component {
```
Inwerpsel, line 8: It could also just have the same name as the WP component it's wrapping (`URLInput`). Sure it's not the exact same component, but it's still an input for a url, so would be nice if it still had that descriptive name. It would also mean you don't need to change the component name everywhere it's used, you only need to change the import to point to the new one and add the new attributes.

You can alias the import of WP's `URLInput` to resolve the name conflict.

assets/src/components/CustomLink/CustomLink.js
```diff
@@ -0,0 +1,48 @@
+import {Component} from '@wordpress/element';
+import {URLInput} from '@wordpress/block-editor'
+import {TextControl as BaseTextControl} from '@wordpress/components';
+import withCharacterCounter from '../../components/withCharacterCounter/withCharacterCounter';
+
+const TextControl = withCharacterCounter( BaseTextControl );
```
Inwerpsel, line 6: Not sure if any url field would require a character counter? The `TextControl` elements that were used for urls in some places technically were enhanced with this function because they didn't have a dedicated element, however it didn't do anything anyway as the props required for the counter were not there.

But can't we just make every instance use `UrlInput`? That would make this new component a lot simpler.

assets/src/components/CustomLink/CustomLink.js
```diff
@@ -0,0 +1,48 @@
...
+            label={this.props.label}
+            placeholder={this.props.placeholder}
+            value={this.props.value}
+            onChange={this.props.onChange}
+            disabled={this.props.disabled}
+            help={this.props.help}
+          />
+        }
+        {!this.isCustomLinkValid() &&
+        <span className='input_error'>{ __('The URL must start with "HTTPS://"', 'p4ge') }</span>}
```
Inwerpsel, line 44: Maybe not within the scope of this ticket, but maybe we can add some kind of validation so that a post cannot be published when it has a block with an invalid url?

assets/src/components/CustomLink/CustomLink.js
dpivo: I split up the falsy check with the validation check.

assets/src/components/URLInput/URLInput.js
```diff
@@ -0,0 +1,50 @@
...
+
+    const isCustomLinkValid = () => {
+      if ( !this.props.value ) {
+        return true;
+      }
+
+      try {
+        const url = new URL(this.props.value);
+        if ( url.protocol !== 'https:') {
+          errorMessage = 'The URL must start with "HTTPS://"';
```
Inwerpsel, line 25: Probably would be more recognizable to the user to have `https` in lowercase?

assets/src/blocks/Articles/Articles.js
```diff
@@ -61,11 +62,11 @@ export class Articles extends Component {
 
 
         <div>
-          <TextControl
+          <URLInput
             label={__('Button Link', 'p4ge')}
-            placeholder={__('Add read more button link', 'p4ge')}
-            value={attributes.read_more_link}
-            onChange={ toAttribute('read_more_link')}
+            placeholder={ __('API URL of completed number. If filled in will overide the \'Completed\' field', 'p4ge') }
```
Inwerpsel, line 67: Looks like this text got here by mistake.

assets/src/blocks/Counter/Counter.js
```diff
@@ -86,11 +87,12 @@ export class Counter extends Component {
...
         <div>
-          <TextControl
-            label= { __('Completed API URL', 'p4ge') }
-            placeholder= { __('API URL of completed number. If filled in will overide the \'Completed\' field', 'p4ge') }
+          <URLInput
+            label={ __('Completed API URL', 'p4ge') }
+            placeholder={ __('API URL of completed number. If filled in will overide the \'Completed\' field', 'p4ge') }
             value={ this.props.completed_api }
             onChange={ this.props.onCompletedAPIChange }
+            disabled={ false }
```
Inwerpsel, line 95: Other usages don't have this, is it needed here?

assets/src/components/URLInput/URLInput.js
```diff
@@ -0,0 +1,50 @@
...
+
+  render() {
+
+    const {__} = wp.i18n;
+    const { label, placeholder, value, onChange, disabled, help } = this.props
+
+    let errorMessage = "";
+
+
+    const isCustomLinkValid = () => {
```
Inwerpsel, line 17: Probably best to use "URL" here for the naming as well.

From the naming you would not expect this function to have side effects and just return whether the url is valid. However it also changes `errorMessage`, which makes the logic a bit harder to follow.

Maybe it makes sense to have a `UrlValidationMessage` component instead? It would be a very simple component that just has the validation logic and only renders the `<span>` if it gets an invalid url. Then this `URLInput` component would also become simpler.

assets/src/components/URLInput/URLInput.js
dpivo: possibly. However the spec of the ticket asked for it to be upper case. It is possibly to highlight it.

assets/src/components/URLValidationMessage/URLValidationMessage.js
```diff
@@ -0,0 +1,42 @@
...
+
+    const isURLValid = () => {
+      if ( !url ) {
+        return true;
+      }
+
+      try {
+        const urlObject = new URL(url);
+        if ( urlObject.protocol !== 'https:') {
+          errorMessage = 'The URL must start with "HTTPS://"';
```
Inwerpsel, line 24: Can you make `https` lowercase?

PR #271, opened 2020-05-11
PLANET-4774 Add Duplicate postmeta report & WP CLI command to clean duplicate postmeta

sagarsdeshmukh:
[JIRA 4774](https://jira.greenpeace.org/browse/PLANET-4774)

- Add duplicate postmeta report in WP admin under "Blocks" menu
- The admin menu is visible only to 'administrator' user
- Add "Duplicate Postmeta" button on top of report to delete all duplicate postmeta records
- Add WP CLI command  to clean duplicate postmeta (`wp p4-gblocks remove_duplicate_postmeta`)

eg. https://k8s.p4.greenpeace.org/sagardev/wp-admin/admin.php?page=postmeta_report

### DIFF
--- a/classes/command/class-controller.php
+++ b/classes/command/class-controller.php
@@ -56,7 +56,36 @@ class Controller {
 		WP_CLI::log( 'Conversion duration: ' . round( $seconds_elapsed ) . ' seconds' );
 	}
 
-	// Add here new sub-commands e.g. wp p4-blocks new_sub_command.
+	/**
+	 * Sub command that removes duplicate postmeta records
+	 *
+	 * @throws WP_CLI\ExitException The thrown exception.
+	 */
+	public function remove_duplicate_postmeta() {
+
+		$start = microtime( true );
+
+		try {
+			WP_CLI::log( 'Removing duplicate postmeta records...' );
+
+			$deleted_rows = Duplicated_Postmeta::remove();
+
+			if ( $deleted_rows ) {
+				WP_CLI::success( "Removed $deleted_rows duplicate postmeta record/s" );
+			} else {
+				WP_CLI::log( 'No duplicate postmeta record found.' );
+			}
+		} catch ( \Error $e ) {
+			WP_CLI::error( $e->getMessage() );
+		} catch ( \Exception $e ) {
+			WP_CLI::log( 'Exception: ' . $e->getMessage() );
+		}
+
+		$seconds_elapsed = microtime( true ) - $start;
+		WP_CLI::log( 'Execution duration: ' . round( $seconds_elapsed ) . ' seconds' );
+	}
+
+	// Add here new sub-commands e.g. wp p4-gblocks new_sub_command.
 	// public function new_sub_command() {}.
 }
 
--- a/classes/command/class-duplicated-postmeta.php
+++ b/classes/command/class-duplicated-postmeta.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Detect and remove duplicate postmeta command
+ *
+ * @package P4GBKS
+ */
+
+namespace P4GBKS\Command;
+
+/**
+ * Class for detect and remove duplicate postmeta records.
+ */
+class Duplicated_Postmeta {
+
+	/**
+	 * List of meta_key check in duplicate list.
+	 *
+	 * @const array META_KEY_LIST.
+	 */
+	public const META_KEY_LIST = [
+		'sm_cloud',
+		'weight',
+		'_credit_text',
+		'_wp_attachment_image_alt',
+		'_searchwp_last_index',
+		'_media_restriction',
+	];
+
+	/**
+	 * Remove duplicate postmeta records
+	 *
+	 * @return int
+	 */
+	public static function remove() {
+
+		global $wpdb;
+
+		// phpcs:disable
+		$sql = "DELETE t1 FROM wp_postmeta t1 
+				INNER JOIN wp_postmeta t2  
+				WHERE  t1.meta_id < t2.meta_id 
+				AND  t1.meta_key = t2.meta_key 
+				AND t1.post_id = t2.post_id 
+				AND t1.meta_key IN (" . self::generate_placeholders( self::META_KEY_LIST, 1 ) . ")";
+
+		$prepared_sql = $wpdb->prepare( $sql, self::META_KEY_LIST );
+
+		return $wpdb->query( $prepared_sql );
+		// phpcs:enable
+	}
+
+	/**
+	 * Detect duplicate postmeta records
+	 *
+	 * @return object
+	 */
+	public static function detect() {
+
+		global $wpdb;
+
+		// phpcs:disable
+		$sql = "SELECT `post_id`, COUNT(meta_id) AS all_count , COUNT(DISTINCT meta_key) AS unique_count
+				FROM `wp_postmeta`
+				WHERE meta_key IN (" . self::generate_placeholders( self::META_KEY_LIST, 1 ) . ")
+				GROUP by `post_id`
+				HAVING all_count <> unique_count 
+				ORDER BY `all_count` DESC ";
+
+		$prepared_sql = $wpdb->prepare( $sql, self::META_KEY_LIST );
+
+		return $wpdb->get_results( $prepared_sql );
+		// phpcs:enable
+	}
+
+
+	/**
+	 * Generate a bunch of placeholders for use in an IN query.
+	 *
+	 * @param array $items The items to generate placeholders for.
+	 * @param int   $start_index The start index to use for creating the placeholders.
+	 * @return string The generated placeholders string.
+	 */
+	private static function generate_placeholders( array $items, int $start_index ): string {
+		$placeholders = [];
+		foreach ( range( $start_index, count( $items ) + $start_index - 1 ) as $i ) {
+			$placeholders[] = "'%$i\$s'";
+		}
+		return implode( ',', $placeholders );
+	}
+}
--- a/classes/controller/menu/class-postmeta-check-controller.php
+++ b/classes/controller/menu/class-postmeta-check-controller.php
@@ -1,0 +1,70 @@
+<?php
+/**
+ * Post meta duplicate records check class
+ *
+ * @package P4BKS\Controllers\Menu
+ */
+
+namespace P4GBKS\Controllers\Menu;
+
+use P4GBKS\Command\Duplicated_Postmeta;
+
+/**
+ * Class Postmeta_Check_Controller
+ */
+class Postmeta_Check_Controller extends Controller {
+	/**
+	 * Create menu/submenu entry.
+	 */
+	public function create_admin_menu() {
+		$current_user = wp_get_current_user();
+		if ( in_array( 'administrator', $current_user->roles, true ) ) {
+			add_submenu_page(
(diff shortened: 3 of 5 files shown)


classes/command/class-remove-duplicate-postmeta.php
```diff
@@ -0,0 +1,33 @@
...
+	 * Remove duplicate postmeta records
+	 *
+	 * @return boolean
+	 */
+	public function remove_duplicate() {
+
+		global $wpdb;
+
+		// phpcs:disable
+		$sql = "DELETE t1 FROM wp_postmeta t1
```
Inwerpsel, line 24: There are some postmeta keys that actually need to have multiple values. I don't know if there are many of these, but I guess it would be prudent to work with a whitelist here of keys that we know can be removed?

classes/command/class-controller.php
```diff
@@ -56,7 +56,37 @@ public function convert_to_gutenberg( $args ) {
...
+	 * @throws WP_CLI\ExitException The thrown exception.
+	 */
+	public function remove_duplicate_postmeta() {
+
+		$start = microtime( true );
+
+		try {
+			WP_CLI::log( 'Removing duplicate postmeta records...' );
+
+			$instance   = new Remove_Duplicate_Postmeta();
```
Inwerpsel, line 71: You can make `remove_duplicate` static as it doesn't need an instance.

classes/command/class-controller.php
```diff
@@ -56,7 +56,37 @@ public function convert_to_gutenberg( $args ) {
...
+			$is_deleted = $instance->remove_duplicate();
+
+			if ( $is_deleted ) {
+				WP_CLI::success( 'Removed all duplicate postmeta records' );
```
Inwerpsel, line 75: `$is_deleted` will actually contain the number of deleted rows, so you could log this number here too. Maybe you can rename it to `$deleted_rows`.

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,129 @@
...
+ * Post meta duplicate records check class
+ *
+ * @package P4BKS\Controllers\Menu
+ */
+
+namespace P4GBKS\Controllers\Menu;
+
+use P4GBKS\Command\Remove_Duplicate_Postmeta;
+
+if ( ! class_exists( 'Postmeta_Check_Controller' ) ) {
```
Inwerpsel, line 12: You can remove this `class_exists` check, it doesn't load this file more than once.

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,129 @@
...
+		 * Create menu/submenu entry.
+		 */
+		public function create_admin_menu() {
+			if ( ! current_user_can( 'manage_options' ) ) {
+				return;
+			}
+
+			$current_user = wp_get_current_user();
+
+			if ( in_array( 'administrator', $current_user->roles, true ) && current_user_can( 'manage_options' ) ) {
```
Inwerpsel, line 28: `current_user_can( 'manage_options')` will always be true at this point, as the function returns earlier if it's not.

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,129 @@
...
+					} else {
+						$data['message'] = __( 'No duplicate postmeta record found.', 'planet4-blocks-backend' );
+					}
+				} catch ( \Error $e ) {
+					$data['message'] = $e->getMessage();
+				} catch ( \Exception $e ) {
+					$data['message'] = __( 'Exception: ', 'planet4-blocks-backend' ) . $e->getMessage();
+				}
+			}
+			return true;
```
Inwerpsel, line 69: The return value here is not used and it's also always true. Probably makes more sense to not return anything here?

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,129 @@
...
+			// phpcs:disable
+			$sql = "SELECT `post_id`, COUNT(meta_id) AS all_count , COUNT(DISTINCT meta_key) AS unique_count
+					FROM `wp_postmeta`
+					GROUP by `post_id`
+					HAVING all_count > $limit AND all_count <> unique_count
+					ORDER BY `all_count` DESC ";
+
+			$results = $wpdb->get_results( $sql );
+
+			echo '<div id="settings_div" class="wrap">';
```
Inwerpsel, line 97: Maybe you can turn this into a twig file? Though I'm not sure if Twig would need any additional setup to make it work for these admin pages. If that setup would be too complex then it's ok to leave it in PHP.

classes/command/class-remove-duplicate-postmeta.php
Inwerpsel: Maybe you can then change the semantics a bit, so the classname is `Duplicated_Postmeta`. Then you could name the function `Duplicated_Postmeta::remove()`. You could then add the logic that is now in the admin page controller for detection (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/271/files#diff-ccb958aec6798a7ea30f1315a3632727R89) in a function called `Duplicated_Postmeta::detect()`.

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,129 @@
...
+			$limit = 4;
+			if ( isset( $_GET['limit'] ) ) { //phpcs:ignore WordPress.Security.NonceVerification.Recommended
+				$limit = (int) $_GET['limit']; //phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			}
+
+			$data = [];
+			$this->handle_submit( $data );
+
+			// phpcs:disable
+			$sql = "SELECT `post_id`, COUNT(meta_id) AS all_count , COUNT(DISTINCT meta_key) AS unique_count
```
Inwerpsel, line 89: This will list all posts that have at least 4 non-unique postmeta keys, however running the delete will actually remove all the duplicate records. Probably the detection should report everything that will be removed by submitting? Maybe you can write the `WHERE` once and make both detect and remove use it?

classes/command/class-remove-duplicate-postmeta.php
sagarsdeshmukh: > There are some postmeta keys that actually need to have multiple values. I don't know if there are many of these, but I guess it would be prudent to work with a whitelist here of keys that we know can be removed?

I was unaware need of postmeta key with multiple values, could you please provide some examples?

sagarsdeshmukh: > Maybe you can then change the semantics a bit, so the classname is `Duplicated_Postmeta`. Then you could name the function `Duplicated_Postmeta::remove()`. You could then add the logic that is now in the admin page controller for detection (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/271/files#diff-ccb958aec6798a7ea30f1315a3632727R89) in a function called `Duplicated_Postmeta::detect()`.

- yes It will be more suitable. I will update it.

classes/controller/menu/class-postmeta-check-controller.php
sagarsdeshmukh: yes Its better to move it in twig file.

classes/controller/menu/class-postmeta-check-controller.php
sagarsdeshmukh: I removed the `$limit` var, it's not needed any more, so now the both display counts and actual deleting postmeta's will be equal.

classes/command/class-remove-duplicate-postmeta.php
Inwerpsel: > > There are some postmeta keys that actually need to have multiple values. I don't know if there are many of these, but I guess it would be prudent to work with a whitelist here of keys that we know can be removed?
> 
> I was unaware need of postmeta key with multiple values, could you please provide some examples?

It looks like wpml depends on this, they call `get_post_meta` with the `$single` parameter` to false and then loop over it in here:
persistence/app/public/wp-content/plugins/sitepress-multilingual-cms/inc/wpml-api.php:625

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Class Postmeta_Check_Controller
+ */
+class Postmeta_Check_Controller extends Controller {
+	/**
+	 * Create menu/submenu entry.
+	 */
+	public function create_admin_menu() {
+		$current_user = wp_get_current_user();
+		if ( in_array( 'administrator', $current_user->roles, true ) && current_user_can( 'manage_options' ) ) {
```
Inwerpsel, line 21: Won't an administrator by definition always have all capabilities?

sagarsdeshmukh: Yes he has. The second check is't needed here.

classes/command/class-duplicated-postmeta.php
```diff
@@ -0,0 +1,53 @@
...
+	 *
+	 * @return int
+	 */
+	public static function remove() {
+
+		global $wpdb;
+
+		// phpcs:disable
+		$sql = "DELETE t1 FROM wp_postmeta t1 
+				INNER JOIN wp_postmeta t2
```
Inwerpsel, line 25: Can we extract the where condition here and in `detect()` to a function so that it uses the same? That way we're guaranteed that `remove()` will only ever remove records that were returned in `detect()` even if we would make changes to the where condition.

Inwerpsel: If we use the condition that is used currently for the delete in the detect as well, then we can easily make it show not only the posts, but also which meta keys are duplicated and a count per post_id + meta_key combination. That could help us in quickly seeing which part of the code is causing the duplication.

sagarsdeshmukh:
Hi @Inwerpsel - I updated the PR. Please have a look. 👍

classes/command/class-remove-duplicate-postmeta.php
sagarsdeshmukh: > There are some postmeta keys that actually need to have multiple values. I don't know if there are many of these, but I guess it would be prudent to work with a whitelist here of keys that we know can be removed?

I added a meta_key whitelist at [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/271/files#diff-6909d1fef617f885c6b17e50fc90b6f4R20)

classes/controller/menu/class-postmeta-check-controller.php
```diff
@@ -0,0 +1,70 @@
...
+				$deleted_rows = Duplicated_Postmeta::remove();
+			} catch ( \Error $e ) {
+				$data['message'] = $e->getMessage();
+			} catch ( \Exception $e ) {
+				$data['message'] = __( 'Exception: ', 'planet4-blocks-backend' ) . $e->getMessage();
+			}
+
+			if ( $deleted_rows ) {
+				// translators: %d = The duplicate postmeta count.
+				$data['message'] = sprintf( __( 'Remove %d duplicate postmeta record/s successfully.', 'planet4-blocks-backend' ), $deleted_rows );
```
Inwerpsel, line 51: Should probaby be "records"

PR #273, opened 2020-05-11
PLANET-5017: Re-implement Spreadsheet block frontend in React

pablocubico:
Porting the Spreadsheet block's frontend to React.

Ref: https://jira.greenpeace.org/browse/PLANET-5017


pablocubico:
A few notes about this:

0. I need to squash some commits!

1. I tried to avoid the `render_callback` part on the PHP side by just rendering the node in which we inject the React component in the `save()` method of the editor, but using the `save()` method in the Block definition triggers a validation error in the Editor, as our current blocks return `null`, any value different than `null` will trigger a validation error and ask you to "Resolve" the block (yuck!). We could just return an HTML node there to render the React content in the frontend, but changing the `save()` function will make the current spreadsheets unusable. I just made it a "template-less" render, it just renders a node. Ideally I would love to skip the PHP side in these cases, but let's stick to `null` and `render_callback` for now. ( This is a long standing controversy in Gutenberg, there are a few discussions about it out there:
https://github.com/WordPress/gutenberg/issues/10444 )

2. I added an endpoint to transform the CSV to JSON and keep using WP's cache store the Spreadsheet as JSON data. I was thinking it would be fast since the cache is persistent, so the data should be obtained from cache even from different visitors, but then @Inwerpsel came in and said: "Hey, why don't you just get the spreadsheet from the client side?", which is pretty much the same from the component's point of view, is just a bunch of JSON rows. So, as Pieter said, that could avoid WP's overhead, but I guess it could add some overhead to the Javascript side since we need to turn that CSV into something iterable, which is super simple but if the Spreadsheet is long, it may have a performance impact on low-end cellphones. Which is also a concern for React, the React embed is not that big, but it adds quite a bunch of logic for the browser to process, despite the KB size. There is also some cost in the DNS Lookup and connecting to Google if we load it directly from the client side. 

So, I'm not sure about (2), maybe we can test these assumptions? Is the WP overhead that bad? Is parsing the CSV lines to JSON super fast? What do you think? 

Ping @Inwerpsel @sagarsdeshmukh @comzeradd

comzeradd:
Regarding point 2: I'm leaning mostly towards 1st option. Since we already need to render the block in the backend, it seems less complicated and faster for the end-user to use WP cache store. Unless I'm missing something, it also feels more work to do the 2nd option. Which is not necessarily bad if that's the proper way of implementation. But maybe this block is a good opportunity to see how this works and do something similar with WYSIWYG blocks. Rendering on the client-side could be a very good option if we didn't render the block in the backend at all (eg. the native Embed block for Facebook).

Inwerpsel:
Regarding 1. Seems like the most sensible thing to do indeed. Maybe we can set this up in a generic way for all blocks that are rendered frontend? We could use
```
<div data-block="{{ block.id }}" data-attributes="{{ block.attributes.toJson }}" class="{{ block.placeholder_class }}"></div>
```
for all blocks.

Actually, maybe then we can also circumvent the `save()` issue :thinking: Because we would have only one "block shape change" to deal with, from null to this format, which we can write a resolver for, probably. We would only need to write one as the element and attributes don't need to change to support other blocks. Even if the html would be changed externally, we can write a resolver for that as the content can be entirely reconstructed from the block attributes in the html comment.

Maybe worth a shot? Then we wouldn't need the render on the backend, which is basically just turning one representation of the block's data into another one (block comment -> data attributes).

As for 2. I think the CSV parsing should be really fast and only an issue for huge spreadsheets, like 100k or more rows, far beyond what would make sense to include on a page. But we can try if we want to be sure. Possibly we can get this data as JSON too from the Google sheets API.

The reason I remarked it was because I assume that Google sheets has a pretty extensive CDN to deliver this data efficiently anywhere in the world, so very likely the round trip to our server will be significantly longer, even without WP overhead. Note that currently even though a placeholder for the blocks is rendered on the backend, the data of the sheet doesn't come with that. The frontend needs to do a separate call to a custom endpoint before it can start rendering. So for the end user the difference in speed is between how fast we can serve a static resource vs how fast google can do that.

Though serving it from our backend also has some advantages, as it ensures the sheet will work regardless of possible issues that might prevent the browser from getting the data from Google. There might be some other considerations when including data from external domains.

I'm ok with the current approach of caching it on our server as there's probably no noticeable difference to the end user in most cases. Probably the additional load on our servers is also negligible.

pablocubico:
Hey @Inwerpsel , sorry for the delay.

Regarding 1:
I tried doing that in the `save()` method but my version was using `data-url`, a "per attribute" property, which does not scale well if we change the attributes..., it actually makes sense if to only pass "data-attributes" to it and extract the ones we want in the frontend renderer. I'll give it a try and use the "deprecations" feature of Gutenberg to bypass the validation issue.

Regarding 2:
Alright, let's keep the current version which uses the cache and could help in avoiding some DNS latency or Google issues... in case Google has any.  We can always change it to a client fetch since it's pretty much the same from the frontend side.

pablocubico:
@comzeradd @Inwerpsel 

Branch updated to latest master. I removed the `render_callback` from the PHP side and created a generic method to render blocks in the frontend. The `save()` provides a `className` which includes the block's name but also any additional classes from the [block styles](https://www.billerickson.net/block-styles-in-gutenberg/) feature, so it's quite handy.

I wrote a deprecation handler to avoid any warnings from the `save()` method change. Looks pretty good. I also ported the CSS variables from the spreadsheet colors feature from Twig to React.

assets/src/frontendIndex.js
```diff
@@ -0,0 +1,20 @@
+import { SpreadsheetFrontend } from './blocks/Spreadsheet/SpreadsheetFrontend';
+
+const PLANET4_BLOCK_PREFIX = 'wp-block-planet4-blocks-';
+const COMPONENTS = {
+	spreadsheet: SpreadsheetFrontend
+};
+
+const getFrontendBlockComponent = blockNode => COMPONENTS[
+	Array.from(blockNode.classList)
+		.find( className => className.includes( PLANET4_BLOCK_PREFIX ))
```
Inwerpsel, line 10: Maybe we can create a dedicated `data-` attribute for this instead of a class? Something like `data-render="spreadsheet"` Then we could get the block name from the attribute's value, instead of having to search the classnames and remove the prefix. [The querySelector below](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/273/files?file-filters%5B%5D=.js&file-filters%5B%5D=.php&file-filters%5B%5D=.scss&hide-deleted-files=true#diff-e318476ddd94efc2f71618da8e052ddbR14) would then also be a bit simpler, just matching any elements that have the data attribute, regardless of its value.
```
document.querySelectorAll(`[data-render]`).forEach(...)
```

classes/blocks/class-spreadsheet.php
```diff
@@ -49,9 +48,6 @@ public function __construct() {
 	 * @return array The data to be passed in the View.
 	 */
 	public function prepare_data( $fields ): array {
```
Inwerpsel, line 50: If `prepare_data` is used only in the backend render, then you can remove this function, or make it just return an empty array.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,186 @@
...
+export class SpreadsheetFrontend extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      loading: true,
+      sheet_data: null,
+      invalidSheetId: false,
+      searchText: '',
+      sortDirection: 'asc',
+      sortColumnIndex: 0,
```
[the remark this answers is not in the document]
pablocubico: I'll make it `null` then and leave it unsorted at start. 👍

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,186 @@
...
+    };
+
+    this.onHeaderClick = this.onHeaderClick.bind(this);
+  }
+
+  onHeaderClick( index ) {
+    if ( index == this.state.sortColumnIndex ) {
```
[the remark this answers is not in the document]
pablocubico: Hm I think so, I'll remove it and test.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,186 @@
...
+    })
+    return filteredRows;
+  }
+
+  boldText( text ) {
+    return <b>{ text }</b>;
+  }
+
+  highlightMatches( cellValue, searchText ) {
+    let reg = new RegExp('(' + searchText + ')', 'gi')
```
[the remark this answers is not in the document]
pablocubico: Actually that was my first suggestion when we estimated PLANET-4914 on a backlog refinement call, but you suggested doing it ourselves to avoid including a library as it would be pretty simple... which actually made sense, I just did a quick test adding the library and the `frontendIndex` build went from 7.2K to 24K... I think we can keep this version unless we need extra features as the ones provided by that library or we start using highlighting in other places...

assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -89,20 +89,18 @@ export class Spreadsheet extends Component {
     );
   }
 
+  renderView() {
```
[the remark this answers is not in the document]
pablocubico: Hmmm not sure I follow, this is not rendering the placeholder as we do in the frontend... maybe you saw it's not rendering, the reason the table is not being rendered is because of a bug I introduced in the endpoint here: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/273/files#diff-765c584bd08bac4ddefd3a6bfc6e4706R126

I am validating the `sheet_id` as if it was an integer, I don't know why I thought so... just out of habit I guess...

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,186 @@
...
+          { parts[i] }
+        </b>
+      )
+    }
+
+    return <Fragment>{ parts }</Fragment>
+  }
+
+  renderRows() {
+    const rows = this.filterMatchingRows( this.sortRows( this.state.sheet_data.rows, this.state.sortColumnIndex ) );
```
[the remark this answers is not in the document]
pablocubico: Good catch!! 💯

assets/src/frontendIndex.js
pablocubico: That was the initial approach, but then I couldn't find a way to get the block name in the `save()` method, it receives the `attributes` and a set of `className`s which includes the block's class and any "block style" classes. But I guess I can just make the block name a constant in the block registration constructor and use it in `save()`, and make it a prop of the `FrontendNode`. I think I'll also include the `classNames` (which I'm not doing now) so we can use block styles in the future (or any other CSS stuff).

assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -89,20 +89,18 @@ export class Spreadsheet extends Component {
...
+  }
+
   render() {
     return (
       <div>
         {
           this.props.isSelected
             ? this.renderEdit()
-            : null
+            : this.renderView()
```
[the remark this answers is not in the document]
pablocubico: I would actually love to move the field to the sidebar as it's part of the WYSIWYG refactors doc, most fields in our blocks should be in the sidebar if we want to provide a WYSIWYG UX. I'll see what I can do in terms of UI there.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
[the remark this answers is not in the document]
Inwerpsel: Right, I did suggest that :) Yeah you're right, that increase in frontend footprint doesn't seem worth it if it's our only use case. Still would be nicer if we extracted the highlighting logic from `SpreadsheetFrontend` so we can keep that one a bit more high level.

assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -68,42 +68,60 @@ export class Spreadsheet extends Component {
...
+            <TextControl
+              label={__('Spreadsheet URL', 'planet4-blocks-backend')}
+              placeholder={__('Enter Google Spreadsheet URL', 'planet4-blocks-backend')}
+              value={ attributes.url }
+              onChange={ toAttribute( 'url' ) }
+            />
+            <div className="sidebar-blocks-help">
+              <ul>
+                <li>
+                  { __(`From Your Google Spreadsheet Table choose File -> Publish on web.
```
mleray, line 80: The description has been updated for https://jira.greenpeace.org/browse/PLANET-5022, so I guess it should be updated here too?

pablocubico: Good catch!

pablocubico:
Hey @Inwerpsel @comzeradd , branch rebased to master, this is currently deployed in the Mars environment.

Addressed all the comments in the review, applied most of the suggestions. I didn't decouple the rendering from the API fetch, I think we can do that later if the requirements change, for example, if we are planning to fetch spreadsheets from other sources than Google docs. 

I also considered registering a Redux Store in the Gutenberg side to handle the data, using the `withSelect/withDispatch/subscribe` combo there instead of `apiFetch`, that way we could use the Redux debugger to monitor the requests, even do rewind, play, and such, but it has no real benefit from the frontend side, as there are no Stores registered there, and seems overkill. For the moment, WP is still focused on the Editor side for their React stuff.

classes/blocks/class-spreadsheet.php
```diff
@@ -42,49 +41,10 @@ public function __construct() {
...
-		$matches = [];
-		preg_match( $google_sheets_pattern, $url, $matches );
-
-		$id = $matches[1] ?? null;
-		if ( ! $id ) {
-			throw new \InvalidArgumentException( 'The url provided is not a valid public google sheets url.' );
-		}
-
-		return $id;
+		return $fields;
```
Inwerpsel, line 47: Probably makes more sense to return an empty array here, as this function normally is't supposed to return `$fields`, even though some of our blocks do that.

assets/src/blocks/CSSVariablesToText.js
```diff
@@ -0,0 +1,4 @@
+export const CSSVariablesToText = CSSVariables =>
+	Object.entries( CSSVariables )
+		.map( cssVariable => `--${cssVariable[0]}: ${cssVariable[1]}` )
+		.join(';'); // React does not like an ending semicolon on cssText
```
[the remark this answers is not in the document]
pablocubico: React actually throws an error if there is a semicolon in the last one: `Warning: Style property values shouldn't contain a semicolon` and suggests removing it. I'll clarify my comment.

assets/src/blocks/CSSVariablesToText.js
```diff
@@ -0,0 +1,4 @@
+export const CSSVariablesToText = CSSVariables =>
```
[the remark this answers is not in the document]
pablocubico: Hmmm I was try to imply this is meant to be inlined and used in the `cssText` property (https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/cssText) , I'll figure out a better name... 🤔

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,228 @@
...
+                        { cell }
+                        <ArrowIcon />
+                      </th>
+                    ))
+                  }
+                </tr>
+              </thead>
+              <tbody>
+                {
+                  this.state.loading === true
```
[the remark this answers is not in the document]
pablocubico: This is a feature: there is an initial state where `loading`  is `null`, when there is no URL set yet, and it shows some bogus content in the Editor so you can actually see how the Spreadsheet would look like even if the URL is not set yet.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  render() {
+    const headers = this.state.spreadSheetData
+      ? this.state.spreadSheetData.header
+      : placeholderData.header;
+
+    return (
+      <Fragment>
+        <div className="block-spreadsheet" style={{ cssText: CSSVariablesToInlineText( this.props.css_variables ) }}>
```
comzeradd, line 169: do you think it would be too complex to have the block inside a `<section>`, instead of a `<div>`. To be more semantically meaningful, but also to preserve consistency with existing blocks markup.

pablocubico: Sure, forgot about that.

assets/src/blocks/CSSVariablesToInlineText.js
```diff
@@ -0,0 +1,14 @@
...
+ * Converts an object with CSS Variables to inline text,
+ * for example, to use in a cssText property.
+ *
+ * @param {object} CSSVariables
+ * @returns {string}
+ */
+export const CSSVariablesToInlineText = CSSVariables => {
+	if ( ! CSSVariables ) return '';
+	return Object.entries( CSSVariables )
+		.map( cssVariable => `--${cssVariable[0]}: ${cssVariable[1]}` )
```
Inwerpsel, line 11: You can destructure the array here for better readability:
```
		.map( [ key, value ] => `--${key}: ${value}` )
```

assets/src/blocks/CSSVariablesToInlineText.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Converts an object with CSS Variables to inline text,
+ * for example, to use in a cssText property.
+ *
+ * @param {object} CSSVariables
+ * @returns {string}
+ */
+export const CSSVariablesToInlineText = CSSVariables => {
```
Inwerpsel, line 8: Sorry to continue on the name nit picking :) But the inline part is not really accurate, you could as write the result of this function to a css file with a `::root {}` selector. Maybe `toDeclarations = CssVariables => {}` makes sense here?

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
```diff
@@ -0,0 +1,148 @@
...
+      </Fragment>
+    );
+  }
+
+  renderView() {
+    const {__} = wp.i18n;
+
+    return <Fragment>
+      {
+        ! this.props.attributes.url
```
Inwerpsel, line 117: `this.props.attributes` is used a few times here, maybe easier on the eye if we destructure it `const { attributes } = this.props`.

assets/src/blocks/HighlightMatches.js
```diff
@@ -0,0 +1,17 @@
+import { Fragment } from '@wordpress/element';
+
+export const HighlightMatches = ( cellValue, searchText, className = 'highlighted-text' ) => {
+	let reg = new RegExp('(' + searchText.trim() + ')', 'gi')
```
Inwerpsel, line 4: Not sure if this is the right place to perform the `trim` on the searchtext. It's not impossible that we actually want to match a trailing whitespace. https://github.com/bvaughn/react-highlight-words also works that way, you can check at https://codesandbox.io/s/5v8yqoxv7k?file=/src/index.js

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -0,0 +1,217 @@
...
+    if (this.state.sortDirection === 'desc') {
+      return sortedRows.reverse();
+    } else {
+      return sortedRows;
+    }
+  };
+
+  filterMatchingRows( rows ) {
+    const filteredRows = rows.filter( row => {
+      return !!row.find( cell => cell.toLowerCase().includes( this.state.searchText.trim().toLowerCase() ) );
```
Inwerpsel, line 119: You can use `row.some` instead here, which already returns a boolean for you.

classes/blocks/class-spreadsheet.php
```diff
@@ -42,49 +41,12 @@ public function __construct() {
 	}
 
 	/**
-	 * Get all the data that will be needed to render the block correctly.
+	 * Required by the `Base_Block` class.
 	 *
-	 * @param array $fields This is the array of fields of this block.
-	 *
-	 * @return array The data to be passed in the View.
+	 * @param array $fields Unused, required by the abstract function.
```
Inwerpsel, line 46: I'm usually not a fan of long inheritance chains, but maybe it's more appropriate here if we create sub-types for each kind of block? So that would make
```
|Base_Block
|\
||Static_Block
|\
||Dynamic|Block
```

Inwerpsel: Not proposing to do this in this PR btw :D probably best handled separately.

Inwerpsel: Then we could just move the `prepare_data` function to `Dynamic_Block`.

assets/src/blocks/HighlightMatches.js
Inwerpsel: And we can just trim `searchText` once when setting it on the state, shouldn't be needed afterwards.

pablocubico: Trimming the value on the state actually prevents you from typing any spaces in the controlled input, is not impossible to match a trailing whitespace but I think it's an edge case, also, as this highlight implementation is just making the text bold, you can't actually see a space being highlighted, so I guess I'll just keep it as is for now, we can adopt that library afterwards if it makes sense... I think we will probably add it at some point...

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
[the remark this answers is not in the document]
pablocubico: So it turns out this is actually needed to set the sortDirection to `asc` in case you clicked in a different header, otherwise it could be `desc` and trigger the animation on click. Anyway, Magali and I will review the header behavior as the next iteration, so this will probably need some changes.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
pablocubico: Neat 💯

classes/blocks/class-spreadsheet.php
pablocubico: Indeed, this is calling for a few refactors.

PR #277, opened 2020-05-14
Add permalink to the blocks usage overview

oekeur:
Adding the permalink to the overview allows for quick linking from the overview to the public link.

Added benefit (reason I worked on it): It's easier (and faster) to make overviews as deliverables

### DIFF
--- a/classes/controller/menu/class-blocks-usage-controller.php
+++ b/classes/controller/menu/class-blocks-usage-controller.php
@@ -137,7 +137,7 @@ if ( ! class_exists( 'Blocks_Usage_Controller' ) ) {
 							<th>' . __( 'Title', 'planet4-blocks-backend' ) . '</th>
 					</tr>';
 					foreach ( $results as $result ) {
-						echo '<tr><td>'.$result->ID .'</td>';
+						echo  '<tr><td><a href="' . get_permalink( $result->ID ) . '" >' . $result->ID . '</a></td>';
 						echo  '<td><a href="post.php?post=' . $result->ID . '&action=edit" >' . $result->post_title . '</a></td></tr>';
 					}
 					echo '</table>';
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh:
We have few more block which also need to update the link.
image

Instead of adding a new column of public link, can we just provide a public link on ID field?

oekeur:
> We have few more block which also need to update the link.
> image
> 
> Instead of adding a new column of public link, can we just provide a public link on ID field?

Sure, seems logical also.
Will work on it next week again. :+1:

oekeur:
Requested changes have been done :)

classes/controller/menu/class-blocks-usage-controller.php
```diff
@@ -250,7 +250,7 @@ public function plugin_blocks_report( $type = 'text' ) {
 					<th>' . __( 'Title', 'planet4-blocks-backend' ) . '</th>
 				</tr>';
 				foreach ( $results as $result ) {
-					echo '<tr><td>'.$result->term_id .'</td>';
+					echo  '<tr><td><a href="' . get_permalink( $result->ID ) . '" >' . $result->ID . '</a></td>';
```
sagarsdeshmukh, line 253: Looks like a typo here, at place of `term_id`.

sagarsdeshmukh: also for tag pages the public link would be diff. you may try `get_term_link()` here.

oekeur: > Looks like a typo here, at place of `term_id`.

Good catch! (not really needed anymore, because tags without redirect page are not possible anymore right? But good for backwards compatibilty maybe)

> also for tag pages the public link would be diff. you may try get_term_link() here.

Yep, got it now! :)

oekeur: Should be fixed :)

oekeur:
Bump :) @sagarsdeshmukh @Inwerpsel

sagarsdeshmukh:
@oekeur - could you squash the commits into single commit before merge?

PR #278, opened 2020-05-19
Planet 5058 - Table background colors

mleray:
See https://jira.greenpeace.org/browse/PLANET-5058

We want to offer the same color options for the Table block as we do for the Spreadsheet block. The only difference is that tables also have a footer, which requires an additional color. This PR takes care of several things:
- Move table background colors to an override file, also add white border to table rows when no stripes ("regular" style)
- Create a custom `edit.js` file for the table block, based on a copy of [the original](https://github.com/WordPress/gutenberg/blob/7dd6c58c3c6e17c85423fff7a666eab29d749689/packages/block-library/src/table/edit.js)
- Add a copy of [the original](https://github.com/WordPress/gutenberg/blob/7dd6c58c3c6e17c85423fff7a666eab29d749689/packages/block-library/src/table/state.js) `state.js` file for table block, needed for import
- Add filter to use our custom table block instead of the core one
- Update Spreadsheet block to use newly created color variables

See corresponding styleguide PR [here](https://github.com/greenpeace/planet4-styleguide/pull/56) and master-theme [here](https://github.com/greenpeace/planet4-master-theme/pull/1093)

### DIFF
--- a/assets/src/BlockFilters.js
+++ b/assets/src/BlockFilters.js
@@ -1,10 +1,12 @@
 const { addFilter } = wp.hooks;
 
 import P4ButtonEdit from './components/p4_button/edit';
+import P4TableEdit from './components/p4_table/edit';
 
 export const addBlockFilters = () => {
   addFileBlockFilter();
   addButtonBlockFilter();
+  addTableBlockFilter();
 };
 
 const addFileBlockFilter = () => {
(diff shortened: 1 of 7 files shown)


pablocubico:
@mleray Looks good to me, it only needs an update.

As for this approach we are taking of copying files (used here & in the Button component overrides), I created a follow up ticket to investigate if we can do that in a cleaner way, by extending or importing parts of the `@wordpress/block-library` instead of hard copying the files. But it's a refactor/investigate issue, not a blocker at all: https://jira.greenpeace.org/browse/PLANET-5121

mleray:
@pablocubico Good idea for the follow-up! Maybe those extra changes you see in the submodule are here because I've had to do a rebase? 🤔

comzeradd:
Could you also rebase/squash the commits? You have an unnecessary merge commit here.

PR #280, opened 2020-05-20
PLANET-5022 Update Spreadsheet block description

mleray:
See https://jira.greenpeace.org/browse/PLANET-5022


assets/src/blocks/Spreadsheet/Spreadsheet.js
```diff
@@ -76,7 +76,8 @@ export class Spreadsheet extends Component {
             placeholder={__('Enter Google Spreadsheet URL', 'planet4-blocks-backend')}
             help={__(`
             From Your Google Spreadsheet Table choose File -> Publish on web.
-            No need to choose the output format, any of them will work.
+            A pop-up window will show up, click on the Publish button and then OK when the confirmation message is displayed.
```
Inwerpsel, line 79: Shouldn't we still include that it will work with any output format? Else they might be confused over which to pick.

PR #283, opened 2020-05-25
PLANET-4143-fix improve invalid URL message

dpivo:
- fix invalid URL message to show https validation error and then invalid URL(based on the domain) Message.

### DIFF
--- a/assets/src/components/URLValidationMessage/URLValidationMessage.js
+++ b/assets/src/components/URLValidationMessage/URLValidationMessage.js
@@ -5,38 +5,29 @@ export class URLValidationMessage extends Component {
     super(props);
   }
 
+  isValid( url ) {
+    if ( !url ) {
+      return true;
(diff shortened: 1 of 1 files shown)


assets/src/components/URLValidationMessage/URLValidationMessage.js
```diff
@@ -20,13 +20,18 @@ export class URLValidationMessage extends Component {
...
+        }
+
+        if ( urlObject.hostname.split('.').length < 2 ) {
+          errorMessage = 'Invalid URL'
           return false;
         }
         return true;
       } catch(e) {
-        errorMessage = 'The URL is invalid';
+        errorMessage = 'The URL must start with "HTTPS://"';
```
Inwerpsel, line 34: It's more recognizable for a user if we put `https` in lowercase.

PR #284, opened 2020-05-26
PLANET-5071 Add data attributes on Covers Block

sagarsdeshmukh:
[JIRA 5071](https://jira.greenpeace.org/browse/PLANET-5071)

- Added a data attributes on Cover and TA boxout blocks
- For consistency, change the indentation of twig file from spaces to tab

### DIFF
--- a/templates/blocks/campaign_covers.twig
+++ b/templates/blocks/campaign_covers.twig
@@ -28,13 +28,17 @@
 				<div class="thumbnail-largeview-container limit-visibility">
 					{% for tag in fields.tags %}
 						<div class="campaign-card-column">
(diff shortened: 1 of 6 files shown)


sagarsdeshmukh:
> It looks good 👍 But for the consistency issue, I actually thought that most of our files (twig or not) currently use spaces instead of tabs?

As per the `[editorconfig](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/.editorconfig#L8)` template twig file should use a tab indentation.

mleray:
@sagarsdeshmukh Ah okay, I didn't know, then we have a lot of them to update!

PR #287, opened 2020-05-27
PLANET-5077: Measure JS code test coverage

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-5077

To test, run `npm install` and `npm test`. 
Use `npm run test:watch` to run the tests in slow motion and see what happens in the browser.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ job-references:
       - run:
           name: "eslint"
           when: always
-          command: eslint public/js/**/*.js
+          command: eslint public/js/**/*.js tests/js
       - store_test_results:
           path: /tmp/test-results
 
@@ -119,6 +119,7 @@ jobs:
             export APP_IMAGE=gcr.io/planet-4-151612/planet4-base-app:${BUILD_TAG}
             export OPENRESTY_IMAGE=gcr.io/planet-4-151612/planet4-base-openresty:${BUILD_TAG}
             make -C planet4-docker-compose ci test
+
       - run:
           name: Test - Extract test artifacts
           when: always
--- a/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
+++ b/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
@@ -2,6 +2,8 @@ import { Component, Fragment } from '@wordpress/element';
 import { InspectorControls } from '@wordpress/block-editor';
 import ColorPaletteControl from '../../components/ColorPaletteControl/ColorPaletteControl';
 
+import { debounce } from 'lodash';
+
 import {
   TextControl,
   PanelBody
@@ -43,7 +45,14 @@ export class SpreadsheetEditor extends Component {
     this.state = {
       invalidSheetId: false,
       errorFetchingSpreadsheet: false,
+      url: '',
     };
+
+    this.debounceSearch = debounce(url => {
+      this.props.setAttributes( { url } );
+    }, 300);
+
+    this.debounceSearch = this.debounceSearch.bind(this)
   }
 
   handleErrors( errors ) {
@@ -54,10 +63,6 @@ export class SpreadsheetEditor extends Component {
     const {__} = wp.i18n;
 
     const { attributes, setAttributes } = this.props;
-
-    const toAttribute = attributeName => value => {
-      setAttributes( { [ attributeName ]: value } );
-    };
 
     const toCssVariables = ( value ) => {
       setAttributes( {
@@ -80,8 +85,13 @@ export class SpreadsheetEditor extends Component {
             <TextControl
               label={__('Spreadsheet URL', 'planet4-blocks-backend')}
               placeholder={__('Enter Google Spreadsheet URL', 'planet4-blocks-backend')}
-              value={ attributes.url }
-              onChange={ toAttribute( 'url' ) }
+              value={ this.state.url }
+              onChange={
+                url => {
+                  this.setState({ url });
+                  this.debounceSearch( url );
+                }
+              }
             />
             <div className="sidebar-blocks-help">
               <ul>
--- a/package.json
+++ b/package.json
@@ -7,18 +7,21 @@
   "license": "GPL-3.0",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
     "start": "wp-scripts start",
     "build": "wp-scripts build --mode production",
     "lint:css": "wp-scripts lint-style",
-    "setupgithooks": "bin/copy-git-hooks.sh"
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -7,18 +7,21 @@
...
+    "test:watch": "env JEST_TIMEOUT=30000 PUPPETEER_HEADLESS=false PUPPETEER_SLOWMO=100 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e"
   },
   "devDependencies": {
     "@wordpress/components": "^8.3.2",
     "@wordpress/data": "^4.9.2",
+    "@wordpress/e2e-test-utils": "^4.5.0",
     "@wordpress/scripts": "3.3.0",
     "autoprefixer": "^9.7.0",
     "bootstrap": "^4.4.1",
+    "core-js": "^3.6.5",
```
pablocubico, line 24: I had to upgrade the `core-js` version to solve a missing modules error.

package.json
```diff
@@ -7,18 +7,21 @@
...
     "lint:css": "wp-scripts lint-style",
-    "setupgithooks": "bin/copy-git-hooks.sh"
+    "setupgithooks": "bin/copy-git-hooks.sh",
+    "test": "env JEST_TIMEOUT=10000 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e",
+    "test:watch": "env JEST_TIMEOUT=30000 PUPPETEER_HEADLESS=false PUPPETEER_SLOWMO=100 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e"
   },
   "devDependencies": {
     "@wordpress/components": "^8.3.2",
     "@wordpress/data": "^4.9.2",
+    "@wordpress/e2e-test-utils": "^4.5.0",
```
pablocubico, line 20: The latest version of the E2E test utils is 4.7.0 (https://www.npmjs.com/package/@wordpress/e2e-test-utils?activeTab=readme), but it had some differences with our setup, for example, it threw:
```
No node found for selector: .block-editor-inserter__search-input
``` 
When trying to use the `insertBlock` helper function, because the selector changed from one version to another. 

I found this message on the E2E docs:
```
It works properly with the minimum version of Gutenberg 5.6.0 or the minimum version of WordPress 5.2.0.
```
But I'm not sure what they mean with `Gutenberg 5.6.0` as Gutenberg is actually a set of packages, there is no package named `gutenberg` in NPM... the latest WP version is 5.4.1. So, I don't know what that `5.6.0` version refers to...

mleray: Maybe they're talking about the version of the Gutenberg plugin? https://wordpress.org/plugins/gutenberg/

.circleci/config.yml
```diff
@@ -111,14 +111,18 @@ jobs:
       - run:
           name: Test - Clone planet4-docker-compose
           command: |
-            git clone https://github.com/greenpeace/planet4-docker-compose
+            git clone -b planet-5077 https://github.com/greenpeace/planet4-docker-compose
```
pablocubico, line 114: Temporarily pinning the docker-compose branch to test this setup.

.circleci/config.yml
```diff
@@ -111,14 +111,18 @@ jobs:
...
+          name: Test - Run Codeception tests
           command: |
             export BUILD_TAG="build-$(cat /tmp/workspace/var/circle-build-num)"
             export APP_IMAGE=gcr.io/planet-4-151612/planet4-base-app:${BUILD_TAG}
             export OPENRESTY_IMAGE=gcr.io/planet-4-151612/planet4-base-openresty:${BUILD_TAG}
             make -C planet4-docker-compose ci test
+      - run:
+          name: Test - Run E2E Jest/Puppeteer tests
+          command: |
+            make -C planet4-docker-compose jstests
```
Inwerpsel, line 125: If there is a lot happening in a make target, it can be handy to include the sub-make targets (like `installnpm installdeps assets runjstests
` here) as separate steps. That will give us the duration of each part, and also make the logs a lot more manageable (Less need to scroll through large amount of console output to find relevant piece.).

Maybe here not each one of `installnpm installdeps assets runjstests` needs a separate step, but definitely for the one that actually runs the test this would be nice.

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
```diff
@@ -2,6 +2,8 @@ import { Component, Fragment } from '@wordpress/element';
 import { InspectorControls } from '@wordpress/block-editor';
 import ColorPaletteControl from '../../components/ColorPaletteControl/ColorPaletteControl';
 
+import { debounce } from 'lodash';
```
Inwerpsel, line 5: Definitely good idea to use debounce here, though maybe this PR is not the best place to change the block's behavior.

tests/js/e2e-tests-helpers.js
```diff
@@ -0,0 +1,70 @@
...
+} from '@wordpress/e2e-test-utils';
+
+/**
+ * Selects a block by name.
+ *
+ * @param {string} name The block name.
+ */
+export const selectBlockByName = async ( name ) => {
+  await selectBlockByClientId(
+    ( await getAllBlocks() ).find( ( block ) => block.name === name ).clientId
```
Inwerpsel, line 13: Don't forget to catch the case where the block is not found.

Inwerpsel: I guess it is expected that this throws an exception, but catching it would still allow to throw a more meaningful error.

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
pablocubico: Indeed, it's just that I realized we needed debounce because of the tests!

comzeradd:
Does it make sense to bring the two extra commits from #316 here? Leaving aside the CI part (the last commit on this othe PR) everything seems ready to be able to run tests locally with `npm test`.

pablocubico:
@comzeradd I just realized those two commits are squashed in this PR, I'll rebase the CI PR to this one.

tests/js/e2e-tests-helpers.js
pablocubico: Added a `throw` in case the block is not found. Tested it and it fails gracefully.  👍

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
pablocubico: This PR is quite old and approved, no need to create a different one for the `debounce`. :shipit:

PR #288, opened 2020-05-27
PLANET-5017 Use a data attribute for rendering

Inwerpsel:


pablocubico:
@Inwerpsel I just realized I was testing without compiling the assets, it actually needs a few changes (e.g.: the `data-render` attribute is never rendered in the frontend), but I'll do these in my branch.

pablocubico:
I actually cherry-picked your commit to add the fix on top of it and avoid creating a merge commit, it will be easier for squashing and rebasing in the original branch. Closing this one, thanks!. 👍

PR #291, opened 2020-05-28
Planet 5060 - Simplify passive state of Take Action cards

mleray:
This PR takes care of the changes for both the Take Action card (https://jira.greenpeace.org/browse/PLANET-5060) and the Take Action boxout (https://jira.greenpeace.org/browse/PLANET-5126):
- Hide CTA in passive state, show it only on rollover
- Make whole card clickable (instead of just the CTA), adjust cursor to pointer
- Underline heading on rollover

### DIFF
--- a/templates/blocks/old_covers.twig
+++ b/templates/blocks/old_covers.twig
@@ -25,42 +25,48 @@
 				<div class="row limit-visibility">
 					{% for cover in covers %}
 						<div class="col-lg-4 col-md-6 cover-card-column">
-							<div
-								class="cover-card card-one"
-								style="background-image: url({{ cover.image }});"
-								data-ga-category="Take Action Covers"
-								data-ga-action="Image"
-								data-ga-label="n/a">
-									<div class="cover-card-content">
-										{% if ( cover.tags ) %}
-											{% for tag in cover.tags %}
-												<a
-													class="cover-card-tag"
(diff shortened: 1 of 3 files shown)


templates/blocks/old_covers.twig
```diff
@@ -25,7 +32,7 @@
 				<div class="row limit-visibility">
 					{% for cover in covers %}
 						<div class="col-lg-4 col-md-6 cover-card-column">
-							<div class="cover-card card-one" style="background-image: url({{ cover.image }});">
+							<div onclick="onCardClick('{{ cover.button_link }}')" class="cover-card card-one" style="background-image: url({{ cover.image }});">
```
comzeradd, line 35: Do you think it's feasible to create the whole card as a `<a>` element and avoid javascript?
I'm thinking accessibility here, to makes sure the whole card is semantically marked as a link, since there will be no other link on it.

mleray: I've tried this approach, but <a> tags can't have children... Then I've tried making an <a> tag positioned absolute and taking the whole div, but then the tags were no longer clickable and the rollover effect on the heading was also lost 🙁

mleray: I've opened [this PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/293) to explore the `<a>` tag option further!

PR #292, opened 2020-05-29
PLANET-4934: Trigger LazyLoad update on Article block

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4934

Trigger a `lazyLoad.update()` on Article block on the admin side, to display images in articles list.  

## Fix

There's no easy React-way to do this using vanilla-lazyload. When the block appears, its content is not yet present and is fetched by a `ServerSideRender` component, so we have to find a way to call lazyLoad only after the server-side rendered content appears, knowing that this does not trigger a `componentDidUpdate` in the parent component.

While I was toying with `setTimeout`, @pablocubico gave me a far better solution that checks the presence of a rendered div and clears itself when it's done.

We could think of different solutions like modifying the output or the server to not use lazyload on this case, but it would be more critical for the frontend, and the fact that we have planned to remove `ServerSideRender` usage at some point pushes us more to a js solution.

### Doc
Vanilla-lazyload : 
https://github.com/verlok/vanilla-lazyload/blob/754a55de96e6fa0dbc5a8262b0425932316c3abc/README.md#usage-with-react 

## Test

On the `/admin` side
- Add a new page with an `Articles` block
  This block should have some articles with images to display.
- Images should appear, in view mode and in edit mode

### DIFF
--- a/assets/src/blocks/Articles/Articles.js
+++ b/assets/src/blocks/Articles/Articles.js
@@ -11,12 +11,36 @@ import TagSelector from '../../components/TagSelector/TagSelector';
 import PostSelector from '../../components/PostSelector/PostSelector';
 import PostTypeSelector from '../../components/PostTypeSelector/PostTypeSelector';
 import {URLInput} from "../../components/URLInput/URLInput";
-
+import { v4 as uuidv4 } from 'uuid';
 
 const TextControl = withCharacterCounter( BaseTextControl );
 const TextareaControl = withCharacterCounter( BaseTextareaControl );
 
 export class Articles extends Component {
+  constructor(props) {
+    super(props);
+    this.instanceName = `planet4-blocks-articles-${ uuidv4() }`;
+  }
+
+  componentDidMount() {
+    this.lazyLoadInterval = setInterval(() => {
+      if (document.querySelector(`.${this.instanceName} section.articles-block`)) {
+        this.updateLazyLoad();
+        clearInterval(this.lazyLoadInterval);
+      }
+    }, 250);
+  }
+
+  componentDidUpdate() {
+    this.updateLazyLoad();
+  }
+
+  updateLazyLoad() {
+    if (window.lazyLoad) {
+      window.lazyLoad.update();
+    }
+  }
+
   renderEdit() {
     const {__} = wp.i18n;
 
@@ -147,6 +171,7 @@ export class Articles extends Component {
         }
         <Preview showBar={ this.props.isSelected }>
           <ServerSideRender
+            className={ this.instanceName }
             block={ 'planet4-blocks/articles' }
             attributes={ this.props.attributes }>
           </ServerSideRender>


assets/src/blocks/Articles/Articles.js
```diff
@@ -11,12 +11,36 @@ import TagSelector from '../../components/TagSelector/TagSelector';
...
+    this.instanceName = `planet4-blocks-articles-${ uuidv4() }`;
+  }
+
+  componentDidMount() {
+    this.lazyLoadInterval = setInterval(() => {
+      if (document.querySelector(`.${this.instanceName} section.articles-block`)) {
+        this.updateLazyLoad();
+        clearInterval(this.lazyLoadInterval);
+      }
+    }, 250);
```
Inwerpsel, line 31: Too bad `ServerSideRender` doesn't provide a callback when rendering is complete, though I understand why they don't provide it as it's probably not a good idea in most cases, except in this one it would probably be somewhat better than polling. But this indeed looks like the only way to achieve what we want :+1:

Inwerpsel: Another option could be to extend the `ServerSideRender` component, as it's pretty easy to implement the `onChange` handler, like here https://github.com/WordPress/gutenberg/issues/8379#issuecomment-446815846 . That would make it require a lot less code to apply the lazy loading update to multiple blocks. But maybe it won't work as it's only called when the `ServerSideRender` component updates, which is just before it renders.

Inwerpsel:
Screenshot from 2020-06-23 10-48-07
Looks like changing our circleci jobs confused github.

comzeradd:
>  Looks like changing our circleci job names confused github. Maybe it's solved if you amend with no changes?

Probably because this PR was opened before the change. @lithrel can you rebase?

lithrel:
Rebased 👍 

> Looks good, just want to check if you may have explored if it's possible to wrap ServerSideRender to add a change listener. If that would work it could be a lot easier to apply the same fix in multiple blocks.

I didn't, mainly because we are planning to drop usage of ServerSideRender

assets/src/blocks/Articles/Articles.js
pablocubico: @lithrel  @Inwerpsel  I did a few tests to wrap it but couldn't get to a decent solution, and since we are rolling out the ServerSideRender in the near future, and this block will be converted to WYSIWYG, this looks like a reasonable tradeoff for now. 

Maybe a more elegant option for the WYSIWYG blocks would be to use one of the lazy-load React components out there like these ones:
https://www.npmjs.com/package/react-lazyload
https://github.com/Aljullu/react-lazy-load-image-component

I guess we can go ahead with this fix in the meantime and create a ticket to revisit the lazy-load implementation towards a "Reactier" (?) approach for the WYSIWYG blocks.

mleray:
Is this PR still relevant now that the Articles block is being converted to WYSIWYG and WordPress is about to include lazy loading by default? 🤔

lithrel:
> Is this PR still relevant now that the Articles block is being converted to WYSIWYG and WordPress is about to include lazy loading by default? 🤔

I'd say it is not relevant anymore, and your PR https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/314 should be considered as solving this issue.

PR #293, opened 2020-05-29
PLANET-5060 Make Take Action covers and boxouts links

mleray:
This PR is to explore the possibility of making the entire Take Action card a link (`<a>` tag) in order to achieve better accessibility. See [this PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/291) for more details (in particular the comment from @comzeradd)

The only issue is that `<a>` tags cannot have other `<a>` tags as children (https://www.w3.org/TR/html401/struct/links.html#h-12.2.2), so if we are going to choose this approach, the Take Action card's other links need to be replaced with other HTML tags. It's not a problem for the button as it anyway redirects to the same page as the rest of the card, but we have an issue for the tags (#oil for example) that link to other pages... I think that for these ones we don't have any other choice than using JavaScript.

### DIFF
--- a/assets/src/styles/blocks/Covers/TakeActionCovers.scss
+++ b/assets/src/styles/blocks/Covers/TakeActionCovers.scss
@@ -74,6 +74,11 @@
   background-size: cover;
   background-position: top;
   color: $grey-80;
+  display: block;
+
+  .external-icon {
+    display: none;
+  }
 
   @include medium-and-up {
     flex-basis: 48%;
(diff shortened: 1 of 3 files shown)


comzeradd:
Maybe we could use a trick like this: https://codepen.io/pwkip/pen/oGMZjb
It uses an absolute position for the card link to overlay it on top of everything and the links inside still work. This also removes the problem of having link inside links (at least on the markup). Not sure how applicable to our cards markup.

mleray:
@comzeradd Nice find, I'll try it out!

PR #295, opened 2020-06-05
PLANET-5114 Add css variables for easier child-theme customisation

mleray:
This is a non-exhaustive list of added variables that child-themes could use for easier customisation of Planet4 (see [PLANET-5114](https://jira.greenpeace.org/browse/PLANET-5114) for more details). The ones listed below are already in the code and could be used right away, but maybe they need to be renamed?

#### Campaign colors
- Navigation bar background color: [--campaign_nav_color](https://github.com/greenpeace/planet4-styleguide/blob/master/src/layout/_navbar.scss#L83)
- Footer background color: [--footer_color](https://github.com/greenpeace/planet4-styleguide/blob/master/src/layout/_footer.scss#L47)
- Footer links color: [--footer_links_color](https://github.com/greenpeace/planet4-styleguide/blob/master/src/layout/_footer.scss#L68)

#### Spreadsheet colors
- Spreadsheet header background color: [--spreadsheet-header-background](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Spreadsheet.scss#L27)
- Spreadsheet even rows background color: [--spreadsheet-even-row-background](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Spreadsheet.scss#L41)
- Spreadsheet odd rows background color: [--spreadsheet-odd-row-background](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Spreadsheet.scss#L51)

Corresponding styleguide PR: https://github.com/greenpeace/planet4-styleguide/pull/59
Corresponding master-theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1117
Related ticket: https://jira.greenpeace.org/browse/PLANET-5104

### DIFF
--- a/assets/src/styles/blocks/Covers/CampaignCovers.scss
+++ b/assets/src/styles/blocks/Covers/CampaignCovers.scss
@@ -105,7 +105,6 @@
         text-align: center;
         font-size: $font-size-md;
         font-family: $roboto;
-        background-color: transparent;
         color: $yellow;
         font-weight: 500;
         position: absolute;
@@ -128,8 +127,6 @@
 
         &:hover {
           text-decoration: none;
-          background-color: transparent;
-          color: $yellow;
         }
       }
     }
--- a/assets/src/styles/blocks/Covers/CampaignCovers.scss
+++ b/assets/src/styles/blocks/Covers/CampaignCovers.scss
@@ -105,6 +105,7 @@
         text-align: center;
         font-size: $font-size-md;
         font-family: $roboto;
+        background-color: transparent;
         color: $yellow;
         font-weight: 500;
         position: absolute;
@@ -127,6 +128,8 @@
 
         &:hover {
           text-decoration: none;
+          background-color: transparent;
+          color: $yellow;
         }
       }
     }
--- a/assets/src/styles/editorOverrides.scss
+++ b/assets/src/styles/editorOverrides.scss
@@ -74,7 +74,7 @@
   text-align: center;
   text-decoration: none;
   text-transform: uppercase;
-  color: $white;
+  color: var(--btn_color, $white);
   font-weight: 500;
   border-radius: 0;
   border: 1px solid transparent;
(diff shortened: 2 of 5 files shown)


assets/src/styles/blocks/Covers/CampaignCovers.scss
```diff
@@ -105,7 +105,8 @@
         text-align: center;
         font-size: $font-size-md;
         font-family: $roboto;
-        color: $yellow;
+        background-color: var(--campaign-covers-block_yellow-cta_background_color, transparent);
```
mleray, line 108: This should probably be renamed? `yellow-cta` is a bit too "color specific", and it really doesn't make sense if NROs are going to change it... Also it's overriden very often in campaign themes, sometimes directly like [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/campaigns/themes/_theme_forest.scss#L159)) or using mixins

Inwerpsel: Yeah especially for campaigns there are many complicated cases like this one. Maybe for now it's easier if we leave them out and start with the more clear-cut ones.

Though ultimately it hints at the issue we're addressing with the css variables. If we condense it to what we really need it could be
```
.cta {
  // There is actually only the yellow-cta class on this block currently, so we probably better drop the color prefix there.
  background-color: var(--covers-block-cta-background-color, $yellow);
}
```
where campaigns would then just have the right value specified in their variables.

Inwerpsel: Something we can consider here too is that you can fallback to another variable. That would allow more granular control if needed but still apply the general cta color variable otherwise.
```
  background-color: var(--cta-background-color, var(--covers-block-cta-background-color, $yellow));
```

mleray: Maybe we can leave this one out for now and create a separate ticket to take care of it in the future? (renaming `yellow-cta` to something more generic and find a way make it more customisable for child-themes)

assets/src/styles/editorOverrides.scss
```diff
@@ -74,7 +74,7 @@
   text-align: center;
   text-decoration: none;
   text-transform: uppercase;
-  color: $white;
+  color: var(--btn-color, $white);
```
Inwerpsel, line 77: Shouldn't this also be introduced for the frontend in [the styleguide PR](https://github.com/greenpeace/planet4-styleguide/pull/59/files)?

mleray: Good one @Inwerpsel ! I'll fix this now

PR #299, opened 2020-06-09
PLANET-5116 - Fix block vertical spacing

mleray:
See https://jira.greenpeace.org/browse/PLANET-5116 (UAT already passed)
- Corresponding styleguide PR: https://github.com/greenpeace/planet4-styleguide/pull/61
- Corresponding master-theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1109

### DIFF
--- a/assets/src/styles/blocks/core-overrides/Image.scss
+++ b/assets/src/styles/blocks/core-overrides/Image.scss
@@ -2,15 +2,26 @@
   position: relative;
   width: auto;
   max-width: 100%;
+  margin-top: $space-xs;
+  margin-bottom: $space-md;
 
-  figure:not(.alignleft):not(.alignright) {
-    margin-top: $space-xs;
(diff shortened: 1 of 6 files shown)


assets/src/styles/blocks/core-overrides/Image.scss
```diff
@@ -3,6 +3,16 @@
   width: auto;
   max-width: 100%;
 
+  figure:not(.alignleft):not(.alignright) {
```
comzeradd, line 6: Do you think it's possible to make this rule less specific? Because it would be very difficult to be overridden by child themes. Maybe add these rules for all `figure` and then override for `.alignleft` and `.alignright`?

mleray: @comzeradd Updated!

PR #302, opened 2020-06-10
PLANET-5186: Switch script version to filectime

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5186

Script version is not automatically updated when we make changes, which leads to cache problems if we forget it.

## Fix

Switch from hardcoded version to `filectime`.

Versions were found with regex `[0-9]{1,2}\.[0-9]{1,2}`, and by looking at `wp_enqueue_style` and `wp_enqueue_script` usage.

I took the liberty to regroup those calls in a static function, to limit the number of parameters used. If you feel like it's too big of a change and I went too far, I can roll this back :)

## Test

On the editor side while editing a page, in the network calls or in html source, we should see a call to `planet4-plugin-gutenberg-blocks/assets/build/editorIndex.js?ver=1591791157` with a timestamp as version number. You can filter by `ver=159` to see those calls.  
No network call related to this plugin should be a 404, on the public and the admin sides.

#### We need to talk...

... about `Using short ternaries is not allowed (WordPress.PHP.DisallowShortTernary.Found)`. It's making me sad, I feel like writing PHP4 😢 _Just read this https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/ , still not convinced.  
edit: this comment https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/#comment-37128 actually makes sense, but I don't feel it applies to us._

### DIFF
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -191,7 +191,7 @@ class Articles extends Base_Block {
 
 		// Enqueue js for the frontend.
 		if ( ! $this->is_rest_request() ) {
-			wp_enqueue_script( 'load-more', P4GBKS_PLUGIN_URL . 'public/js/load_more.js', [ 'jquery' ], '0.3', true );
+			\P4GBKS\Loader::enqueue_local_script( 'load-more', 'public/js/load_more.js', [ 'jquery' ] );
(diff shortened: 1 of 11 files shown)


comzeradd:
We could take the opportunity and fix all versioned assets in this repo.

lithrel:
PR updated with changes across this plugin and text updated.

PR #308, opened 2020-06-22
Planet 5151 Merge EN plugin functionality into blocks repo

sagarsdeshmukh:
To make a Review bit easier making this PR. (Thanks to @Inwerpsel )

please add further review comments in this PR instead of #303 .

old PR - [Planet 5151 Migrate EN forms plugin to blocks plugin [follow up of #297 PR] #303](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/303)


classes/controller/menu/class-enform-post-controller.php
```diff
@@ -0,0 +1,443 @@
...
+	/**
+	 * Load assets for new/edit form page.
+	 */
+	public function load__new_page_assets() {
+		wp_enqueue_script( 'jquery-ui-core' );
+		wp_enqueue_script( 'jquery-ui-sortable' );
+		wp_enqueue_script( 'jquery-ui-dialog' );
+		wp_enqueue_script( 'jquery-ui-tooltip' );
+		wp_enqueue_style( 'wp-jquery-ui-dialog' );
+		wp_enqueue_style( 'p4en_admin_style_blocks', P4GBKS_ADMIN_DIR . 'css/admin_en.css', [], '0.1' );
```
lithrel, line 379: Since https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/302 scripts versions should be replaced by [`\P4GBKS::file_ver( $filepath )`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d2fcc534d317dba2c5967deef37d89decbe454a7/classes/class-loader.php#L587).
For `wp_enqueue_script`, you can use [`\P4GBKS::enqueue_local_script('p4en_admin_style_blocks', 'css/admin_en.css')`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d2fcc534d317dba2c5967deef37d89decbe454a7/classes/class-loader.php#L607) instead.

classes/controller/class-enform-fields-list-table.php
```diff
@@ -100,7 +100,7 @@ public function get_columns() {
 	 *
 	 * @return string Content for column.
 	 */
-	protected function column_default( $item, $column_name ) {
+	protected function column_default( $item, $column_name ): string {
```
lithrel, line 103: it won't return anything if we don't give the proper `$item` parameter, so either add `return ''` at the end or as `default`, or `return null` and change the return type to `?string`
Same for `Enform_Fields_List_Table::column_default()`

classes/controller/menu/class-controller.php
```diff
@@ -49,5 +61,136 @@ public function load() {
...
+		 *
+		 * @return mixed Array if validation is ok, false if validation fails.
+		 */
+		public function valitize( $settings ): array {
+			if ( $this->validate( $settings ) ) {
+				$this->sanitize( $settings );
+				return $settings;
+			} else {
+				return $settings;
+			}
```
lithrel, line 78:
```php
			if ( $this->validate( $settings ) ) {
				$this->sanitize( $settings );
			}
			return $settings;
```

templates/settings.twig
```diff
@@ -1,19 +1,67 @@
...
+	</div>
+
+	{{ fn( 'settings_errors' ) }}
+
+	<form id="p4en_main_settings_form" name="p4en_main_settings_form" method="post" action="options.php">{# ../../../../wp-admin/options.php #}
+
+		{{ fn( 'settings_fields', 'p4en_main_settings_group' ) }}
+		{{ fn( 'do_settings_sections', 'p4en_main_settings_group' ) }}
 
-    <form id="p4bks_main_settings_form" name="p4bks_main_settings_form" method="post">
```
Inwerpsel, line 7: Were these settings unused?

sagarsdeshmukh: yes, it is.

classes/blocks/class-base-block.php
```diff
@@ -37,7 +37,11 @@ public function render( $attributes ) {
 
 		$data = $this->prepare_data( $attributes );
 
-		\Timber::$locations = P4GBKS_PLUGIN_DIR . '/templates/blocks';
+		$template_path = '/templates/blocks';
+		if ( 'enform' === static::BLOCK_NAME ) {
+			$template_path = '/templates/blocks/enform';
```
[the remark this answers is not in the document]
sagarsdeshmukh: Not an exception, I thought its good to group the enform block related template files in one folder. do you think we should keep them at same level of other blocks template files?

classes/blocks/class-enform.php
```diff
@@ -2,47 +2,41 @@
...
 	 * Custom meta field where fields configuration is saved to.
 	 */
-	const FIELDS_META = 'p4enform_fields';
+	private const FIELDS_META = 'p4enform_fields';
 
 	/**
 	 * Class Loader reference
 	 *
-	 * @var \P4GEN\Loader $loader;
+	 * @var \P4GBKS\Loader $loader;
```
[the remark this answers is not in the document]
sagarsdeshmukh: Good point, I removed the dependency of `$loader` instance from `ENForm` class

classes/blocks/class-base-block.php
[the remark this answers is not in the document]
Inwerpsel: Ah I see, this is needed because `Base_Block` uses the `const BLOCK_NAME` to load the template. But if we move only the main `enblock.twig` file outside of the `enform` directory, we can keep all other enform templates grouped inside of that directory and we wouldn't need this block specific code here.

classes/blocks/class-enform.php
```diff
@@ -2,158 +2,144 @@
...
-						'type'    => 'string',
-						'default' => '',
-					],
-					'description'                   => [
-						'type'    => 'string',
-						'default' => '',
-					],
-					'campaign_logo'                 => [
-						'type' => 'boolean',
+		if ( ! WP_Block_Type_Registry::get_instance()->is_registered( 'planet4-blocks/enform' ) ) {
```
Inwerpsel, line 43: If we don't need [these two lines](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/308/files#diff-bfb6ea67fe59d361747a005ff3477325R136-R137) to be executed, we could instead do
```
if ( WP_Block_Type_Registry::get_instance()->is_registered( 'planet4-blocks/enform' ) ) {
    return;
}
```
and it wouldn't modify the history of the block definition.

classes/class-loader.php
```diff
@@ -10,6 +10,8 @@
 
 use WP_CLI;
 use P4GBKS\Command\Controller;
+use P4GBKS\Controllers\Ensapi_Controller as Ensapi;
```
Inwerpsel, line 13: If it's not needed for some reason, can you remove the alias `as Ensapi`?

classes/controller/menu/class-controller.php
```diff
@@ -49,5 +61,134 @@ public function load() {
...
+		}
+
+		/**
+		 * Validates the settings input.
+		 *
+		 * @param array $settings The associative array with the settings that are registered for the plugin.
+		 *
+		 * @return bool
+		 */
+		abstract public function validate( $settings ) : bool;
```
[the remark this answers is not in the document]
sagarsdeshmukh: Yes I agree with you. I cleanup the code and keep it only in en-settings controller. 👍

classes/controller/class-enform-fields-list-table.php
```diff
@@ -100,14 +100,15 @@ public function get_columns() {
...
-	protected function column_default( $item, $column_name ) {
+	protected function column_default( $item, $column_name ): ?string {
 		switch ( $column_name ) {
 			case 'id':
 			case 'name':
 			case 'property':
 			case 'tag':
 				return $item[ $column_name ];
 		}
+		return '';
```
Inwerpsel, line 111: You can make this return `null` instead explicitly, then the return value will be like before.

classes/controller/class-ensapi-controller.php
```diff
@@ -128,35 +128,37 @@ public function get_pages( $params = [
...
-		if ( ! $response['body'] ) {
-			$url = add_query_arg(
-				[
-					'type'   => strtolower( $params['type'] ),
-					'status' => $params['status'],
-				],
-				self::ENS_PAGES_URL
-			);
+		if ( $response['body'] ) {
+			return json_decode( $response['body'], true );
```
Inwerpsel, line 132: Since we're already changing this part, now we can make this a separate variable as it's either returned immediately, or would be overwritten [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/308/files#diff-9fe83910b67fb870e125aead11d0e5d2R144). It's easier to follow the code if these are separate, so we could make it something like 
```
$from_cache= get_transient( 'ens_pages_response_' . implode( '_', $params ) );
if ( $from_cache ) {
	return json_decode( $from_cache, true );
}
```

classes/controller/class-ensapi-controller.php
```diff
@@ -266,30 +271,33 @@ public function get_supporter_fields() {
 	 */
 	public function get_supporter_questions() {
 		$response['body'] = get_transient( 'ens_supporter_questions_response' );
-		if ( ! $response['body'] ) {
-			$url = self::ENS_SUPPORTER_URL . '/questions';
+		if ( $response['body'] ) {
```
Inwerpsel, line 274: We can do the same as here https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/308/files#r449458709

classes/controller/class-ensapi-controller.php
```diff
@@ -302,30 +310,33 @@ public function get_supporter_questions() {
 	 */
 	public function get_supporter_question_by_id( $question_id ) {
 		$response['body'] = get_transient( 'ens_supporter_question_by_id_response_' . $question_id );
-		if ( ! $response['body'] ) {
-			$url = self::ENS_SUPPORTER_URL . '/questions/' . $question_id;
+		if ( $response['body'] ) {
```
Inwerpsel, line 313: We can do the same as here https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/308/files#r449458709

en-plugin/classes/controller/menu/class-pages-datatable-controller.php
```diff
@@ -1,237 +0,0 @@
-<?php
```
Inwerpsel, line 1: Was this unused? Can't find it in the new code.

sagarsdeshmukh: After rechecking on it, found unused. hence removed with all references.

classes/controller/menu/class-controller.php
```diff
@@ -49,5 +48,24 @@ public function load() {
...
 		abstract public function create_admin_menu();
+
+		/**
+		 * Get underscore template from filesystem.
+		 *
+		 * @param string $template Template name.
+		 *
+		 * @return bool|string
+		 */
+		protected function get_template( $template ) {
```
Inwerpsel, line 59: Do we need to add this function? It looks like it's not used anywhere.

sagarsdeshmukh: It's not in use. good catch 👍

PR #314, opened 2020-06-29
PLANET-4801 WYSIWYG implementation for Articles block

mleray:
See https://jira.greenpeace.org/browse/PLANET-4801

**Tasks:**
- Convert the Title and Description, and Button Text into RichText components.
- Move Link, Open in new tab, Articles Count, Tags, Post Types and Ignore categories options to the sidebar.
- Convert to JSX rendering

More details in the [doc](https://docs.google.com/document/d/1RB4mOYWzUZM7uWYcyllfSdIGCQat4n04nrsFXLGIPBE/edit#heading=h.z78wqkd0dndk)


assets/src/blocks/Articles/ArticlesEditor.js
```diff
@@ -0,0 +1,172 @@
...
+export class ArticlesEditor extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      total_pages: 0
+    };
+    this.toAttribute = this.toAttribute.bind(this);
+  }
+
+  toAttribute(attributeName) {
```
Inwerpsel, line 30: It makes sense to add this as an instance method. Just wondering, since this function will be identical in each block, if there's a way we could write it once instead of in each block?

Inwerpsel: Another option would be to handle this in a function. That would also allow us to prevent specifying the attribute name twice. I made a branch with such a function (might need a bit of work still). https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/compare/for-attribute?expand=1

classes/rest/class-rest-api.php
```diff
@@ -141,6 +142,74 @@ public static function endpoints(): void {
 				],
 			]
 		);
+		/**
+		 * Endpoint to retrieve the latest posts for the Articles block
+		 */
```
Inwerpsel, line 147: Maybe to make it easier to review, you could rename the function in `class-articles.php` and then call it here instead of moving the code to here? That way it will be a lot easier to see what actually changed about the code.

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,134 @@
...
+const { apiFetch } = wp;
+const { addQueryArgs } = wp.url;
+
+export class ArticlesFrontend extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      posts: [],
+      total_pages: 0,
+      dataset: {},
```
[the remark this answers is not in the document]
mleray: Indeed! Removing it now 👍

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,134 @@
...
+
+    const args = {
+      article_count,
+      post_types,
+      posts,
+      tags,
+      ignore_categories
+    };
+
+    if (page) args.offset = page * article_count;
```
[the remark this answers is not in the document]
mleray: Fixed!

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,134 @@
...
+      tags.length !== prevProps.tags.length ||
+      posts.length !== prevProps.posts.length ||
+      post_types.length !== prevProps.post_types.length ||
+      ignore_categories !== prevProps.ignore_categories
+    ) {
+      this.loadArticles();
+    }
+  }
+
+  loadArticles(page) {
```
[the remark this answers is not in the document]
mleray: Done! 👍

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,155 @@
...
+          <div className="container">
+            {!isEditing &&
+              <header>
+                <h2 className="page-section-header">{article_heading || __('Latest Articles', 'p4ge')}</h2>
+              </header>
+            }
+            {articles_description && !isEditing &&
+              <div className="page-section-description">{articles_description}</div>
+            }
+            <div className="article-list-section clearfix">
```
Inwerpsel, line 117: Looks like this part is the only thing that is rendered when `isEditing` if false. So maybe it's easier to extract this, so that we can include it separately when editing, instead of the entire `ArticlesFrontend`? Then you could remove all the `!isEditing` checks in this component.

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+        post_date,
+        post_excerpt
+      },
+      isEditing
+    } = this.props;
+
+    const date = new Date(post_date);
+
+    let articleClassName = "article-list-item";
+    if (tags && tags.length > 0) {
```
Inwerpsel, line 72: If you do any array method, you never need to check if the length is 0.

Inwerpsel: I think in this case you also don't need to check ` if (tags)`, as the default value of the `tags` attribute is an empty array.

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+      tags.forEach(tag => articleClassName += ` ${tag.slug}`);
```
Inwerpsel, line 73: The tag classes need to be prefixed with `tag-`.

Inwerpsel: You can declare `articleClassName` as a const if you use reduce instead
```
const tagClasses = tags.reduce((classname, tag)=> classname + ` tag-${tag.slug}`, '')
```

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+    if (isEditing) articleClassName += ' editing';
```
Inwerpsel, line 75: Do we need to set the editing class inside here? Maybe it's easier to set that class on a wrapping element? Maybe here https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/940c184d14d572892b1fa3a9a8400ab54b107cee/assets/src/blocks/Articles/ArticlesEditor.js#L171-L171 ?

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+
+    return (
+      <article className={articleClassName} >
+        {thumbnail_ratio < 1 ?
+          <div className="article-list-item-image">
```
Inwerpsel, line 80: Could make sense to move the thumbnail part into a function so we don't lose overview in here? I think you could also avoid a lot of duplication inside of it, as the entire a tag is identical in both cases.

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+        <span className="article-list-item-author">{__('by', 'planet4-blocks')}{' '}
+          {(author_override || isCampaign) ?
+            author_name
+            :
+            <a href={author_url}>{author_name}</a>
+          }
+        </span>
+      )
+    } else if (author) {
+      return (
```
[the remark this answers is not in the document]
Inwerpsel: I think we can also avoid some duplication here.

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        <div className="article-list-item-body">
+          {(tags || page_type) &&
+            <div className="article-list-item-tags top-page-tags">
+              {page_type &&
+                this.getPageTypesTags(page_type, page_type_link)
+              }
+
+              {tags && tags.length > 0 &&
```
Inwerpsel, line 110: Here we still need the length check, but we can remove `tags &&`

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,152 @@
...
+    };
+
+    if (page) args.offset = page * article_count;
+
+    if (postType === 'post') args.exclude_post_id = postId;
+
+    const queryArgs = {
+      path: addQueryArgs('/planet4/v1/get-articles', args)
+    };
+    apiFetch(queryArgs).then(result => {
```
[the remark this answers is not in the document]
mleray: Done 👍

assets/src/blocks/Articles/ArticlePreview.js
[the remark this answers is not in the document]
mleray: Done 👍

assets/src/blocks/Articles/ArticlePreview.js
mleray: Done 👍

assets/src/blocks/Articles/ArticlePreview.js
mleray: Done 👍

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,134 @@
...
+          total_pages
+        });
+        if (setTotalPages) setTotalPages(total_pages);
+      }
+    });
+  }
+
+  render() {
+    const {
+      article_heading,
```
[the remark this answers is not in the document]
mleray: This is a bit complicated to change, because the frontend is rendered via [this file](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/frontendIndex.js) and changing the behavior there would mean changing it for other WYSIWYG blocks too...

assets/src/blocks/Articles/ArticlesFrontend.js
mleray: I'm not sure how to handle this properly, because we load the articles in this component (they're not props) and then the button to load more posts is there too... An idea maybe would be to move the `loadArticles` function to a separate file? Then we could use it both for loading the initial articles (needed both in the frontend and in the editor) and for loading more articles (needed only in the frontend) But then we would need to pass down all the arguments as params for the function instead of using props, I guess that's fine if we use an object 🤔

assets/src/blocks/Articles/ArticlePreview.js
mleray: Nice! 👍

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -0,0 +1,145 @@
...
+        author,
+        author_override,
+        author_url
+      },
+      isCampaign
+    } = this.props;
+
+    if (author_name) {
+      return (
+        <span className="article-list-item-author">{__('by', 'planet4-blocks')}{' '}
```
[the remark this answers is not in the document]
mleray: I don't think it's a problem to keep `isCampaign` here 🤔 The author info is taken care of in the php, I'm not sure it's a good idea to modify it in the frontend afterwards...? Also this makes it clearer in my opinion.

assets/src/blocks/Articles/ArticlesEditor.js
Inwerpsel: Guess we can look at that later :+1:

classes/blocks/class-articles.php
```diff
@@ -162,110 +115,52 @@ public function prepare_data( $fields ): array {
...
-			$args = $this->filter_posts_for_tag_page( $fields );
+		$fields_diff = count( array_diff( [ 'post_types', 'posts' ], array_keys( $fields ) ) );
+		if ( is_tag() && ! empty( $fields['tags'] ) && 2 === $fields_diff ) {
+			$args = self::filter_posts_for_tag_page( $fields );
 		} elseif ( ! empty( $fields['post_types'] ) ||
 				! empty( $fields['tags'] ) ||
-				! empty( $exclude_post_id ) ) {
-			$args = $this->filter_posts_by_page_types_or_tags( $fields );
+				! empty( $fields['exclude_post_id'] ) ) {
+			$args = self::filter_posts_by_page_types_or_tags( $fields );
```
Inwerpsel, line 124: Can you also change the functions to static? Calling non static functions with self:: works, but can cause warnings in logs or IDE.

assets/src/blocks/Articles/ArticlesFrontend.js
Inwerpsel: >  An idea maybe would be to move the loadArticles function to a separate file? 

:+1: Then this component will also be a lot simpler.

> Then we could use it both for loading the initial articles (needed both in the frontend and in the editor) and for loading more articles

Why do we need a different call for the initial page and the subsequent ones? I mean, it looks like the count in the first call is only used to see if we need to show the button for the next posts. But we could use also use that for any page.

assets/src/components/PostSelector/PostSelector.js
```diff
@@ -1,36 +1,38 @@
...
-    this.handleChange = this.handleChange.bind( this );
-    this.getValue = this.getValue.bind( this );
+    this.handleChange = this.handleChange.bind(this);
+    this.getValue = this.getValue.bind(this);
     this.state = {
       posts: [],
     };
+  }
 
+  componentDidMount() {
```
Inwerpsel, line 20: :+1:

assets/src/styles/blocks/Articles.scss
```diff
@@ -27,6 +23,21 @@
     }
   }
 
+  .editing & {
```
Inwerpsel, line 26: Maybe it's better to make a separate file for the admin scss, to help keep the frontend bundle size small.

mleray: There is actually already one, I missed it! Good one 👍

assets/src/blocks/Articles/ArticlesFrontend.js
Inwerpsel: https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook

assets/src/styles/blocks/ArticlesEditor.scss
```diff
@@ -1,3 +1,33 @@
 .ignore-categories-wrapper {
   margin-top: 10px;
 }
+
+.articles-editor {
+  div.btn-secondary.article-load-more.btn.btn-block {
```
Inwerpsel, line 6: Do we need the `div` and all classes in this selector for specificity? Would be nicer if it was just `.article-load-more`

assets/src/styles/editorOverrides.scss
```diff
@@ -213,3 +213,18 @@ input.describe[type=text][data-setting=caption] {
   font-size: 13px;
   color: red;
 }
+
+// Articles block overrides
```
Inwerpsel, line 217: Maybe we can include these change in assets/src/styles/blocks/ArticlesEditor.scss too?

classes/blocks/class-articles.php
```diff
@@ -275,7 +170,7 @@ public function load_more() {
 	 *
 	 * @return array
 	 */
-	private function populate_post_items( $posts ) {
+	public function populate_post_items( $posts ) {
```
Inwerpsel, line 173: These can still be private as they're only used in the class. What I meant was adding `static` keyword to it :slightly_smiling_face:

Inwerpsel: Otherwise the usages look like this :stuck_out_tongue: 
Screenshot from 2020-07-21 10-36-28

assets/src/blocks/Articles/ArticlesFrontend.js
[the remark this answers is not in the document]
Inwerpsel: Maybe we can instead ensure that the frontend block only gets the attributes? As that will be the only data it should have available in the frontend. That way we also keep editor and frontend code separated. If the editor needs to render something else than the frontend, it's probably better to split the frontend node in multiple parts. Then we can just include only the parts we need in the editor, and we don't need the conditionals in our frontend component.

classes/blocks/class-articles.php
mleray: 🙈 Sorry I guess I read your comment too fast!

assets/src/styles/editorOverrides.scss
mleray: Indeed 👍

assets/src/styles/blocks/ArticlesEditor.scss
mleray: It was needed before to make sure that it wouldn't apply in the frontend, but now indeed since these styles are only loaded in the editor it's no longer necessary 👍

assets/src/blocks/Articles/ArticlesFrontend.js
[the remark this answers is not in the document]
mleray: That would be ideal 😄 I've already removed the `isEditing`, but there is actually an issue for `postType` since it's passed in the editor via `withSelect`... It's not present in the frontend 😱 I guess we should add it to the attributes...?

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -0,0 +1,69 @@
...
+      articles_description,
+      read_more_text,
+      read_more_link,
+      button_link_new_tab,
+    } = this.props;
+
+    const { totalPages, page } = this.state;
+
+    return (
+      <Fragment>
```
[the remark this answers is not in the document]
Inwerpsel: Haha, didn't notice this pending comment, not applicable anymore :slightly_smiling_face:

PR #315, opened 2020-06-30
PLANET-5151 Remove unneeded class_exists check, add newline at EOF

sagarsdeshmukh:
Related to https://jira.greenpeace.org/browse/PLANET-5151


Inwerpsel:
> I guess if you do a `git mv en-plugin/classes/controller/menu/class-settings-controller.php en-plugin/classes/controller/menu/class-en-settings-controller.php` before modifying it, that would display even less diff (not sure though) ?

`git mv` does not guarantee that git will consider it as a move if you change the file afterwards (see https://stackoverflow.com/a/3010099/4961158). If you add those changes to the same commit, git still just checks for each deleted file if a new file was added with the same content, regardless of whether you used `git mv`, which is basically just a shorthand for `mv old new; git add new; git rm old`.

So in this case, since we unwrapped the code, it didn't pick it up: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/315/files#diff-3ff3b63fff519fe4a0d0b8c3189a25e5

Inwerpsel:
> `git mv` does not guarantee that git will consider it as a move if you change the file afterwards (see https://stackoverflow.com/a/3010099/4961158).
Important detail here: git considers this not at the time you commit, but when viewing the history (log, but also when performing a merge). So you never really "lose history", rather you can lose the ability of git to detect it by default.

PR #319, opened 2020-07-07
PLANET-5293 EN Form Block: Make "Goal" field mandatory

sagarsdeshmukh:
- Add `<Br>` tag before the select goal field (it will help to highlight goal as a separate field)
- Add `--- Select Goal ---` as a Default option in goals dropdown
- Add `"Select Goal" field is required!` validation 

image

### DIFF
--- a/assets/src/blocks/ENForm/ENForm.js
+++ b/assets/src/blocks/ENForm/ENForm.js
@@ -78,11 +78,12 @@ export class ENForm extends Component {
                 { __( 'Check your EngagingNetworks settings!', 'planet4-engagingnetworks-backend' ) }
               </InlineFormFeedback>
           }
-
+          <br />
           <SelectControl
             label={__( '- Select Goal -', 'planet4-engagingnetworks-backend' )}
             value={this.props.enform_goal}
             options={[
+              { label: __( '--- Select Goal ---', 'planet4-engagingnetworks-backend' ), value: 'not set' },
               { label: 'Petition Signup', value: 'Petition Signup' },
               { label: 'Action Alert', value: 'Action Alert' },
               { label: 'Contact Form', value: 'Contact Form' },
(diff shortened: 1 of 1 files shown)


assets/src/blocks/ENForm/ENForm.js
```diff
@@ -78,11 +78,12 @@ export class ENForm extends Component {
                 { __( 'Check your EngagingNetworks settings!', 'planet4-engagingnetworks-backend' ) }
               </InlineFormFeedback>
           }
-
+          <br />
           <SelectControl
             label={__( '- Select Goal -', 'planet4-engagingnetworks-backend' )}
             value={this.props.enform_goal}
             options={[
+              { label: '--- Select Goal ---', value: 'not set' },
```
lithrel, line 86: maybe we could make the label translatable ? So that it can correspond to the label

assets/src/blocks/ENForm/ENForm.js
```diff
@@ -312,6 +313,9 @@ export class ENForm extends Component {
       if ( undefined === this.props.en_page_id || 0 === this.props.en_page_id ) {
         validationMessage.push( __( '"Engaging Network Live Pages" field is required!', 'planet4-engagingnetworks-backend' ));
       }
+      if ( undefined === this.props.enform_goal || 'not set' === this.props.enform_goal ) {
```
lithrel, line 316: I didn't look at what's currently in the database for existing forms (just broke my dev env 😭 ), but if we happen to have an empty `enform_goal` field (so, not undefined or 'not set', just empty string) we should add it to this condition

sagarsdeshmukh: Good point, I will add it 👍

sagarsdeshmukh:
> selectControl

I am a newbie in react too, but I guess native `selectControl` won't offer a option for 'defaultValue'. ([Ref.](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/select-control))

sagarsdeshmukh:
> Questions from a non-react guy: shouldn't we add a 'defaultValue' to the selectControl ?
> 
> Nothing blocking 👍
I am a newbie in react too, but I guess native selectControl won't offer a option for 'defaultValue'. ([Ref.](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/select-control))

PR #323, opened 2020-07-09
PLANET-5263: Decrease H1 size on mobile

pablocubico:
Pinning styleguide branch.

Ref: https://jira.greenpeace.org/browse/PLANET-5263


mleray:
This is no longer needed since the styleguide is now automatically updated via CircleCI

PR #326, opened 2020-07-15
PLANET-5014: Prevent the skip-links class from pushing page-content down

oekeur:
Because of typography changes using a too general css selector the skip-links class was pushing content down on some pages.
Excluding .skip-links now from displaying as inline-block and remove some duplicated rules

### DIFF
--- a/assets/src/styles/blocks/core-overrides/BlockSpacing.scss
+++ b/assets/src/styles/blocks/core-overrides/BlockSpacing.scss
@@ -25,13 +25,6 @@
 .post-details > h2,
 .post-details > ul {
   display: inline-block;
-  margin-top: $space-xs;
-  margin-bottom: $space-md;
-
-  @include large-and-up {
-    margin-top: $space-md;
-    margin-bottom: $space-lg;
-  }
 }
 
 .page-template > ul,
--- a/assets/src/styles/blocks/core-overrides/BlockSpacing.scss
+++ b/assets/src/styles/blocks/core-overrides/BlockSpacing.scss
@@ -21,9 +21,9 @@
 .wp-block-file,
 .wp-block-button,
 .page-template > h2,
-.page-template > ul,
+.page-template > ul:not(.skip-links),
 .post-details > h2,
-.post-details > ul {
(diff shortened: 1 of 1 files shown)


assets/src/styles/blocks/core-overrides/BlockSpacing.scss
```diff
@@ -21,17 +21,10 @@
 .wp-block-file,
 .wp-block-button,
 .page-template > h2,
-.page-template > ul,
+.page-template > ul:not(.skip-links),
```
Inwerpsel, line 24: Be aware that adding a `:not()` selector increases the specificity of this rule, which can result in other rules in chlid themes not being applied anymore.

In general `:not()` is something to be used sparingly at it makes the code a lot harder to understand. So instead of changing the rule that applies to all ul's, can we fix it instead in the `.skip-links` class?

Inwerpsel: Looks like putting `display: block !important;` in `.skip-links` also fixes it.

comzeradd: We already have a place for `.skip-links` in [the styleguide](https://github.com/greenpeace/planet4-styleguide/blob/master/src/components/_skip-links.scss). Maybe it's a better place to override attributes.

Inwerpsel: Won't be needed after all, we found the issue. `.page-template` is not only on the `div` wrapping the post content, it is also on the `body` element of those pages. So `.skip-links` is a direct descendant of the body it also matched.

oekeur:
Check, implemented suggested changes (in styleguide and use display:block on the class itself)

New PR at: https://github.com/greenpeace/planet4-styleguide/pull/70

assets/src/styles/blocks/core-overrides/BlockSpacing.scss
Inwerpsel: So we can fix the issue by changing it to `div.page-template`

PR #330, opened 2020-07-22
Planet 4801 hooks

Inwerpsel:


assets/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -65,19 +66,26 @@ export class ArticlesBlock {
...
-          setAttributes={setAttributes}
-          isSelected={isSelected}
-          postId={postId}
-        />
-      }),
+          attributes={ attributes }
+          setAttributes={ setAttributes }
+          isSelected={ isSelected }
+          postType={ postType }
+          postId={ postId }
```
mleray, line 84: You pass down the `postId` here but I don't see it used anywhere in `ArticlesEditor` anymore...?

assets/src/blocks/Articles/ArticlesList.js
```diff
@@ -1,108 +1,17 @@
...
-            key={post.post_title}
-            isCampaign={postType === 'campaign'}
-            post={post}
-          />
-        )}
-      </div>
-    );
-  }
+export const ArticlesList = (props) => {
+  const { posts, postType } = props;
```
mleray, line 4: Instead of passing down `posts` as props, couldn't we use `useArticlesFetch` to get them here too?

assets/src/blocks/Articles/useArticlesFetch.js
```diff
@@ -0,0 +1,66 @@
...
+    setPosts([]);
+    loadPage(true);
+  }, [attributes]);
+
+  return {
+    posts,
+    setPosts,
+    totalPosts,
+    loading,
+    error,
```
mleray, line 62: I think this error is currently not used anywhere? 🤔

assets/src/blocks/Articles/ArticlesList.js
Inwerpsel: Hmm, short story: we could, but not sure if we should :)

long story: Initially I had set it up so that `useArticlesFetch` returned a function to reset the posts, for when the attributes change, intended only for the editor. That's why I couldn't call this function here. But then I removed that function as it's a lot easier to handle this inside `useArticlesFetch`.

So after that change it would in fact work the same way if we call `useArticlesFetch` here. However there is a high chance that we want to use some of the state in `ArticlesEditor`: for example when they change the number of articles shown, we actually know how many posts there are in total. So we could display that currently there are a total of `x` posts matched by the parameters of the block.

assets/src/blocks/Articles/useArticlesFetch.js
Inwerpsel: It's not, but we could display it in the editor.

Inwerpsel:
Closing in favor of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/332/files

PR #331, opened 2020-07-22
PLANET-5300 Add JS acceptance tests for Counter block

sagarsdeshmukh:
Ref. [JIRA 5300](https://jira.greenpeace.org/browse/PLANET-5300)

To test, run `npm install` and `npm test`.
Use `npm run test:watch` to run the tests in slow motion and see what happens in the browser.

image

### DIFF
--- a/tests/js/counter.spec.js
+++ b/tests/js/counter.spec.js
@@ -1,0 +1,131 @@
+import {
+  createNewPost,
+  enablePageDialogAccept,
+  getEditedPostContent,
+  insertBlock,
+} from '@wordpress/e2e-test-utils';
+
+import {
+  selectBlockByName,
+  openSidebarPanelWithTitle,
+  typeInInputWithLabel,
+  selectStyleByName,
+  typeInInputWithPlaceholderLabel,
+  typeInTextareaWithLabel,
+} from './e2e-tests-helpers';
+
+import 'expect-puppeteer';
+
+const COUNTER_API_URL = 'https://global-petition-counter.appspot.com/counter/palmoil';
+
+describe( 'Counter block', () => {
+  beforeAll( async () => {
+    // This helps in overriding Wordpress dialogs preventing actions
+    await enablePageDialogAccept();
+  } );
+  beforeEach( async () => {
+    // Before running each test, go to the create post page
+    await createNewPost();
+
+    // Insert block by title
+    await insertBlock( 'Counter' );
+  } );
+
+  // This is the first test, tests starts with it().
+  it( 'is inserted into the Editor', async () => {
+    // Check if block was inserted
+    expect( await page.$( '[data-type="planet4-blocks/counter"]' ) ).not.toBeNull();
+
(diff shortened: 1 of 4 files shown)


tests/js/counter.spec.js
```diff
@@ -0,0 +1,148 @@
...
+    await typeInInputWithPlaceholderLabel( 'Enter title', 'Text only counter block' );
+    await typeInInputWithPlaceholderLabel( 'Enter description', 'Counter block description' );
+
+    // Select the Text only counter style
+    await openSidebarPanelWithTitle( 'Styles' );
+    await selectStyleByName( 'Text Only' );
+
+    // Select the Text only counter style
+    await openSidebarPanelWithTitle( 'Setting' );
+    await typeInInputWithLabel( 'Completed', '75' );
```
mleray, line 69: Maybe we should integers instead of strings for the "Completed" and "Target" fields?

tests/js/counter.spec.js
```diff
@@ -0,0 +1,148 @@
...
+
+    // Select the Text only counter style
+    await openSidebarPanelWithTitle( 'Styles' );
+    await selectStyleByName( 'Progress Dial' );
+
+    await setUpResponseMocking( COUNTER_API_MOCK_RESPONSES );
+
+    // Add Setting inputs
+    await openSidebarPanelWithTitle( 'Setting' );
+    // TO DO: Need some changes in counter block to work below line of code.
```
mleray, line 137: What is the issue here?

sagarsdeshmukh: The api url is a long string and on each keypress the react preview got refresh in test execution, which turns into test timeout. I am thinking to use similar solution like [spreadsheet block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L92)

mleray: Is there a ticket for this issue?

sagarsdeshmukh: Nope I just added it ( [PLANET-5372](https://jira.greenpeace.org/browse/PLANET-5372) ) (I was tring to fix it my own but fall into many other errors)
Would you like to take it forward?

tests/js/counter.spec.js
sagarsdeshmukh: The `page.keyboard.type();` function expect the text input. If we pass number it throws - “TypeError: text is not iterable”

tests/js/counter.spec.js
sagarsdeshmukh: The issue was solved by [adding a catch](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/331/commits/67f3890b8c6af97c54a27e9bea2b323da17223b8) on fetch() function.

sagarsdeshmukh:
Hi @mleray - I updated the PR with code review fixes, please have a look. :)

mleray (approved):
Looks good to me 👍

PR #332, opened 2020-07-22
PLANET-4801 Make ArticlesEditor function compo

Inwerpsel:
* That way we can call both hooks inside of it and registerBlockType
is a lot simpler.


assets/src/blocks/Articles/useArticlesFetch.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  };
+
+  useEffect(() => {
+    setDisplayedPosts([]);
+    loadPage(true);
+  }, [ article_count, post_types, posts, tags, ignore_categories ]);
+
+  return {
+    posts: displayedPosts,
```
Inwerpsel, line 60: I kept the original `posts` name intentionally, as `displayedPosts` is a bit redundant outside the context of this function.

PR #336, opened 2020-07-23
PLANET-5296 Fix empty TA Card on Posts

sagarsdeshmukh:
Related to [JIRA #5296](https://jira.greenpeace.org/browse/PLANET-5296) 

Fix -
- Add  'post_status' = 'publish' condition on selected TAB page
- Return empty array if no TAB page is active
- Fix `PHP Notice:  Undefined variable: tags`

Test -
- Add a new TA page
- add/edit post and select newly added TA page from "Take Action Page Selector" dropdown at bottom of post or add a "Take Action Boxout" block from P4 blocks with new TA page
- Check the new post on frontend, the TAB is visible as intended
- Now change the status of TA page to draft
- The TAB should no longer visible on post page

e.g. https://k8s.p4.greenpeace.org/test-venus/story/28509/fukushima-and-the-2020-olympics/

### DIFF
--- a/classes/blocks/class-takeactionboxout.php
+++ b/classes/blocks/class-takeactionboxout.php
@@ -144,8 +144,9 @@ class TakeActionBoxout extends Base_Block {
 		}
 
 		$args = [
-			'p'         => intval( $page_id ), // ID of a page, post.
-			'post_type' => 'any',
+			'p'           => (int) $page_id, // ID of a page, post.
+			'post_type'   => 'any',
+			'post_status' => 'publish',
 		];
 
 		// Try to find the page that the user selected.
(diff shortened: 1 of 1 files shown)


classes/blocks/class-takeactionboxout.php
```diff
@@ -168,13 +169,15 @@ public function prepare_data( $fields ): array {
 					];
 				}
 			}
+		} else {
```
lithrel, line 172: Would you mind reverting the previous condition, so that we don't need an `else`?
```php
if ( ! $query->have_posts() ) {
    return [];
}

// If page is found populate the necessary fields for the block.
// ...
```

PR #337, opened 2020-07-24
PLANET-5265: Remove CarouselHeader block specific typography [skip ci]

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5265

- Removed font-size, font-weight, line-height for h1 to h3


lithrel:
[Ticket](https://jira.greenpeace.org/browse/PLANET-5265) was re-assigned.

PR #341, opened 2020-07-28
PLANET-4801 Render block node in backend

Inwerpsel:
* Otherwise older blocks that weren't saved with the block node in the
block content would not display until they are saved in the wywiwyg
version.


Inwerpsel:
> Maybe we could add a comment, to eventually remove this afterwards? It shouldn't be needed once all existing blocks are converted

:+1: also created a ticket for it https://jira.greenpeace.org/browse/PLANET-5364

PR #340, opened 2020-07-27
PLANET-5343: Plastics theme: items should be center aligned

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5343

> Task column block copy, icon and button copy should be centre aligned and not left aligned

This PR aligns icon and button of plastic campaign columns to the center.

Screenshot 2020-07-27 at 17 22 20

PR #342, opened 2020-07-28
PLANET-5343: Plastics theme: items should be center aligned

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5343
Depends on https://github.com/greenpeace/planet4-base-fork/pull/89 for tests
Replaces #340

> Task column block copy, icon and button copy should be centre aligned and not left aligned

This PR aligns icon and button of plastic campaign columns to the center.  
It also changes the columns_tasks markup to be closer to columns and use flexbox by removing most of the container divs and using `block-style-image` rules.

## Tests

- Export campaigns from the [staging site](https://release.k8s.p4.greenpeace.org/international/) (Tools > Export)
- Import the file locally (Tools > Import)
- Display the campaign _Lobby for Single-Use Ban in Your Community_
- Icons, title and buttons should be centered

Screenshot 2020-07-28 at 14 45 45


assets/src/styles/campaigns/themes/_theme_plastic.scss
```diff
@@ -361,6 +361,12 @@ body.theme-plastic {
           text-transform: none;
         }
       }
+
+      .steps-action {
+        .row > .col {
+          text-align: center;
```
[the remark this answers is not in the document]
Inwerpsel: Actually the `steps-action` wrapper is not there anymore now we use flexbox, so this selector is not doing anything. Probably flexbox is making it centered now.

templates/blocks/columns_tasks.twig
```diff
@@ -28,71 +28,42 @@
...
-										{% endif %}
-									</div>
-								{% endfor %}
+										<img src="{{ col.attachment }}" alt=""/>
+									{% if ( col.cta_link ) %}
+										</a>
+									{% endif %}
+								{% endif %}
+								{% if ( col.cta_text and col.cta_link ) %}
+									<a
```
Inwerpsel, line 55: We need to add some margin above this one now (or below the element above it). Previously this margin was ensure by the default `p` margin-bottom, but with flexbox that margin is not used.

Inwerpsel: Actually I saw this wrong. The margin is there, but in the child theme there is a rule to set the p bottom margin to 0. Then they had a rule using `.steps-action` to set a top margin to the CTA, which is now not applying anymore.

Inwerpsel: [From their commit message](https://github.com/greenpeace/planet4-child-theme-netherlands/commit/32c803d93c8b27b0fab457ed41cab72083f1b194#diff-3f958b82f02ca6ad0d9a567fa9ebc9c6R77) it looks like that rule was only added to provide alignment. So maybe a lot of their custom columns css can be removed if we fixed those things in base. @dirkjf can you confirm/correct my assumptions?

dirkjf: @Inwerpsel, yes, the rule was added simply for alignment.

Inwerpsel: @dirkjf So if we merge this PR we would need to ensure [this rule](https://github.com/greenpeace/planet4-child-theme-netherlands/blob/develop/assets/css/scss/blocks/_columns.scss#L116) is removed before we deploy it (maybe best to also double check if more things can be removed or rules need to be adapted to the new markup). We need to make sure both changes are in the same deploy. Would that work for you?

comzeradd: Let's unblock this. I can pause the GPNL deploy on next release and send a notification for this rule to be removed before approving production.

lithrel:
Removed `.post-content` declarations (block not used in posts anymore) and `.steps-action` declarations, ready for another review @Inwerpsel

Inwerpsel (approved):
Looks good :100:

PR #344, opened 2020-07-29
PLANET-5371 Translate defaults on the backend

Inwerpsel:
* For translations to work on the frontend we need to setup some things
first, so this workaround restores what was previously translated on the
backend until we have set up the frontend translations.

Ref: https://jira.greenpeace.org/browse/PLANET-5371
Deployed on https://k8s.p4.greenpeace.org/test-pandora/wp-admin/

### DIFF
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -43,7 +43,7 @@ class Articles extends Base_Block {
 				// todo: Remove when all content is migrated.
 				'render_callback' => static function ( $attributes ) {
 					if ( empty( $attributes['read_more_text'] ) ) {
-						$attributes['read_more_text'] = __('Load More', 'planet4-blocks');
+						$attributes['read_more_text'] = __( 'Load More', 'planet4-blocks' );
 					}
 
 					if ( empty( $attributes['article_heading'] ) ) {
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -42,6 +42,14 @@ class Articles extends Base_Block {
 				'editor_script'   => 'planet4-blocks',
 				// todo: Remove when all content is migrated.
 				'render_callback' => static function ( $attributes ) {
+					if ( empty( $attributes['read_more_text'] ) ) {
+						$attributes['read_more_text'] = __('Load More', 'planet4-blocks');
+					}
+
+					if ( empty( $attributes['article_heading'] ) ) {
+						$attributes['article_heading'] = __( 'Related Articles', 'planet4-blocks' );
+					}
+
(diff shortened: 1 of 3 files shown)


classes/blocks/class-articles.php
```diff
@@ -42,6 +42,14 @@ public function __construct() {
 				'editor_script'   => 'planet4-blocks',
 				// todo: Remove when all content is migrated.
 				'render_callback' => static function ( $attributes ) {
+					if ( empty( $attributes['read_more_text'] ) ) {
+						$attributes['read_more_text'] = __('Load More', 'planet4-blocks');
+					}
+
+					if ( empty( $attributes['article_heading'] ) ) {
+						$attributes['article_heading'] = __( 'Related Articles', 'planet4-blocks' );
```
mleray, line 50: If we're gonna use this instead of "Latest Articles", we need to update it in several other places:
- https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesFrontend.js#L23
- https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesBlock.js#L18
- https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesBlock.js#L63 (I'm not sure if this one matters, since it's deprecated 🤔)

sagarsdeshmukh: The [Default value](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/33ca30cfb8b7d08b1af6b59dee7e9e5175991e23/classes/blocks/class-articles.php#L143) before wysiwyg was 'Related Articles' , so in that case if we are explicitly not changing it to 'Latest Articles' then we need to update strings as @mleray highlighted.
Otherthan that PR looks good to me. We will shortly fix the JS string translation functionality .

Inwerpsel: @mleray I removed those other instances as they wouldn't be used anyway (attribute is now always set in backend), and that way it's also in one place.

Inwerpsel: But kept the one in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesBlock.js#L18, that is used in the editor.

mleray: Tested on local, works properly 👍

Inwerpsel: @mleray Good that you pointed out the deprecated attribute. Turns out that if this is different from the current default value of that attribute, then WP will put the old default as the attribute value, which we probably don't want to happen. So I removed the defaults from deprecated.

PR #346, opened 2020-07-30
PLANET-5263: Follow header typography rules for campaigns

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5263

> This screen shot looks like H1 is not resizing for Mobile. This is the non campaign rules (...) So either apply H1 or H2 rules to it. H1 looks the closest.
 
Removing `page-header-(title|subtitle)` typography declarations to follow our styleguide rules.  `.page-header-title` is always associated with `h1`, `.page-header-subtitle` with `h2`, so no additional rule is needed.  
An exception was made for theme Arctic, as the font is really big and would break layout if treated as `h1/h2`. It takes `h2/h3` rules.   
Title can take a larger space on large screens, so it doesn't look like a column with large font-size.

mob-comp

### DIFF
--- a/assets/src/styles/campaigns/themes/_theme_antarctic.scss
+++ b/assets/src/styles/campaigns/themes/_theme_antarctic.scss
@@ -62,14 +62,10 @@ body.theme-antarctic {
     @include page-header-block($sanctuary);
 
     .page-header-title {
-      font-size: 3.125rem;
-      line-height: 4.375rem;
       color: $antarctic-blue;
     }
 
     .page-header-subtitle {
-      font-size: $font-size-xl;
-      line-height: 2.5rem;
       color: $antarctic-dark-blue;
     }
 
--- a/assets/src/styles/campaigns/themes/_theme_arctic.scss
+++ b/assets/src/styles/campaigns/themes/_theme_arctic.scss
@@ -74,15 +74,23 @@ body.theme-arctic {
     @include page-header-block($save-the-arctic);
 
     .page-header-title {
-      font-size: 3.125rem;
-      line-height: 3.75rem;
       color: $arctic-blue;
+
+      @include mobile-only {
+        line-height: 1.38;
+        margin-bottom: 5px;
+        font-size: $font-size-xl;
+      }
     }
 
     .page-header-subtitle {
-      font-size: $font-size-xl;
-      line-height: 2.125rem;
       color: $arctic-blue;
+
+      @include mobile-only {
+        line-height: 1.38;
+        margin-bottom: 5px;
+        font-size: $font-size-lg;
+      }
     }
 
     .page-header-background:after {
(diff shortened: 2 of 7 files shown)


assets/src/styles/campaigns/themes/_theme_arctic.scss
```diff
@@ -73,16 +73,30 @@ body.theme-arctic {
...
+
     .page-header-title {
-      font-size: 3.125rem;
-      line-height: 3.75rem;
       color: $arctic-blue;
+
+      @include mobile-only {
+        line-height: 1.38;
+        margin-bottom: 5px;
+        font-size: $font-size-xl;
```
Inwerpsel, line 88: Maybe we can solve this too by introducing variables for the font size on each screen size? Then we can give it a default value in our base css, and only set this variable to something else for the arctic theme.

assets/src/styles/campaigns/themes/_theme_antarctic.scss
```diff
@@ -61,15 +61,17 @@ body.theme-antarctic {
   .page-header {
     @include page-header-block($sanctuary);
 
+    .container > *:not(.row) {
```
Inwerpsel, line 64: Why do you we need these in campaigns only?

lithrel: The change in font size is significant in x-large-and-up, and was breaking the previous layout of many titles I saw, so I tried to keep them more or less with the same number of break lines. But I understand if you prefer to drop specific rules, that also makes sense

assets/src/styles/campaigns/themes/_theme_arctic.scss
lithrel: As seen during sprint review, this will be done in a different PR

assets/src/styles/campaigns/themes/_theme_antarctic.scss
Inwerpsel: Makes sense to provide continuity for existing content :+1: Can you give an example of a page where it breaks the layout?

Inwerpsel: If we're adding the exact same rule to each campaign, we can add it in one place somewhere by using `.single-campaign` in the selector.

lithrel: Couldn't find those cases anymore (I should take more notes :sweat:), so I removed those changes; it will pop up at some point if something looks odd in some campaigns, and we can devise a proper fix at that time.

comzeradd:
Based on the scope of the ticket we can wrap this up? @Inwerpsel?

There is some activity on the "Campaign Themes" story, so we will re-work on this at some point to make it more consistent.

Inwerpsel (approved):
Looks good :+1:

PR #347, opened 2020-07-30
PLANET-4811 Fix Submenu rendering

mleray:
This should prevent breaking the existing Submenu blocks when merging the new WYSIWYG version.


assets/src/blocks/Submenu/SubmenuBlock.js
```diff
@@ -30,6 +29,10 @@ export class SubmenuBlock {
           type: 'string',
           default: ''
         },
+        submenu_style: { // Needed for old blocks conversion
```
Inwerpsel, line 32: We could also keep the deprecated entry if we would only add the field here to not break older blocks?

PR #352, opened 2020-08-04
Planet 4807 - Happypoint block WYSIWYG

mleray:
See https://jira.greenpeace.org/browse/PLANET-4807


assets/src/blocks/Articles/ArticlesEditor.js
```diff
@@ -153,7 +153,7 @@ const renderView = ({ attributes, postType, posts, totalPosts }, toAttribute) =>
 export const ArticlesEditor = (props) => {
   const { isSelected, attributes, setAttributes } = props;
 
-  const { postType, postId } = useSelect((select) => ({
+  const { postType, postId } = ((select) => ({
```
pablocubico, line 156: Heads up! this looks like an accident ☝️

mleray: Oops indeed! 😱

assets/src/blocks/Happypoint/HappypointFrontend.js
```diff
@@ -0,0 +1,108 @@
...
+    this.getData = this.getData.bind(this);
+  };
+
+  componentDidMount() {
+    this.getData();
+  }
+
+  componentDidUpdate(prevProps) {
+    const { id, iframe_url, mailing_list_iframe } = this.props;
+    if (
```
pablocubico, line 22: Not a blocker, but if we are going to use hooks more and more, this looks like it could be expressed better as an effect hook with dependencies ( ìd`, `iframe_url`, `mailing_list_iframe`). Maybe it wouldn't cost too much to turn this class into a functional component and let this comparisons happen automatically through memoization.

pablocubico:
Hey @mleray I noticed some issues with the focal point picker, I couldn't trace down the bug yet, I'm attaching some screenshots.

* When adding a brand new Happypoint block, the image does not show in the picker:
Captura de pantalla 2020-08-31 a la(s) 19 09 44

* After using the pencil icon on top to change the image, it still shows the previous one
Captura de pantalla 2020-08-31 a la(s) 19 09 01

* This also happens after removing the image using the trashcan icon:
Captura de pantalla 2020-08-31 a la(s) 19 09 21

assets/src/blocks/Happypoint/HappypointEditor.js
```diff
@@ -0,0 +1,178 @@
...
+  const dimensions = { width: 400, height: 100 };
+
+  const { imageUrl } = useSelect(select => {
+    let imageUrl = '';
+    if (id && id > 0) {
+      const imageDetails = select('core').getMedia(id);
+      imageUrl = (imageDetails && imageDetails.source_url) || '';
+    }
+    return { imageUrl };
+  }, []);
```
pablocubico, line 40: Maybe we should list `id`  as a dependency here, so it gets memoized and this hook is called again in case it changes, to make sure the `FocalPointPicker` using `imageURL` below is re-rendered (check my screenshots in the Conversation tab!).

pablocubico:
Hey @mleray I'm still having this issue, the focal point picker image is not getting updated, is that working for you?

happypoint-test-1

mleray:
> Hey @mleray I'm still having this issue, the focal point picker image is not getting updated, is that working for you?
> 
> happypoint-test-1

Hey @pablocubico, you don't seem to have the latest version of the code! The focal point is now moved to the sidebar 🙂

assets/src/blocks/Happypoint/HappypointFrontend.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  return (
+    <section className="block block-footer block-wide happy-point-block-wrap">
+      <picture>
+        <img {...imgProps} />
+      </picture>
+      <div className="container">
+        <div className="row justify-content-md-center">
+          {mailing_list_iframe && url &&
```
pablocubico, line 47: Looks great! One possibly dangerous issue: the old version renders the URL anyway if its set, despite the `mailing_list_iframe` being set or not, this is from the old Twig template:
 
`{% if ( fields.mailing_list_iframe == 'true' or fields.iframe_url ) %}`

That was wrong, the new version is correct according to the sidebar hint, but maybe we should:
- a) replicate the old behavior and create a new ticket to correct it for the next release
- b) flag the ticket to let editors know they have to ensure the `Use mailing list iframe` switch is ON...

Maybe @comzeradd can help deciding.

pablocubico:
@mleray you were right, I did some Docker experiments and is not syncing properly, tested and everything looks ok now, except for a single comment I added, please check!

assets/src/blocks/Happypoint/HappypointFrontend.js
mleray: Good catch @pablocubico! If we decide to keep this new behaviour we also probably need to update the documentation 🤔

comzeradd: @mleray Let's go with option **a**. Better replicate the previous behavior and open a bug about the toggle to refine it.

I'd personally remove the toggle completely. If the url field has a values it should be enough. But let's discuss this after this is done.

mleray: @comzeradd @pablocubico done! I'll open a ticket to investigate and improve the behaviour.

mleray: Ticket created: https://jira.greenpeace.org/browse/PLANET-5489

comzeradd (approved):
:100:

PR #354, opened 2020-08-05
PLANET-5301 Added JS acceptance tests for Articles block

sagarsdeshmukh:
Ref. [JIRA 5301](https://jira.greenpeace.org/browse/PLANET-5301)

image

To test, run `npm install` and `npm run test`.
Use `npm run test:watch` to run the tests in slow motion and watch execution in browser.

### DIFF
--- a/tests/js/articles.editor.spec.js
+++ b/tests/js/articles.editor.spec.js
@@ -1,0 +1,103 @@
+import {
+  createNewPost,
+  enablePageDialogAccept,
+  getEditedPostContent,
+  insertBlock,
+} from '@wordpress/e2e-test-utils';
+
+import {
+  selectBlockByName,
+  openSidebarPanelWithTitle,
+  typeInInputWithLabel,
+  typeInInputWithPlaceholderLabel,
+  clickElementByText,
+  typeInDropdownWithLabel,
+  clearPreviousTextWithLabel,
+  clearPreviousTextWithPlaceholder,
+} from './e2e-tests-helpers';
+
+import 'expect-puppeteer';
+
+const ARTICLES_INVALID_URL_WARNING = 'The URL must start with "https://"';
+
+describe( 'Articles block', () => {
+  beforeAll( async () => {
+    // This helps in overriding Wordpress dialogs preventing actions.
+    await enablePageDialogAccept();
+  } );
+  beforeEach( async () => {
+    // Before running each test, go to the create post page.
+    await createNewPost({
+      postType: 'page',
+      title: 'Test Articles block',
+    } );
+
+    // Insert block by title.
+    await insertBlock( 'Articles' );
+    await page.waitForSelector( '.articles-title-container' );
+  } , 40000);
+
+  // This is the first test, tests starts with it().
+  it ( 'is inserted into the Editor', async () => {
+    // Check if block was inserted
+    expect( await page.$( '[data-type="planet4-blocks/articles"]' ) ).not.toBeNull();
+
+    expect( await getEditedPostContent() ).toMatchSnapshot();
+  } );
+
+  it ( 'should work with all editor inputs', async () => {
+    await selectBlockByName( 'planet4-blocks/articles' );
+
+    // Add richtext inputs.
+    await clearPreviousTextWithPlaceholder( 'Enter title' );
+    await typeInInputWithPlaceholderLabel( 'Enter title', 'News' );
+    await typeInInputWithPlaceholderLabel( 'Enter description', 'Latest news details' );
+
+    // Add loadmore button text in wysiwyg editor.
+    await clearPreviousTextWithLabel( 'Button Text' );
+    await typeInInputWithPlaceholderLabel( 'Enter text', 'More News' );
+
+    // Add Setting inputs
+    await openSidebarPanelWithTitle( 'Setting' );
+    await clearPreviousTextWithLabel( 'Button Text' );
+    await typeInInputWithLabel( 'Button Text', 'Read more' );
+    await typeInInputWithLabel( 'Button Link', 'https://www.greenpeace.org' );
+    await clickElementByText( 'label', 'Open in a new Tab' );
+    await clearPreviousTextWithLabel( 'Articles count' );
+    await typeInInputWithLabel( 'Articles count', '4' );
+    await typeInDropdownWithLabel( 'Select Tags', 'Climate' );
+    await typeInDropdownWithLabel( 'Post Types', 'Story' );
+    await clickElementByText( 'label', 'Ignore categories' );
+
+    expect( await getEditedPostContent() ).toMatchSnapshot();
+  }, 50000 );
+
+  it ( 'should show a warning if the URL is wrong', async () => {
+    await typeInInputWithLabel( 'Button Link', 'this is not a URL' );
+
+    // The warning component should appear.
+    await expect( page ).toMatchElement( '.edit-post-sidebar .input_error' );
+
+    // The warning text should appear.
+    await expect( page ).toMatch( ARTICLES_INVALID_URL_WARNING );
+  } );
+
+  it.skip ( 'should use only the manually chosen posts if any are entered', async () => {
+    await openSidebarPanelWithTitle( 'Setting' );
+
+    await typeInDropdownWithLabel( 'Manual override', 'Duis posuere' );
+    await typeInDropdownWithLabel( 'Manual override', 'Mauris quis dictum magna' );
+    await typeInDropdownWithLabel( 'Manual override', 'Lorem ipsum dolor sit amet' );
+
+    // The only 3 manually override post should appear in articles block.
+    await expect(page).toMatchElement('article:nth-child(1) h4 a', { text: 'Duis posuere' });
+    await expect(page).toMatchElement('article:nth-child(2) h4 a', { text: 'Mauris quis dictum magna' });
+    await expect(page).toMatchElement('article:nth-child(3) h4 a', { text: 'Lorem ipsum dolor sit amet' });
+
+    // The "Load more" should not appear.
+    await expect( page ).not.toMatchElement( '.block-editor .article-load-more' );
+
+    expect( await getEditedPostContent() ).toMatchSnapshot();
+  } , 50000 );
+} );
+
--- a/tests/js/articles.frontend.spec.js
+++ b/tests/js/articles.frontend.spec.js
@@ -1,0 +1,139 @@
+import {
+  createNewPost,
+  enablePageDialogAccept,
+  setPostContent,
+  publishPost,
+} from '@wordpress/e2e-test-utils';
+
+import {
+  selectBlockByName,
+  openSidebarPanelWithTitle,
+  clickElementByText,
+  typeInDropdownWithLabel,
+} from './e2e-tests-helpers';
+
+import 'expect-puppeteer';
+
+const ARTICLES_BLOCK = `
+<!-- wp:planet4-blocks/articles {"article_count":2} -->
+<div class="wp-block-planet4-blocks-articles" data-render="planet4-blocks/articles" data-attributes="{&quot;attributes&quot;:{&quot;article_heading&quot;:&quot;Related Articles&quot;,&quot;article_count&quot;:2,&quot;tags&quot;:[],&quot;posts&quot;:[],&quot;post_types&quot;:[],&quot;read_more_text&quot;:&quot;Load More&quot;,&quot;read_more_link&quot;:&quot;&quot;,&quot;button_link_new_tab&quot;:false,&quot;ignore_categories&quot;:false},&quot;innerBlocks&quot;:[]}"></div>
+<!-- /wp:planet4-blocks/articles -->
+`;
+
+describe( 'Articles block frontend', () => {
+  beforeAll( async () => {
+    // This helps in overriding Wordpress dialogs preventing actions.
+    await enablePageDialogAccept();
+  } );
+
+  it ( 'test post thumbnail behaviour', async () => {
+    // Create new post.
+    await createNewPost({
+      title: 'Test Articles with no thumbnail',
+      content: 'Lorem Ipsum is simply dummy text of the printing and typesetting industry.',
+      showWelcomeGuide: false,
+    });
+    await publishPost();
+    await page.waitForSelector( '.post-publish-panel__postpublish' );
+
+    //Create a new page and add newly created post in it.
+    await createNewPost({
+      postType: 'page',
+      title: 'Test Articles block thumbnail',
+      showWelcomeGuide: false,
+    });
+
+    // Add articles block in page.
+    await setPostContent( ARTICLES_BLOCK );
+
+    await selectBlockByName( 'planet4-blocks/articles' );
+
+    await openSidebarPanelWithTitle( 'Setting' );
+
+    // Manually add post(without thumbnail img) in articles blocks.
+    await typeInDropdownWithLabel( 'Manual override', 'Test Articles with no thumbnail' );
+
+    // Manually add post(post with image inside post content, which was set as a feature image
+    // & display as a thumbnail in articles block) in articles blocks.
+    await typeInDropdownWithLabel( 'Manual override', 'Lorem ipsum dolor sit amet' );
+
+    // Publish the page.
+    await publishPost();
+
+    await new Promise((r) => setTimeout(r, 500));
+
+    await page.waitForSelector( '.post-publish-panel__postpublish' );
+
+    // Test articles block on frontend.
+    await clickElementByText( '*[@class="editor-post-publish-panel"]//a', 'View Page' );
+
+    await page.waitForNavigation();
+
+    await page.waitForSelector( '.page-template' );
+
+    // The "Articles Block" should appear on page.
+    await expect( page ).toMatchElement( '.block.articles-block' );
+
+    // If there is no image in the description of the post(also no featured image set),
+    // the blank space display as a thumbnail photo in articles block.
+    await expect(page).toMatchElement('article:nth-child(1) img', { src: ''} );
+    // If no Featured Image set, first image of post is set as a Featured Image &
+    // appear as a thumbnail in articles block.
+    const img_src = await page.$eval('article:nth-child(2) img', el => el.src);
+    expect( img_src ).not.toBeNull();
+  } , 50000 );
+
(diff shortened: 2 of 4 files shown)


tests/js/e2e-tests-helpers.js
```diff
@@ -73,3 +83,34 @@ export const openSidebarPanelWithTitle = async ( title ) => {
     await panel.click();
   }
 };
+
+/**
+ * Types in an inline input element(richtext) based on its placeholder label.
+ *
+ * @param {string} lable Placeholder text of the rich text input.
```
mleray, line 90: Tiny typo here 🙂 `lable` => `label`

tests/js/e2e-tests-helpers.js
```diff
@@ -73,3 +83,34 @@ export const openSidebarPanelWithTitle = async ( title ) => {
...
+  const [ element ] = await page.$x( `//*[contains(@class,"block-editor-rich-text__editable")][contains(@aria-label,"${ label }")]` );
+  await element.click();
+  await page.waitForSelector( ':focus.rich-text' );
+  await page.keyboard.type( value );
+};
+
+/**
+ * Types in an autocomplete input element based on its label.
+ *
+ * @param {string} lable Label text of the text input.
```
mleray, line 103: Same as above!

tests/js/articles.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    expect( await page.$( '[data-type="planet4-blocks/articles"]' ) ).not.toBeNull();
+
+    expect( await getEditedPostContent() ).toMatchSnapshot();
+  }, 50000 );
+
+  it( 'check all articles inputs', async () => {
+    await selectBlockByName( 'planet4-blocks/articles' );
+
+    // Add richtext inputs.
+    await typeInInputWithPlaceholderLabel( 'Enter description', 'Articles details' );
```
mleray, line 50: Looks like the title input is missing (the one with default "Related articles")? Maybe it would also be nice to test the "button text" RichText input?

tests/js/articles.spec.js
```diff
@@ -0,0 +1,136 @@
...
+      await expect( page ).toMatchElement( '.edit-post-sidebar .input_error' );
+
+      // The warning text should appear.
+      await expect( page ).toMatch( ARTICLES_INVALID_URL_WARNING );
+  } );
+
+
+  it ( 'test Manual override', async () => {
+    await openSidebarPanelWithTitle( 'Setting' );
+    //await jestPuppeteer.debug();
```
mleray, line 78: Can this be removed? 🙂

sagarsdeshmukh: yeap, my bad 🙈.

mleray:
Also, it looks like there are some linting issues that make the tests fail:
Screenshot 2020-08-06 at 13 40 55

sagarsdeshmukh:
Hi @mleray - I updated the PR with code review fixes, please have a look. :)

tests/js/articles.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+  // This is the first test, tests starts with it().
+  it ( 'is inserted into the Editor', async () => {
+    // Check if block was inserted
+    expect( await page.$( '[data-type="planet4-blocks/articles"]' ) ).not.toBeNull();
+
+    expect( await getEditedPostContent() ).toMatchSnapshot();
+  }, 50000 );
+
+  it ( 'check all articles inputs', async () => {
```
Inwerpsel, line 48: Some semantic nit-picking: We should probably make all of these strings into sentences that start with it. Otherwise it will be confusing/ambiguous to read.

Inwerpsel: Although in some cases (like here) it might get a bit contrived to find something fitting with "it".

Inwerpsel: Maybe
```
it ( ' Should put what you enter in the editor into the right attributes')
```

tests/js/articles.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    await typeInInputWithLabel( 'Button Link', 'this is not a URL' );
+
+    // The warning component should appear.
+    await expect( page ).toMatchElement( '.edit-post-sidebar .input_error' );
+
+    // The warning text should appear.
+    await expect( page ).toMatch( ARTICLES_INVALID_URL_WARNING );
+  } );
+
+  it ( 'test Manual override', async () => {
```
Inwerpsel, line 85: ``` it ( 'Should only show manually chosen posts if any are entered')``` ?

tests/js/articles.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    // Add Setting inputs
+    await openSidebarPanelWithTitle( 'Setting' );
+    await clearPreviousTextWithLabel( 'Articles count' );
+    await typeInInputWithLabel( 'Articles count', '4' );
+
+    // Publish the page.
+    await publishPost();
+
+    // Test articles block on frontend.
+    await clickElementByText( 'a', 'View Page' );
```
Inwerpsel, line 115: I think navigating away from the editor and to the frontend page this way will make the tests slow if we do that a lot. But we actually don't need to enter the articles count "manually" during this test. We already have a test which confirms that entering into the editor will put the value in the attribute. So in this test we could just have a snapshot with a fixed page (and fixed other content) where we could navigate directly to, and perform the assertions below this line.

Inwerpsel: Maybe it would make sense to put the frontend and editor tests into separate files? Then the editor test would correspond be for https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesEditor.js and the rest would be for https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesFrontend.js

sagarsdeshmukh: Yes, I agree with you. Created a separate [tests](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/354/commits/44fffdf3ad5fd3c6bb11bff4f2c1754787a7c544#diff-681f3eaecf153bfb54d00857cbae7c41) for frontend.

sagarsdeshmukh:
Hi @Inwerpsel - Thanks for the review, I updated the PR with fixes, could you please have a look. :)

Inwerpsel (commented):
A few more things we should check first, but almost there :+1:

tests/js/e2e-tests-helpers.js
```diff
@@ -130,6 +129,41 @@ export const typeInInputWithCheckbox = async ( name, value ) => {
...
+ * Types in an autocomplete input element based on its label.
+ *
+ * @param {string} label Label text of the text input.
+ * @param {string} value Value to be applied to the input.
+ */
+export const typeInDropdownWithLabel = async ( label, value ) => {
+  // Wait for 0.5 sec.
+  await new Promise((r) => setTimeout(r, 500));
+  const [ inputEl ] = await page.$x( `//label[contains(text(),"${ label }")]/following-sibling::div//input[@class="components-form-token-field__input"]` );
+  if ( inputEl ) {
```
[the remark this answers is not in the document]
sagarsdeshmukh: If due to some reason the inputEL not found on page the test fails with Snapshot not match error. without `if condition` test timeout at this point.

tests/js/e2e-tests-helpers.js
```diff
@@ -130,6 +129,41 @@ export const typeInInputWithCheckbox = async ( name, value ) => {
...
 
+/**
+ * Types in an autocomplete input element based on its label.
+ *
+ * @param {string} label Label text of the text input.
+ * @param {string} value Value to be applied to the input.
+ */
+export const typeInDropdownWithLabel = async ( label, value ) => {
+  // Wait for 0.5 sec.
+  await new Promise((r) => setTimeout(r, 500));
```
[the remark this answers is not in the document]
sagarsdeshmukh: This field is "Manual override" autocomplete text. without wait time the execution of this function is inconsistent, sometime works sometime won't.

tests/js/articles.spec.js
```diff
@@ -0,0 +1,103 @@
...
+const ARTICLES_INVALID_URL_WARNING = 'The URL must start with "https://"';
+
+describe( 'Articles block', () => {
+  beforeAll( async () => {
+    // This helps in overriding Wordpress dialogs preventing actions.
+    await enablePageDialogAccept();
+  } );
+  beforeEach( async () => {
+    // Before running each test, go to the create post page.
+    await createNewPost({
```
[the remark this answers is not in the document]
sagarsdeshmukh: Thats true, it could adjust with setTimeout() function till the time we get some other option.

tests/js/articles.spec.js
```diff
@@ -0,0 +1,103 @@
+import {
```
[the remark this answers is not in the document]
sagarsdeshmukh: I make it `articles.editor.spec.js` we need spec or test after test filename to understand jest its a test file.

sagarsdeshmukh:
> A few more things we should check first (mainly [#354 (comment)](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/354#discussion_r484875594)), but almost there 👍

PR updated, please have a look :)

tests/js/articles.frontend.spec.js
```diff
@@ -0,0 +1,139 @@
...
+
+    //Create a new page and add newly created post in it.
+    await createNewPost({
+      postType: 'page',
+      title: 'Test Articles block thumbnail',
+      showWelcomeGuide: false,
+    });
+
+    // Add articles block in page.
+    await setPostContent( ARTICLS_BLOCK );
```
Inwerpsel, line 47: Would adding the post content in `createNewPost` directly, on previous line, have the same result?

tests/js/articles.frontend.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    // Manually add post(post with image inside post content, which was set as a feature image
+    // & display as a thumbnail in articles block) in articles blocks.
+    await typeInDropdownWithLabel( 'Manual override', 'Lorem ipsum dolor sit amet' );
+
+    // Publish the page.
+    await publishPost();
+
+    await new Promise((r) => setTimeout(r, 500));
+
+    await page.waitForSelector( '.post-publish-panel__postpublish' );
```
Inwerpsel, line 65: The steps in [this part](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/354/files#diff-681f3eaecf153bfb54d00857cbae7c41R40-R65) need to be executed one by one in the editor. If these functions are already tested in the editor test, then we don't need to execute them here and can just insert the corresponding block comment into the post. That could make the test a lot faster.

Inwerpsel: You might need to get the post id for the post that is created [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/354/files#diff-681f3eaecf153bfb54d00857cbae7c41R40) in order to to that.

Inwerpsel (approved):
Looks good :100:  There's a few opportunities to make the test faster, though we could take them up later as an improvement if they're too time consuming to fix now.

tests/js/articles.frontend.spec.js
```diff
@@ -0,0 +1,139 @@
...
+import {
+  selectBlockByName,
+  openSidebarPanelWithTitle,
+  clickElementByText,
+  typeInDropdownWithLabel,
+} from './e2e-tests-helpers';
+
+import 'expect-puppeteer';
+
+const ARTICLS_BLOCK = `
```
Inwerpsel, line 17: Maybe we can make the snapshot into a template string where we call json encode? 
```
const attributes = {article_count: 2};
const ARTICLS_BLOCK = <!-- wp:planet4-blocks/articles ${JSON.stringify(attributes)} -->
<div class="wp-block-planet4-blocks-articles" data-render="planet4-blocks/articles" data-attributes="${JSON.stringify(attributes)}"></div>
<!-- /wp:planet4-blocks/articles -->
```

tests/js/articles.frontend.spec.js
sagarsdeshmukh: With `createNewPost` the blocks comments are treated as text data, I tried that earlier.

tests/js/articles.frontend.spec.js
sagarsdeshmukh: Thats my initial thought to get post_id of new post and add it directly in block comment and use it, but somehow I didn't find a way to get post_id that way... :(

tests/js/articles.editor.spec.js
```diff
@@ -0,0 +1,103 @@
...
+  } );
+
+  it ( 'should use only the manually chosen posts if any are entered', async () => {
+    await openSidebarPanelWithTitle( 'Setting' );
+
+    await typeInDropdownWithLabel( 'Manual override', 'Duis posuere' );
+    await typeInDropdownWithLabel( 'Manual override', 'Mauris quis dictum magna' );
+    await typeInDropdownWithLabel( 'Manual override', 'Lorem ipsum dolor sit amet' );
+
+    // The only 3 manually override post should appear in articles block.
```
mleray, line 92: There might be an issue here: now the articles in manual override are always returned ordered by date, regardless of the order that they are entered in the input (see [this ticket](https://jira.greenpeace.org/browse/PLANET-5462) for more info). So I think this part will now be failing, but I'm not sure how we can test this behaviour properly since we would need to know the dates of the articles... 🤔 Maybe we can just test that they are rendered, without checking the order?

tests/js/articles.frontend.spec.js
mleray, line 17: Small typo here:  `ARTICLS_BLOCK` => `ARTICLES_BLOCK` 🙂

tests/js/articles.frontend.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    await page.waitForNavigation();
+
+    await page.waitForSelector( '.page-template' );
+
+    // The "Articles Block" should appear on page.
+    await expect( page ).toMatchElement( '.block.articles-block' );
+
+    // If there is no image in the description of the post(also no featured image set),
+    // the blank space display as a thumbnail photo in articles block.
+    await expect(page).toMatchElement('article:nth-child(1) img', { src: ''} );
```
mleray, line 79: Here again, since these articles were added using the manual override, the order might now have changed due to [planet-5462](https://jira.greenpeace.org/browse/PLANET-5462) and this might fail 😕

tests/js/articles.frontend.spec.js
```diff
@@ -0,0 +1,139 @@
...
+    // If no Featured Image set, first image of post is set as a Featured Image &
+    // appear as a thumbnail in articles block.
+    const img_src = await page.$eval('article:nth-child(2) img', el => el.src);
+    expect( img_src ).not.toBeNull();
+  } , 50000 );
+
+  it ( 'should loads the right amount of posts when clicking the load more button', async ( done ) => {
```
mleray, line 86: Small typo: `loads` => `load`

tests/js/__snapshots__/articles.spec.js.snap
```diff
@@ -0,0 +1,19 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Articles block is inserted into the Editor 1`] = `
+"<!-- wp:planet4-blocks/articles -->
+<div class=\\"wp-block-planet4-blocks-articles\\" data-render=\\"planet4-blocks/articles\\" data-attributes=\\"{&quot;attributes&quot;:{&quot;article_heading&quot;:&quot;Related Articles&quot;,&quot;article_count&quot;:3,&quot;tags&quot;:[],&quot;posts&quot;:[],&quot;post_types&quot;:[],&quot;read_more_text&quot;:&quot;Load More&quot;,&quot;read_more_link&quot;:&quot;&quot;,&quot;button_link_new_tab&quot;:false,&quot;ignore_categories&quot;:false},&quot;innerBlocks&quot;:[]}\\"></div>
+<!-- /wp:planet4-blocks/articles -->"
+`;
+
+exports[`Articles block should uses only the manually chosen posts if any are entered 1`] = `
```
mleray, line 9: Small typo here: `uses` => `use`

mleray (changes requested):
I think some of the tests need to be updated following the changes induced by [planet-5462](https://jira.greenpeace.org/browse/PLANET-5462) (manual override articles are now returned ordered by date). Other than that, and a few typos here and then, looks good 🙂

sagarsdeshmukh:
> I think some of the tests need to be updated following the changes induced by [planet-5462](https://jira.greenpeace.org/browse/PLANET-5462) (manual override articles are now returned ordered by date). Other than that, and a few typos here and then, looks good 🙂

Can we add a separate ticket for the extended scope of manual override functional changes of test case?

comzeradd:
> > I think some of the tests need to be updated following the changes induced by [planet-5462](https://jira.greenpeace.org/browse/PLANET-5462) (manual override articles are now returned ordered by date). Other than that, and a few typos here and then, looks good slightly_smiling_face
> 
> Can we add a separate ticket for the extended scope of manual override functional changes of test case?

Since the scope of these tests changed a few times already, let's merge it (after we fix the typos) and work on a follow up ticket specifically for manual override.

tests/js/articles.editor.spec.js
```diff
@@ -0,0 +1,103 @@
...
+    await typeInInputWithLabel( 'Button Link', 'this is not a URL' );
+
+    // The warning component should appear.
+    await expect( page ).toMatchElement( '.edit-post-sidebar .input_error' );
+
+    // The warning text should appear.
+    await expect( page ).toMatch( ARTICLES_INVALID_URL_WARNING );
+  } );
+
+  it.skip ( 'should use only the manually chosen posts if any are entered', async () => {
```
sagarsdeshmukh, line 85: Skip the test from execution. Will fix this test in [PLANET-5493](https://jira.greenpeace.org/browse/PLANET-5493)

tests/js/articles.frontend.spec.js
sagarsdeshmukh: We create a new post "Test Articles with no thumbnail" which will display first and the existing post "Lorem ipsum dolor sit amet" on second position, (as per new manual override rule the newly publish post will display first) hence the test will pass here :)

mleray (approved):
Looks good to me! 👍

PR #328, opened 2020-07-16
PLANET-4811 Submenu WYSIWYG

mleray:
See https://jira.greenpeace.org/browse/PLANET-4811

- Turn the LayoutSelector into block styles
- Turn the Title field into a RichText component
- Move level options to the Sidebar

More details in the [doc](
https://docs.google.com/document/d/1RB4mOYWzUZM7uWYcyllfSdIGCQat4n04nrsFXLGIPBE/edit#heading=h.qo3z75gqnr39)

PR #356, opened 2020-08-06
Planet 4811 - Remove endpoint for Submenu block WYSIWYG

mleray:


assets/src/blocks/Submenu/submenuFunctions.js
```diff
@@ -1,3 +1,5 @@
+import md5 from 'md5';
```
Inwerpsel, line 1: We actually don't need to use md5 for this, not sure why it was done like that. Since we have access to all heading titles we could just use the title itself and suffix it with a number if we encounter it twice.

assets/src/blocks/Submenu/SubmenuFrontend.js
```diff
@@ -1,8 +1,6 @@
 import { Component, Fragment } from '@wordpress/element';
-import { getSubmenuStyle, addSubmenuActions } from './submenuFunctions';
+import { getSubmenuStyle, addSubmenuActions, loadMenuItems } from './submenuFunctions';
 const { __ } = wp.i18n;
-const { apiFetch } = wp;
-const { addQueryArgs } = wp.url;
 
 export class SubmenuFrontend extends Component {
```
Inwerpsel, line 5: This could easily be turned in a function component as the class doesn't have state. But we could do that in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/328/files too

Inwerpsel: Oh it does but the lines were hidden in the diff :facepalm:

assets/src/blocks/Submenu/submenuFunctions.js
```diff
@@ -48,32 +48,80 @@ export const addSubmenuActions = submenu => {
...
+        menuItems[menuItems.length - 1].children.push(menuItem);
+      } else {
+        menuItems.push(menuItem);
+      }
+    });
+  }
+  return menuItems;
+};
+
+const getTags = (headings, className = 'page-template') => {
```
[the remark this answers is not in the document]
mleray: We use this function here too: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/3dbc5d307a3d7f9fab6bb810730469ce80437a18/assets/src/blocks/Submenu/submenuFunctions.js#L68

assets/src/blocks/Submenu/submenuFunctions.js
sagarsdeshmukh: it was might be because of special char in header like `'` (apostrophe). right now if the header has apostrophe the submenu link wont work. (eg string "it's new to world" )

sagarsdeshmukh:
Functionality wise the new wysiwyg submenu block looks good to me 👍 
Found few observations -
- The transition to move focus from clicked submenu link to actual heading is different  from old submenu block (in old submenu block it was slow, In new one its fast)
- If heading title contains a special char's like apostrophe the submenu link won’t works
- The heading title got hide under top menu bar when scrolled to link

https://drive.google.com/file/d/1t_oVWunSj1pyCXS-YnND4lqjbAhoVHn6/view?usp=sharing

PR #357, opened 2020-08-07
PLANET-5292 Check EN feature toggle

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5292
Needs to be merged after https://github.com/greenpeace/planet4-master-theme/pull/1160 ! It uses classes introduced in that PR. The acceptance tests for this PR will fail until that is merged anyway.

Check the feature toggle of the EN block for the following cases:
* Whether the block can be used in the editor
* Whether the Engaging Networks admin menu item is displayed.
* Whether you can use the "Progress Bar inside EN Form" style on the counter block.

Because CMB2 interprets a falsy value on a boolean setting as "it should not store a record", it's not possible to set the default value to true, as that would make the field always true. So we need to run a script to turn the feature on everywhere instead. This is done in a post-deploy script (as part of the `p4-run-activator` command). There is a gap between when code is on production and when the post-deploy script runs. So to ensure the block is available before this script has run, we check whether the script has already run, which is stored in the db. That check can be removed in a subsequent deploy.

### DIFF
--- a/assets/src/blocks/Counter/CounterBlock.js
+++ b/assets/src/blocks/Counter/CounterBlock.js
@@ -81,7 +81,7 @@ export class CounterBlock {
 
       ]
 
-    if (window.p4ge_vars.planet4_options.feature_engaging_networks) {
+    if (window.p4ge_vars.features.feature_engaging_networks) {
       styles.push({
         name: 'en-forms-bar',
         label: __('Progress Bar inside EN Form', 'planet4-blocks')
--- a/assets/src/blocks/Counter/CounterBlock.js
+++ b/assets/src/blocks/Counter/CounterBlock.js
@@ -64,10 +64,7 @@ export class CounterBlock {
     // Remove the default style since it's the same as "text only"
     unregisterBlockStyle(BLOCK_NAME, 'default');
 
-    // Add our custom styles
-    registerBlockStyle(
-      BLOCK_NAME,
-      [
+    const styles = [
         {
           name: 'plain',
           label: __( 'Text Only', 'planet4-blocks' ),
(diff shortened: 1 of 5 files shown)


assets/src/blocks/Counter/CounterBlock.js
```diff
@@ -81,11 +78,16 @@ export class CounterBlock {
...
         },
-        {
-          name: 'en-forms-bar',
-          label: __( 'Progress Bar inside EN Form', 'planet4-blocks' )
-        }
+
       ]
-    );
+
+    if (!window.p4ge_vars.planet4_options.feature_engaging_networks) {
```
sagarsdeshmukh, line 84: This condition need to reverse. when EN feature is enabled then only EN form counter block style is available.

sagarsdeshmukh (commented):
PR looks good to me 👍  , added a comment about counter block EN form style.

classes/controller/menu/class-en-settings-controller.php
```diff
@@ -18,8 +22,12 @@ class En_Settings_Controller extends Controller {
 	 * Create menu/submenu entry.
 	 */
 	public function create_admin_menu() {
+		// We need to check if the migration already ran, as the EN block is on by default, but we cannot give an option
+		// that was added using CMB2 a default value of 'on', because then it can't be turned off.
+		$migration_ran     = MigrationLog::from_wp_options()->already_ran( M001EnableEnFormFeature::get_id() );
```
lithrel, line 27: I guess we can remove it once this migration passes ? let's keep a post-it or a ticket or something

lithrel (approved):
:+1:

PR #358, opened 2020-08-10
PLANET-4808: Convert Split Two Columns to WYSIWYG

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4808
Also solves:
- https://jira.greenpeace.org/browse/PLANET-4317
- https://jira.greenpeace.org/browse/PLANET-5205

## Conversion of Split Two Columns block to WYSIWYG.

Screenshot 2020-08-10 at 09 27 18

Just to be clear:
15-i-have-no-idea jpg=w768
So don't hesitate to make me redo everything.

## Details

The port to wysiwyg is pretty straightforward, `SplittwocolumnsFrontend.js` displays the frontend version, `SplittwocolumnsInPlaceEdit.js` is the same block in edit mode and `SplittwocolumnsSettings.js` has the sidebar settings.  
What's more complicated is the handling of old data. Technical choices for this are specified in the **Additions** section.    
Some state has to be saved, so that we know when to overwrite content with the selected issue/tag content, and when the user has edited something (in which case we keep the user data). The `edited` property keeps track of this.

## Tests

- Create, edit and view a new block Split Two Columns in a new page.
  - The only text not editable in-place is the Tag name
- View and edit existing blocks to verify they still work as intended.

## Other fixes

Fix for **PLANET-5205** Implement data attributes on Split Two Columns Block

Fix for **PLANET-4317** Split 2 column block: Text in button has a line break

### Addition of `.babelrc` file to allow for more js syntaxes
It makes the code using API responses more concise.

- nullish coalescing operator `??` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
- optional chaining operator `?.` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

### DIsplay of CMB2 fields in the REST API response

Introduced by https://github.com/greenpeace/planet4-master-theme/pull/1162, it was necessary to gather the information that was previously fetched by the PHP code (overwrite of title and description on the selected issue). CMB2 fields are not in the REST response by default.

### Display of `tag_attachment_id` in the REST API response

I needed the ID of the image associated to a tag, so it now appears when this block is registered. See `Splittwocolumns::__construct()`.

### New endpoint `POST planet4/v1/update_block/<block_name>`

Receives the current block structure, and returns the same block in the last structure version.

The reasoning is the following:  
- The block needs a new structure if we want to use React instead of Twig on the frontend. Twig was doing server-side rendering, which is cacheable and doesn't impact visitors as much as frontend side rendering.
- Either we convert existing blocks through a patch or on the fly before display, there has to be a PHP function that does this job, so it stays fast and compatible with the previous version of this block. Specifically, _images, descriptions, titles, links_, have to be fetched using only the issue, tag and images ids.  
- We could add a JS `migrate` function in the block code that does all this, but it would add many requests on the front side for visitors, and subsequently more complexity for the frontend code, so it has to be done on the backend.
- The new version adds these parameters on edition, they are available on the frontend without any more work. It doesn't need compatibility code.
- If we don't patch existing blocks, for better usability and less version juggling, it is easier to convert old blocks to the new structure during edition of those blocks; because the code for this is already ready for display compatibility in `SplitTwoColumns::update_data()`, we can also call this code through the API to transform the block, instead of writing this again in JS.

### Versioning of the structure

For future use, I considered that adding a version number to the block data was an easier way to handle changes that trying to compare properties existence/content.


assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
```diff
@@ -0,0 +1,104 @@
+import { Component, Fragment } from '@wordpress/element';
+
+const { apiFetch } = wp.apiFetch;
+
+export const SplittwocolumnsFrontend = (props) => {
+  const {
```
mleray, line 6: You could define these directly instead of `props` in the previous line, makes it a bit cleaner 🙂

assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
```diff
@@ -0,0 +1,104 @@
...
+  const analytics = (action) => {
+    return {
+      "data-ga-category": 'Split Two Columns',
+      "data-ga-action": action,
+      "data-ga-label": 'n/a'
+    }
+  }
+
+  return (
+    <Fragment>
```
mleray, line 35: I don't think you need to use `Fragment` here since there is only one child 🤔

assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
```diff
@@ -0,0 +1,104 @@
...
+               {...analytics('Tag Title')}>
+              #{tag_name}
+            </a>
+          }
+          {tag_description &&
+            <p className="split-two-column-item-subtitle"
+               dangerouslySetInnerHTML={{__html: tag_description}}
+            />
+          }
+          {button_text &&
```
mleray, line 92: Shouldn't we also check if `button_link` is set? Doesn't really make sense to have a link to nowhere 🤔

assets/src/blocks/Splittwocolumns/SplittwocolumnsInPlaceEdit.js
```diff
@@ -0,0 +1,111 @@
...
+export const SplittwocolumnsInPlaceEdit = ({attributes, charLimit, setAttributes}) => {
+  const toAttribute = attributeName => value => setAttributes({
+    [attributeName]: value
+  });
+  const debounceToAttribute = attributeName => debounce(
+    toAttribute(attributeName), 400
+  );
+
+  return (
+    <Fragment>
```
mleray, line 19: Same here, I don't think you need `Fragment` 🙂

assets/src/blocks/Splittwocolumns/SplittwocolumnsInPlaceEdit.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+        <div className="split-two-column-item-content">
+          <RichText
+            tagName="h2"
+            className="split-two-column-item-title"
+            placeholder={__('Enter Title', 'planet4-blocks-backend')}
+            value={attributes.title}
+            onChange={debounceToAttribute('title')}
+            characterLimit={charLimit.title}
+            multiline={false}
```
mleray, line 39: You need to use a string here instead of a boolean for this to work (weird, I know), so: `multiline="false"`... Same for other occurrences in the file 🙂

assets/src/blocks/Splittwocolumns/SplittwocolumnsInPlaceEdit.js
```diff
@@ -0,0 +1,111 @@
...
+  });
+  const debounceToAttribute = attributeName => debounce(
+    toAttribute(attributeName), 400
+  );
+
+  return (
+    <Fragment>
+    <section className="block-wide split-two-column">
+      <div className="split-two-column-item item--left">
+        {attributes.issue_image_src &&
```
mleray, line 22: Maybe you could destructure `attributes` at the top of the component instead of always repeating `attributes.`?

assets/src/blocks/Splittwocolumns/SplittwocolumnsSettings.js
```diff
@@ -0,0 +1,200 @@
...
+  const tagsList = useSelect((select) => {
+    const taxonomy_args = {hide_empty: false, per_page: 50};
+    return select('core').getEntityRecords('taxonomy', 'post_tag', taxonomy_args) || [];
+  }, []);
+
+  const onIssueChange = (issue_id) => {
+    issue_id = parseInt(issue_id);
+    const issue = issues.find(issue => issue.id === issue_id) || null;
+    setAttributes({
+      'select_issue': issue_id,
```
mleray, line 34: I don't think that you need to use the `' '` here (same for other occurrences)

assets/src/blocks/Splittwocolumns/SplittwocolumnsSettings.js
```diff
@@ -0,0 +1,200 @@
...
+      ),
+      'tag_link': tag?.link || '',
+      'button_text': attributes.button_text || __( 'Get Involved', 'planet4-blocks' ),
+      'button_link': attributes.button_link || tag?.link || ''
+    });
+  }
+
+  const onImageChange = (image_type, image) => {
+    setAttributes({
+      [image_type + '_id']: parseInt(image.id),
```
mleray, line 66: Here you could use avoid string concatenation 🙂 something like ```[`${image_type}_id`]``` should work!

assets/src/blocks/Splittwocolumns/SplittwocolumnsSettings.js
```diff
@@ -0,0 +1,200 @@
...
+      [image_type + '_src']: image.url,
+      [image_type + '_srcset']: '',
+      [image_type + '_title']: image.title
+    });
+  }
+
+  const onFocalChange = (focal_name, {x,y}) => {
+    setAttributes({[focal_name]: parseInt(x*100) + '% ' + parseInt(y*100) + '%' });
```
mleray, line 74: Same as above, you could avoid string concatenation 🙂: 
````${parseInt(x*100)}% ${parseInt(y*100)}%````

File: assets/src/blocks/Splittwocolumns/SplittwocolumnsSettings.js [hunk 10]
```diff
@@ -0,0 +1,200 @@
...
+  return (
+    <InspectorControls>
+      <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
+        <div>
+          {issue_page_list &&
+            <SelectControl
+              label={__('Select an issue', 'planet4-blocks-backend')}
+              value={attributes.select_issue}
+              options={issue_page_list}
+              onChange={(issue_id) => {onIssueChange(issue_id)}}
```
mleray, line 95: This can probably be simplified to just `{onIssueChange}`?

assets/src/blocks/Splittwocolumns/SplittwocolumnsInPlaceEdit.js
mleray: Same in SplittwocolumnsSettings.js

assets/src/blocks/Splittwocolumns/SplittwocolumnsSettings.js
```diff
@@ -0,0 +1,200 @@
...
+        </div>
+        }
+        <hr/>
+        <div>
+          {tag_list &&
+            <SelectControl
+              label={__('Select a tag', 'planet4-blocks-backend')}
+              value={attributes.select_tag}
+              options={tag_list}
+              onChange={(tag_id) => {onTagChange(tag_id)}}
```
mleray, line 139: Here I also think you can directly write `{onTagChange}`

assets/src/components/ImageOrButton/ImageOrButton.js
```diff
@@ -11,7 +11,7 @@ export class ImageOrButton extends Component {
     const {__} = wp.i18n;
 
     if ( typeof this.props.disabled == 'undefined' ) {
-      this.props.disabled = false;
+      // this.props.disabled = false;
```
mleray, line 14: Is this a mistake...? If not, this entire `if` statement can be removed 😄

classes/blocks/class-splittwocolumns.php
```diff
@@ -121,46 +93,69 @@ public function __construct() {
 	 * @return array The data to be passed in the View.
 	 */
 	public function prepare_data( $fields ): array {
```
mleray, line 95: The comment for this function needs to be updated 🙂

assets/src/components/ImageOrButton/ImageOrButton.js
lithrel: Hmm I think it was temporary, I had an error message saying that this property couldn't be set because it was readonly, I need to check that again

lithrel: Right, `can't define property "disabled": Object is not extensible`, gonna look for a better fix.

assets/src/blocks/Splittwocolumns/SplittwocolumnsEditor.js
```diff
@@ -0,0 +1,57 @@
...
+import { SplittwocolumnsSettings as SidebarSettings } from './SplittwocolumnsSettings';
+import { SplittwocolumnsInPlaceEdit as InPlaceEdit } from './SplittwocolumnsInPlaceEdit';
+
+const { apiFetch } = wp;
+const { addQueryArgs } = wp.url;
+
+export const SplittwocolumnsEditor = ({ attributes, setAttributes, isSelected }) => {
+  updateDeprecatedData(attributes, setAttributes, BLOCK_NAME, VERSION);
+  return (
+    <Fragment>
```
mleray, line 13: This `Fragment` can probably be removed too... Since `renderEdit` has one, and `renderView` is only one `section`

assets/src/blocks/Splittwocolumns/SplittwocolumnsInPlaceEdit.js
```diff
@@ -0,0 +1,126 @@
...
+          <div className="split-two-column-item-image">
+            <img
+              src={issue_image_src}
+              alt={issue_image_title || ''}
+              style={{objectPosition: focus_issue_image}}
+            />
+          </div>
+        }
+        <div className="split-two-column-item-content">
+          <RichText
```
mleray, line 48: In order to be consistent with the other WYSIWYG blocks (for example [ArticlesEditor](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesEditor.js#L115)), we should use `keepPlaceholderOnFocus={true}` in these RichText components (it also avoids weird behaviour on click)

mleray (commented):
There seems to be small bugs in the editor:
- Sometimes when switching issues (from Energy to People for example) the description doesn't update (I think because People's description is empty). Also, if switching back to `--Select issue--`, the title and description don't reset
- For tags the same issue happens when switching back to `--Select Tag--` (title and description don't reset), in addition the `Campaign button link` field in the sidebar also doesn't reset

lithrel:
> There seems to be small bugs in the editor:

Yes, I coded it like that, considering that any present text could have been modified by the user, I didn't want to risk losing it.  I'm gonna inquire about the expected behavior when switching issue/tag depending if text is present/empty/modified (and how to define those states), and document this.

assets/src/blocks/Splittwocolumns/SplittwocolumnsEditor.js
```diff
@@ -0,0 +1,75 @@
...
+import { SplittwocolumnsSettings as SidebarSettings } from './SplittwocolumnsSettings';
+import { SplittwocolumnsInPlaceEdit as InPlaceEdit } from './SplittwocolumnsInPlaceEdit';
+
+const { apiFetch } = wp;
+const { useSelect } = wp.data;
+
+
+const useImage = (image_id, current_url) => {
+  const { image = {url: current_url} } = useSelect(select => {
+    return image_id ? { image: select('core').getMedia(image_id) } : { image: null };
```
mleray, line 13: Maybe here you could write the check inside of the object, to avoid a bit of duplicate code? Sthg like `{ image: image_id ? select('core').getMedia(image_id)  : null }`

mleray (approved):
Looks good to me! Just a bit more outdated code that should be removed before merging:
- [this line](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/command/class-shortcode-to-gutenberg.php#L41) in `class-shortcode-to-gutenberg.php`
- [the twig template](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/templates/blocks/split_two_columns.twig)

PR #363, opened 2020-08-14
PLANET-4811 Submenu WYSIWYG

Inwerpsel:
* Since we only need to watch for changes in the editor, we don't need
to make the code that queries the DOM update the block when the DOM
changes. Instead we can use `useSelect` to get the blocks and filter the
headers (and potentially other blocks that can output a heading).
* Decouple fetching the headers from creating a hierarchical
representation.
* Just fetch the headings in render of SubmenuFrontend. This component
doesn't get re-rendered anyway, and also the headings won't change
(unless other blocks still need to be rendered, we probably should check
that).
* Split functions into dedicated files.
* Add backtop as an effect, which simply toggles the display of it. The
backtop markup and style will be added to the base template in
greenpeace/planet4-master-theme#1166
* Use plain old links instead of implementing our own behavior. Mainly
did this to remove (a lot of) complexity from the code while doing other
changes.  We should probably evaluate if we still want to do a transition,
though this can probably still make use of anchors.


assets/src/blocks/Submenu/SubmenuEditor.js
```diff
@@ -69,11 +71,64 @@ const renderEdit = (attributes, setAttributes) => {
...
+const renderView = (attributes, setAttributes, className) => {
+
+  const { blocks } = useSelect(select => {
+    return ({ blocks: select('core/editor').getBlocks() });
+  });
+
+  const { levels: selectedLevels } = attributes
+
+  const flatHeaders = [];
+  blocks.forEach(block => {
```
Inwerpsel, line 90: I still need to move some things from here to a separate function to keep `SubmenuEditor` a bit more high level.

assets/src/blocks/Submenu/getHeadingsFromDom.js
```diff
@@ -0,0 +1,26 @@
+import { generateAnchor } from './generateAnchor';
```
Inwerpsel, line 1: I didn't make this handle headers that are on the page more than once. We could do that by keeping track of the already generated anchors and add a number at the end.

Another option could be to concatenate the previous level heading anchors. That would be more future proof compared to appending a number. If a heading is added somewhere in between, then old links would be pointing to the wrong section if we're using numbers.

assets/src/blocks/Submenu/BackTop.js
```diff
@@ -0,0 +1,18 @@
+import useScrollPosition from '@react-hook/window-scroll';
+
+export const BackTop = ({}) => {
```
mleray, line 3: No need for the brackets here 🙂 `export const BackTop = () => {` should be fine

assets/src/blocks/Submenu/BackTop.js
```diff
@@ -0,0 +1,18 @@
...
+  const cookies = document.getElementById('set-cookie');
+
+  const cookiesVisible = cookies && cookies.style.display !== 'none';
```
mleray, line 6: This will not work... I was initially using `style.display` but found out that I had to use `getComputedStyle` instead... I think it's because the display property isn't set directly in the style object, but via css!

assets/src/blocks/Submenu/SubmenuFrontend.js
```diff
@@ -1,20 +1,20 @@
...
   const style = getSubmenuStyle(className, submenu_style);
 
   return (
-    <section className={`block submenu-block submenu-${style}`}>
-      <h2>{title}</h2>
-      <SubmenuItems menuItems={menuItems} />
+    <section className={ `block submenu-block submenu-${ style }` }>
+      <h2>{ title }</h2>
+      <SubmenuItems menuItems={ menuItems }/>
+      <BackTop/>
```
mleray, line 17: Having the `BackTop` element here (inside the section) causes z-index issues and makes the button disappear under the footer when you scroll all the way to the bottom... That's why previously I appended it to the body instead!

assets/src/blocks/Submenu/BackTop.js
```diff
@@ -0,0 +1,18 @@
...
+
+  const scrollPosition = useScrollPosition(60);
```
mleray, line 8: Isn't it possible to use `window.scrollY` here?

assets/src/blocks/Submenu/SubmenuFrontend.js
```diff
@@ -1,20 +1,20 @@
...
+  const headings = getHeadingsFromDom(levels);
+  const menuItems = makeHierarchical(headings);
   const style = getSubmenuStyle(className, submenu_style);
 
   return (
-    <section className={`block submenu-block submenu-${style}`}>
-      <h2>{title}</h2>
-      <SubmenuItems menuItems={menuItems} />
+    <section className={ `block submenu-block submenu-${ style }` }>
+      <h2>{ title }</h2>
```
mleray, line 15: We should probably hide this if there is no title? 🤔  Currently it displays as an empty tag

assets/src/blocks/Submenu/makeHierarchical.js
```diff
@@ -0,0 +1,33 @@
...
+      possibleParent = possibleParent.parent;
+    }
+
+    const parent = possibleParent;
+
+    const container = parent === menuItems ? menuItems : parent.children;
+
+    const menuItem = {
+      text: content,
+      style: style,
```
mleray, line 20: Can just write `style` here 🙂

assets/src/blocks/Submenu/makeHierarchical.js
```diff
@@ -0,0 +1,33 @@
...
+      children: [],
+      parent: parent,
```
mleray, line 22: Same as above, can just write `parent` here 🙂

assets/src/blocks/Submenu/BackTop.js
Inwerpsel: a yeah that's a leftover :+1:

assets/src/blocks/Submenu/BackTop.js
Inwerpsel: Then how will it update if the scroll position changes?

assets/src/blocks/Submenu/SubmenuFrontend.js
Inwerpsel: Yeah we should probably decouple the backtop from the component in some way, maybe we can make it a page level option? We could then turn on that option whenever the page contains a Submenu block

assets/src/blocks/Submenu/makeHierarchical.js
Inwerpsel: Yeah that's one of the parts I still need to clean up :+1:

assets/src/blocks/Submenu/SubmenuEditor.js
Inwerpsel: done :heavy_check_mark:

assets/src/blocks/Submenu/BackTop.js
Inwerpsel: Ended up removing the component in favor of putting it into the base template here: https://github.com/greenpeace/planet4-master-theme/pull/1166/files

assets/src/blocks/Submenu/SubmenuFrontend.js
Inwerpsel: For now I just included the back to top markup into master theme here https://github.com/greenpeace/planet4-master-theme/pull/1166/files . Still need to add something to toggle its visibility when you're at the top of the page. But already fixed the z-index issue.

assets/src/blocks/Submenu/SubmenuEditor.js
```diff
@@ -69,11 +115,24 @@ const renderEdit = (attributes, setAttributes) => {
   );
 }
 
-const renderView = (attributes, setAttributes) => {
+// We can put the other blocks that can have a header inside in here along with the attribute containing the heading text.
+// Then we can also filter those to include them in the menu.
+const blockTypesWithHeadings = [
```
Inwerpsel, line 120: I included this logic to include headers in our blocks, to achieve the same behavior as the previous block. But we could check existing content to see if there actually are pages that have such titles in the submenu. If there aren't we could consider dropping this part, as it could be confusing to the user. They don't see the heading level in the editor for these headings like they do in heading blocks.

If we keep it then we need to add the other blocks with headings here too.

sagarsdeshmukh:
As we don't have inner block html content to edit, we could safely disable the `Edit as HTML` option for submenu block.
image

To disable it we need to add `html:false` in block support options -
`
supports: {
	html: false,
},
`

assets/src/blocks/Submenu/SubmenuEditor.js
```diff
@@ -0,0 +1,123 @@
...
+const { RichText } = wp.blockEditor;
+const renderEdit = (attributes, setAttributes) => {
+  function addLevel() {
+    const [previousLastLevel] = attributes.levels.slice(-1);
+    const newLevel = previousLastLevel.heading + 1;
+    setAttributes({ levels: attributes.levels.concat({ heading: newLevel, link: false, style: 'none' }) });
+  }
+
+  function onHeadingChange(index, value) {
+    let levels = JSON.parse(JSON.stringify(attributes.levels));
```
pablocubico, line 21: Not a blocker: This should be safe given the content of `levels` but I personally prefer using LoDash's `deepClone` because it's safer for different types and faster, but also is more expressive when reading the code for someone that is not familiar with this technique.

assets/src/blocks/Submenu/getHeadingsFromBlocks.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    if (block.name === 'core/freeform') {
+      const parser = new DOMParser();
+      const selector = selectedLevels.map(({heading})=>`h${heading}`).join();
+      const doc = parser.parseFromString(block.attributes.content, 'text/html');
+
+      const classicHeadings = doc.querySelectorAll(selector);
+
+      headings.push(...[...classicHeadings].map(h => {
```
pablocubico, line 41: Is the leftmost spread really needed here? 🤔

assets/src/blocks/Submenu/SubmenuEditor.js
Inwerpsel: Yes LoDash `deepClone` is better for this. But I actually didn't touch `SubmenuLevel` yet, it's still a class. Maybe it can be written in a way that doesn't require the `index` lookup, and maybe also doesn't need a deep clone. Then we can avoid adding a dependency on LoDash here.

Inwerpsel:
Rebasing seems to have confused GitHub, it's showing commits in the wrong order here. However locally git log does have the right order :thinking:

assets/src/blocks/Submenu/getHeadingsFromBlocks.js
Inwerpsel: Yes, `classHeadings` is a `NodeList` which doesn't have map, so the inner one makes it an array, then the outer one spreads that array into multiple arguments of push, else it would push the array itself as a single element.

assets/src/blocks/Submenu/SubmenuEditor.js
Inwerpsel: I made a function that has the same output as the JSON switcheroo has, and is a few times faster than it. It's not as "feature" complete as Lodash, in the sense that it doesn't preserve non-built in classes, but so doesn't the JSON approach we used :shrug: There are also some exotic or deprecated JavaScript peculiarities that don't behave the same way as in Lodash, but we don't use them and we probably never should :smile:

Inwerpsel: Did a quick benchmark
```
/usr/bin/node /home/pieter/deepclone/bm.js
JSON bytes:  18
matches JSON approach
JSON-switcheroo   :  x 1,306,315 ops/sec ±1.03% (85 runs sampled)
lodash            :  x 1,400,846 ops/sec ±1.91% (87 runs sampled)
Custom check refs :  x 5,385,561 ops/sec ±3.58% (83 runs sampled)
Fastest is Custom check refs : 
Fastest is [object Object]
JSON bytes:  2824
matches JSON approach
JSON-switcheroo   :  x 41,676 ops/sec ±0.77% (88 runs sampled)
lodash            :  x 40,323 ops/sec ±1.27% (85 runs sampled)
Custom check refs :  x 91,897 ops/sec ±1.79% (90 runs sampled)
Fastest is Custom check refs : 
Fastest is [object Object]
JSON bytes:  7374
matches JSON approach
JSON-switcheroo   :  x 13,925 ops/sec ±1.13% (91 runs sampled)
lodash            :  x 17,017 ops/sec ±1.41% (83 runs sampled)
Custom check refs :  x 40,898 ops/sec ±3.23% (83 runs sampled)
Fastest is Custom check refs : 
Fastest is [object Object]
JSON bytes:  122662
matches JSON approach
JSON-switcheroo   :  x 762 ops/sec ±1.33% (85 runs sampled)
lodash            :  x 887 ops/sec ±1.30% (86 runs sampled)
Custom check refs :  x 1,873 ops/sec ±1.22% (86 runs sampled)
Fastest is Custom check refs : 
Fastest is [object Object]
JSON bytes:  642680
matches JSON approach
JSON-switcheroo   :  x 176 ops/sec ±2.45% (69 runs sampled)
lodash            :  x 295 ops/sec ±1.61% (82 runs sampled)
Custom check refs :  x 243 ops/sec ±1.09% (81 runs sampled)
Fastest is lodash            : 
Fastest is [object Object]

Process finished with exit code 0

```

Inwerpsel:
> Rebasing seems to have confused GitHub, it's showing commits in the wrong order here. However locally git log does have the right order

The last rebase has un-confused GitHub :tada:

assets/src/functions/deepClone.js
```diff
@@ -0,0 +1,59 @@
...
+  // the same value. Needed for Date, also makes Boolean objects work (even though you shouldn't use them).
+  const param = typeof valueOf === 'object' ? null : valueOf;
+  // Don't try to construct custom objects, use Object instead, which behaves the same as the JSON approach.
+  let constructor = global[value.constructor.name] || Object;
+  const newObject = new constructor(param);
+
+  ancestors.push(value);
+  clones.push(newObject);
+
+  Object.keys(value).forEach(k => newObject[k] = _deepClone(value[k], ancestors, clones));
```
Inwerpsel, line 54: I initially used 
```
return Object.assign({}, ...Object.keys(value).map(k => ({ [k]: deepClone(value[k]) })));
``` 
here, but it's actually 5 times slower than the above.

assets/src/blocks/Submenu/SubmenuItems.js
```diff
@@ -0,0 +1,32 @@
...
+
+  const renderMenuItems = (items) => {
+    return items.map(({ anchor, text, style, shouldLink, children }) => (
+      <li key={anchor} className={`list-style-${style || 'none'} ${shouldLink ? "list-link" : "list-heading"}`}>
+        {shouldLink ?
+          <a
+            className="icon-link submenu-link"
+            href={`#${anchor}`}
+          >
+            {text}
```
lithrel, line 11: Same problem here as the other blocks about RichText automagically escaping things
Screenshot from 2020-10-19 11-43-50
The only solution we have used so far is to use `dangerouslySetInnerHTML`, which doesn't sound super good, we might have to look for a better solution at some point

assets/src/blocks/Submenu/SubmenuBlock.js
```diff
@@ -1,158 +1,72 @@
...
+          default: ''
+        },
+        submenu_style: { // Needed for old blocks conversion
+          type: 'integer',
+          default: 0
         },
         levels: {
           type: 'array',
-          default: [ {heading: 0, link: false, style: 'none'}]
+          default: [{ heading: 0, link: false, style: 'none' }]
```
lithrel, line 38: When adding this block, heading is at 0 but the `Submenu item` selector in the side panel is on `Heading 2`, which means no header is added to the submenu until we actively select something in `Submenu item` even if it indicates a valid value; maybe we should make the default to heading 2, or add an empty option `Select heading level` for the `SubmenuLevel`

assets/src/blocks/Submenu/SubmenuEditor.js
```diff
@@ -0,0 +1,123 @@
...
+          {__('Remove level', 'planet4-blocks-backend')}
+        </Button>
+      </PanelBody>
+    </InspectorControls>
+  );
+}
+
+const renderView = (attributes, setAttributes, className) => {
+
+  const blocks = useSelect(select => select('core/editor').getBlocks(), null);
```
lithrel, line 88: `select('core/editor').getBlocks()` is deprecated (since https://github.com/WordPress/gutenberg/pull/15770, there's a message in console), we have to use `select('core/block-editor').getBlocks()`

assets/src/functions/deepClone.js
```diff
@@ -0,0 +1,59 @@
...
+  const type = typeof value;
+  if (type === 'function') {
+    return undefined;
+  }
+  if (type !== 'object') {
+    return value;
+  }
+  if (ancestors.includes(value)) {
+
+    return clones[ancestors.indexOf(value)];
```
Inwerpsel, line 26: I made this one return the already cloned reference, if there's a circular reference. But maybe it makes more sense to throw an exception here, as that would also happen when doing `JSON.stringify(objectWithCircularReference)`, and circular references should in practice not happen unless we made a mistake somewhere.

assets/src/blocks/Submenu/SubmenuItems.js
Inwerpsel: I instead unescaped what is returned as the block content, that way we can normally pass it to render without having to call `dangerouslySetInnerHTML` https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/363/files#diff-24dc7e1c4c410e70f8e7e29e8f70b65ade34c0800fbcc8877034cb3d00fea05eR28

lithrel (approved):
Good to me :+1:

Minor thing I saw while looking at the Privacy & Cookies page, headings can be transformed to bold/italic (:shrug:) and the tags appear in the editor (doesn't seem to be a problem on the frontend so not blocking and probably not that frequent).
Screenshot from 2020-10-23 14-47-41

Inwerpsel:
> Good to me 
> 
> Minor thing I saw while looking at the Privacy & Cookies page, headings can be transformed to bold/italic () and the tags appear in the editor (doesn't seem to be a problem on the frontend so not blocking and probably not that frequent).
> Screenshot from 2020-10-23 14-47-41

Good catch :ok_hand: Checking if I can improve that, maybe just stripping tags will do here .

Noticed another issue we probably want to fix first: the previews of the styles in the sidebar all look the same, because the block's output is based on the current headings of the post, but these don't seem to get used in the preview. I'll see if I can make it work with the current headings of the post. If not we'll have to rework the block so that it allows to be rendered from a list of static headings as well, so we can just include those literally in the example data of the block.

assets/src/blocks/Gallery/GalleryBlock.js
```diff
@@ -63,23 +63,23 @@ export class GalleryBlock {
         {
           name: 'slider',
           label: getStyleLabel(
-            'Slider',
-            'The slider is a carousel of images. For more than 5 images, consider using a grid.'
+            __('Slider', 'planet4-blocks-backend'),
```
Inwerpsel, line 66: I noticed this issue for the Submenu block, so I made these changes here as well already.

assets/src/blocks/Submenu/SubmenuItems.js
```diff
@@ -0,0 +1,32 @@
+export const SubmenuItems = ({ menuItems }) => {
+
+  const renderMenuItems = (items) => {
+    return items.map(({ anchor, text, style, shouldLink, children }) => (
+      <li key={anchor} className={`list-style-${style || 'none'} ${shouldLink ? "list-link" : "list-heading"}`}>
```
lithrel, line 5: Some of the examples don't have an anchor (I think ?), so react raises an error 
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `SubmenuItems`. See https://fb.me/react-warning-keys for more information.
    in li (created by SubmenuItems)
```
Maybe if there's no anchor we can use the array index

Inwerpsel: Good catch :ok_hand:

Inwerpsel: I used the title as an anchor in the example attributes

PR #364, opened 2020-08-18
Planet 5325 Take Action Boxout A/B test

mleray:
## Description
See https://jira.greenpeace.org/browse/PLANET-5325 and latest designs in https://jira.greenpeace.org/browse/PLANET-5108

### How to test
If you want to test the RTL changes, just add `dir="rtl"` in the inspector to the `html` element!

**Bottom variant**
- Add `action-card-bottom` to the classes of the `action-card` element
- In `post_action.js` write the following in an `else` statement:
```
const element = document.querySelector('.action-card-bottom');
const post = document.querySelector('.post-content');
if (post) {
   post.appendChild(element);
   element.style.display = 'block';
}
```

**Scroll variant**
- Add `action-card-scroll` to the classes of the `action-card` element
- In `post_action.js` write the following in an `else` statement:
```
window.addEventListener('scroll', action_card_scroll);
window.addEventListener('resize', action_card_scroll);
```

Related master-theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1223

### DIFF
--- a/public/js/post_action.js
+++ b/public/js/post_action.js
@@ -3,7 +3,7 @@
 jQuery(function ($) {
   'use strict';
 
-  const $sidebar = $('.post-content').find('> #action-card');
+  const $sidebar = $('.post-content').find('> #action-card').not('.bottom');
   const offset = $sidebar.offset();
   const topPadding = 100;
 
@@ -29,5 +29,12 @@ jQuery(function ($) {
   if ($sidebar.length > 0) {
     window.addEventListener('scroll', scroll_action_card);
     window.addEventListener('resize', scroll_action_card);
+  }
+
+  // If "bottom" style, put take action boxout at the end of the post
+  const $bottom = $('.post-content').find('> #action-card.bottom');
+  if ($bottom.length) {
+    $bottom.appendTo('div.post-content');
+    $bottom.css('display', 'flex');
   }
 });
--- a/public/js/post_action.js
+++ b/public/js/post_action.js
@@ -36,7 +36,9 @@ jQuery(function ($) {
       if ($(window).scrollTop() > postStart && $(window).scrollTop() < (postHeight - postStart)) {
         // We want to align the boxout with the share buttons
         if ($(window).width() >= 992 && shareButtonsLeft && $boxoutScrollNotHidden.offset().left !== shareButtonsLeft) {
-          $boxoutScrollNotHidden.css('left', shareButtonsLeft + 'px');
+          $boxoutScrollNotHidden.css('left', `${shareButtonsLeft}px`);
+        } else if ($(window).width() < 992 && $boxoutScrollNotHidden.offset().left !== 0) {
+          $boxoutScrollNotHidden.css('left', 0);
         }
         $boxoutScrollNotHidden.fadeIn();
         // If mobile/tablet the user can close the boxout by clicking on the cross
--- a/public/js/post_action.js
+++ b/public/js/post_action.js
@@ -3,56 +3,63 @@
 jQuery(function ($) {
   'use strict';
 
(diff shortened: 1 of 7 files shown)


public/js/post_action.js
```diff
@@ -26,8 +26,33 @@ jQuery(function ($) {
     }
   }
 
+  function scroll_action_card_scroll() {
+    const $boxoutScroll = $('.post-content').find('> #action-card-scroll').not('.hidden');
+    const postStart = $('.post-content').position().top;
+    const postHeight = $('.post-content').outerHeight();
+    if ($boxoutScroll.length > 0) {
+      if ($(window).scrollTop() > postStart && $(window).scrollTop() < (postHeight - postStart)) {
```
lithrel, line 34: On a small screen (around 500px) with a short post of just a paragraph (more or less the size of `postStart`), the card never shows up. I don't know how it should behave in this condition.

lithrel (approved):
👍 
Just a remark about the behavior on short posts, I don't think it's blocking but I don't know the proportion of short posts we have on the instances that will serve as test, it may impact the test.

public/js/post_action.js
mleray: Yes that's a good point, already noticed by Will and he wants to discuss it with Magali!

public/js/post_action.js
```diff
@@ -26,8 +26,38 @@ jQuery(function ($) {
     }
   }
 
+  function scroll_action_card_scroll() {
+    const $boxoutScroll = $('.post-content').find('> #action-card-scroll').not('.hidden');
```
lithrel, line 30: The take action boxout block can also be added on pages, which don't have `.post-content` or `.share-buttons`, so this function a lot of console errors. Do we have a condition somewhere to add the changes/event listeners in posts only ?

mleray: Added a check to execute these functions only when there's `.post-content` 🙂

lithrel (approved):
:+1:

lithrel (approved):
Still good to me :+1: 

I have a little suggestion concerning this calculation `if ($(window).scrollTop() > postStart && $(window).scrollTop() < (postHeight - postStart)) {` to make the boxout (dis)appear; on my test with a 3 paragraphs post (windowWith: 1361, postStart: 338.6, postHeight: 883), the box seems to disappear a bit too early on scroll.
So instead of making it disappear as soon as the scrollTop passes a threshold I'm not sure I understand, we could make it disappear when the scrollTop reaches an equivalent of the boxout touching the end of the post ? like 
`if ($(window).scrollTop() > postStart && $(window).scrollTop() < (postHeight + postStart - boxoutHeight)) {`
with `const boxoutHeight = $('#action-card').innerHeight();`

templates/blocks/take_action_boxout.twig
```diff
@@ -1,63 +1,67 @@
...
-                    >
-                        {{ boxout.title|e('wp_kses_post')|raw }}
-                    </a>
-                {% endif %}
-            </div>
-			{% if ( boxout.excerpt ) %}
-				<p>{{ boxout.excerpt|e('wp_kses_post')|excerpt(25)|raw }}</p>
+{% if ( boxout ) %}
+  {% if ( boxout.image ) %}
+    {% set bg_image = 'style="background-image: linear-gradient(180deg, rgba(0, 0, 0, 0.3), rgba(0, 0, 0, 0)), url(\''~boxout.image~'\');"' %}
```
comzeradd, line 5: I think you have mixed indentation here. Both spaces and tabs. Twig files are using tabs.
If you are on VScode there is a "Convert indentation to tabs" option command

comzeradd (approved):
Looks good 💯

PR #366, opened 2020-08-20
PLANET-5227 added role=text to counter text for accessibility

dpivo:
Added role=test to the counter text to stop screen reader splitting up the text from the spans


sagarsdeshmukh (approved):
👍

PR #367, opened 2020-08-24
PLANET-5384 Use native JS instead of jQuery

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5384
* jQuery caused issues locally for new campaigns. Also nothing in the
script couldn't be done with native JS.
* Some readability improvements.

### DIFF
--- a/assets/src/BlockEditorValidation.js
+++ b/assets/src/BlockEditorValidation.js
@@ -8,9 +8,16 @@ export const blockEditorValidation = () => {
 };
 
 const isValid = element => {
-  // Apply validation only for campaign post types.
-  if ('campaign' === $('#post_type').val() && 'required' === $(element).data('validation')) {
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh:
Hi @Inwerpsel - The changes looks good to me, but on my local, the validation applied to post and pages as well. could you please confirm once?

Inwerpsel:
> Hi @Inwerpsel - The changes looks good to me, but on my local, the validation applied to post and pages as well. could you please confirm once?

Good catch, I added just `return` but it needs to be `return true`. Fixed now https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/367/files#diff-c391b7e90009d8476d32127fbf7c1ad8R15

sagarsdeshmukh (approved):
LGTM 👍

PR #368, opened 2020-08-26
PLANET-4813: Convert Timeline Block to WYSIWYG

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4813


assets/src/blocks/Timeline/TimelineBlock.js
```diff
@@ -1,126 +1,56 @@
...
-          setAttributes({timenav_position: value});
-        }
-
-        function onStartAtEndChange(value) {
-          setAttributes({start_at_end: value});
-        }
-
-        return <Timeline
-          {...attributes}
+        return <TimelineEditor
```
mleray, line 39: You could directly write `edit: TimelineEditor` since the props are exactly the same 🙂

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,196 @@
...
+];
+
+const position = [
+	{label: 'Bottom', value: 'bottom'},
+	{label: 'Top', value: 'top'},
+]
+
+let url_desc  = __(
+	'Enter the URL of the Google Sheets spreadsheet containing your timeline data.',
+	'p4ge'
```
mleray, line 91: I think we need to replace these `p4ge` domains by `planet4-blocks-backend`?

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,196 @@
...
+        withoutInteractiveFormatting
+        characterLimit={200}
+      />
+      <TimelineFrontend {...attributes} />
+		</Fragment>
+	)
+}
+
+export const TimelineEditor = (props) => {
+  const { isSelected, attributes, setAttributes } = props;
```
mleray, line 184: You could destructure these directly in the declaration to avoid an extra line 🙂 aka:
```
export const TimelineEditor = ({ isSelected, attributes, setAttributes }) => {
```

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,196 @@
...
+
+  const toAttribute = attributeName => value => setAttributes({ [attributeName]: value });
+
+  return (
+    <div>
+      {
+        isSelected && renderEdit(attributes, toAttribute)
```
mleray, line 191: It's a bit strange that these 2 functions (`renderEdit` & `renderView`) don't have the same signature, one has `attributes` passed down in an object and the other one has it passed directly... Not a big deal ofc, just a bit odd 😄

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,196 @@
...
+);
+url_desc += '<br>';
+url_desc += __(
+	"Once you have done so, use the URL from your address bar (not the one provided in Google's 'publish to web' dialog).",
+	'p4ge'
+);
+
+const renderEdit = (attributes, toAttribute) => {
+	return (
+    <Fragment>
```
mleray, line 111: I don't think that `Fragment` is needed here, since it has only one child? 🤔

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,196 @@
...
+      <InspectorControls>
+        <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
+					<URLInput
+						label={__('Google Sheets URL', 'p4ge')}
+						placeholder={__('Enter URL', 'p4ge')}
+						help={ <RawHTML>{url_desc}</RawHTML> }
+						value={attributes.google_sheets_url}
```
mleray, line 118: You could destructure `attributes` at the top of the function, or directly in the declaration, rather than repeat `attributes.` all the time 🙂

mleray (changes requested):
I've tested this locally and I've noticed a couple of issues:
- Already existing blocks are not rendered at first, they need to be manually saved in the editor (we probably need to add a render callback as we do [for the Articles block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-articles.php#L44) for example)
- Title and description are not rendered in the frontend (I think it has to do with the `isSelected ?` [selector](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/368/files#diff-50bb3df5e605dc35f10a93edfb4acd51R57))
Other than that, it looks good 🙂

pablocubico:
Hey @mleray thanks for the comments, looks much cleaner now!

I changed the component tree a little to make sure it renders in the frontend, decoupling the timeline node and removing the `isSelected` condition. Please check again!

assets/src/blocks/Timeline/TimelineFrontend.js
```diff
@@ -0,0 +1,29 @@
...
+			{
+				timeline_title
+					? <header>
+							<h2 className="page-section-header">{ timeline_title }</h2>
+						</header>
+					: null
+			}
+			{
+				description
+					? <div className="page-section-description">{ description }</div>
```
mleray, line 23: This should use `dangerouslySetInnerHTML` in order to support multilines 🙂

mleray (changes requested):

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,200 @@
...
+
+      </PanelBody>
+    </InspectorControls>
+	)
+}
+
+const renderView = (attributes, toAttribute) => {
+  return (
+    <section className="block timeline-block">
+      <Tooltip text={__('Edit text', 'planet4-blocks-backend')}>
```
mleray, line 157: Same as for Media block, this tooltip should not be necessary since there is no default value 🙂

mleray (approved):
One small comment but other than that, looks good to me 👍

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,206 @@
...
+  {label: 'Ukrainian', value: 'uk'},
+  {label: 'Urdu', value: 'ur'},
+];
+
+const position = [
+	{label: 'Bottom', value: 'bottom'},
+	{label: 'Top', value: 'top'},
+]
+
+let url_desc  = __(
```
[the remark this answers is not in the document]
Inwerpsel: Maybe would be nicer too if it's in a separate file too

Inwerpsel (changes requested):
Looks good so far, left some comments about immutability and code style.

Definitely we should check the `useScript` and `useStylesheet` functions. The current retry mechanism is probably not what we want to happen in case of a 404, as it would keep sending requests on every render.

assets/src/blocks/Timeline/useStyleSheet.js
```diff
@@ -0,0 +1,69 @@
+// useScript implementation from: https://usehooks.com/useScript/
+import { useEffect, useState } from 'react';
+
+// Hook
+let cachedStyleSheets = [];
+export const useStyleSheet = (href, media = 'all') => {
+
+  // Keeping track of script loaded and error state
+  const [state, setState] = useState({
```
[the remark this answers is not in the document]
Inwerpsel: Another reason (which I'm not sure React mentions) is that with multiple `useState` calls, it is much easier to move a part of the state to a higher component, as you can move the entire call and don't need to disassemble the arguments and `useState` calls.

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,206 @@
...
+  {label: 'Latvian', value: 'lv'},
+  {label: 'Lithuanian', value: 'lt'},
+  {label: 'Luxembourgish', value: 'lb'},
+  {label: 'Malay', value: 'ms'},
+  {label: 'Myanmar', value: 'my'},
+  {label: 'Nepali', value: 'ne'},
+  {label: 'Norwegian', value: 'no'},
+  {label: 'Polish', value: 'pl'},
+  {label: 'Portuguese', value: 'pt'},
+  {label: 'pt-br', value: 'Portuguese (Brazilian)'},
```
Inwerpsel, line 60: This one did the switcheroo :open_mouth:

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,206 @@
...
+              debounceSheetURLUpdate(value)
+            }
+          }
+      />
+
+        <SelectControl
+          label={__('Language', 'planet4-blocks-backend')}
+          value={language}
+          options={languages}
+          onChange={toAttribute('language')}
```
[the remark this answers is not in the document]
pablocubico: Hmmm you're right, it was like this before WYSIWYG but I guess it doesn't make much sense. As this is feature change, I would split it to another ticket just in case we are missing something, like a post actually having a Timeline in a language that is not the post's language..., but indeed seems unlikely.

assets/src/blocks/Timeline/useStyleSheet.js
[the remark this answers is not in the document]
pablocubico: Yes, I read that, but as the bold text says in that page: "we recommend to split state into multiple state variables based on which values tend to change together", and as these two (`loaded` and `error`) are always changing together, I think there's no need to split them here. It's just two, a very simple state, and they change together in all the useState calls.

And as the last paragraph in that question says: "Both putting all state in a single useState call, and having a useState call per each field can work. Components tend to be most readable when you find a balance between these two extremes, and group related state into a few independent state variables. If the state logic becomes complex, we recommend managing it with a reducer or a custom Hook.", and this is actually a custom Hook! I'm not planning to extract anything outside of it so far.

assets/src/blocks/Timeline/useStyleSheet.js
```diff
@@ -0,0 +1,69 @@
...
+        // Stylesheet event listener callbacks for load and error
+        const onStyleSheetLoad = () => {
+          setState({
+            loaded: true,
+            error: false
+          });
+        };
+
+        const onStyleSheetError = () => {
+          // Remove from cachedStyleSheets we can try loading again
```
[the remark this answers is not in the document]
pablocubico: Hmmm not sure I follow, what if it was a temporary network error?

assets/src/blocks/Timeline/TimelineNode.js
```diff
@@ -0,0 +1,52 @@
+import { useScript } from './useScript';
+import { useStyleSheet } from './useStyleSheet';
+import { useRef, useEffect } from 'react';
+import { uniqueId } from 'lodash';
+
+const TIMELINE_JS_VERSION = '3.6.3';
+
+export const TimelineNode = (props) => {
```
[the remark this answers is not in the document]
pablocubico: My logic was: it renders the node (`<div>`) in which the timeline is injected. Maybe "Element" would be more specific to a div, as a Text node can also be a node. But a div is a Node and an Element. Personally I always think of them as "nodes" (of a tree...).

assets/src/blocks/Timeline/TimelineFrontend.js
```diff
@@ -0,0 +1,29 @@
...
+		description,
+		isSelected,
+		...nodeProps
+	} = props;
+
+	return (
+		<section className="block timeline-block">
+			{
+				timeline_title
+					? <header>
```
Inwerpsel, line 16: We could also write this as ```{ timeline_title && <header>...</header> }```

assets/src/blocks/Timeline/TimelineNode.js
```diff
@@ -0,0 +1,52 @@
...
+		google_sheets_url,
+		timenav_position,
+		start_at_end,
+		language
+	} = props;
+
+	const timelineNode = useRef(null);
+
+	useStyleSheet(
+    `https://cdn.knightlab.com/libs/timeline3/${TIMELINE_JS_VERSION}/css/timeline.css`
```
Inwerpsel, line 19: I assume that in a lot of cases the timeline will be above the fold. Only starting the requests for the stylesheet and js files at this point will definitely cause the block to start displaying too late and couldcause a layout jump. Compared to including the sheet in the html doc, it should definitely be noticeable.

We could still include this stylesheet and the js with the backend if a post has the block in it.

assets/src/blocks/Timeline/TimelineNode.js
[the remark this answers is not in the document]
Inwerpsel: Ah right, didn't make that connection. The component does perform the injection itself, so we could still consider this component as rendering the timeline itself, even though it doesn't fully use React for it. So maybe we can consider it as "internals" and just call the component `Timeline`?

assets/src/blocks/Timeline/TimelineTranslations.js
```diff
@@ -0,0 +1,22 @@
+const { __ } = wp.i18n;
+
+export const URLDescriptionHelp =
```
Inwerpsel, line 3: If this only contains this long help message, then we could just name the file after it instead. Translations is a bit misleading as it doesn't contain actual translations :smile:

assets/src/blocks/Timeline/useStyleSheet.js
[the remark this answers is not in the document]
Inwerpsel: Yeah there's definitely pros and cons to both approaches. I guess it doesn't matter as long as a single component doesn't use a combination of both.

assets/src/blocks/Timeline/useStyleSheet.js
```diff
@@ -0,0 +1,71 @@
...
+      // Add stylesheet to document body
+      document.body.appendChild(linkElement);
+
+      // Remove event listeners on cleanup
+      return () => {
+        linkElement.removeEventListener('load', onStyleSheetLoad);
+        linkElement.removeEventListener('error', onStyleSheetError);
+      };
+    },
+    [href] // Only re-run effect if stylesheet href changes
```
Inwerpsel, line 67: When will this href change?

Inwerpsel (commented):
We can remove the backend rendering but still keep the stylesheet and js inclusion there. That way we don't need to have the component wait for those requests when it starts rendering.

assets/src/blocks/Timeline/TimelineNode.js
Inwerpsel: After checking around looks like mostly the block is used well below the fold. So maybe loading it in here only is preferable after all. It would also delay the rendering of the timeline, which can be pretty intensive.

assets/src/blocks/Timeline/useStyleSheet.js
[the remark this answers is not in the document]
Inwerpsel: It can be a temporary error. But it can also happen that it's not temporary (or temporary but takes like 1 hour). Then it would keep sending requests on every render.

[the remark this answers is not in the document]
Inwerpsel: Which is probably not a problem for the timeline block, but could be for blocks that render more frequently.

assets/src/blocks/Timeline/useStyleSheet.js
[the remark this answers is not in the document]
Inwerpsel: They do say "and group related state into a few independent state variables"...

In my previous comment I rather meant: if we use a hook with `[state, setState] = useState({a:b,c:d})`, and then need to add a second group of state variables that we want to keep separate, you would need to rename the first group to something more specific or it would get confusing.

assets/src/blocks/Timeline/TimelineBlock.js
```diff
@@ -1,126 +1,58 @@
...
+    type: 'string',
+    default: ''
+  },
+  start_at_end: {
+    type: 'boolean',
+    default: false
+  },
+};
 
 export class TimelineBlock {
```
Inwerpsel, line 34: Maybe we should start changing this too while we doing away with classes in favor of hooks. These classes are just a shell around the `registerBlockType` function, where we never use an instance of the class and just call the constructor for a side effect in it. Just a function which has whatever happens in the constructor seems a lot more straightforward.

assets/src/blocks/Timeline/TimelineFrontend.js
Inwerpsel: That way it's consistent with this line https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/368/files?diff=split&w=1#diff-70066a6561381bd819c059627c0e34faR117

assets/src/blocks/Timeline/TimelineNode.js
pablocubico: Yes, so I did ran some tests, which were not super conclusive as each time you ran them you get different scores for performance..., but what is consistent is that if we add the files as we do now (enqueue from PHP), Lighthouse complains about them blocking the main thread:

image

And the timeline gets listed under "blocking" resources here:
image

In this version (load on render using hooks), those issues disappear and you can see the timeline being loaded after the "first contentful paint", "largest contentful pain", and "DOM Content loaded" events:

image

In the process I found out the author's (KnightLab) CDN is not gzipping the files, so I'll see if these exists in another CDN with compression:

image
 
As for the jumpyness, this component has a fixed height (luckily), so we can make it render something to fill up the space and look better.

assets/src/blocks/Timeline/URLDescriptionHelp.js
```diff
@@ -0,0 +1,22 @@
+const { __ } = wp.i18n;
+
+export const URLDescriptionHelp =
+	__(
+		'Enter the URL of the Google Sheets spreadsheet containing your timeline data.',
+		'planet4-blocks-backend'
+	)
+	+ '<br><a href="https://timeline.knightlab.com/#make" target="_blank" rel="noopener noreferrer">';
```
lithrel, line 8: the `;` doesn't allow for the full message to appear :grin:

assets/src/blocks/Timeline/TimelineEditor.js
```diff
@@ -0,0 +1,124 @@
...
+  const [ sheetURL, setSheetURL ] = useState(google_sheets_url);
+  const debounceSheetURLUpdate = useCallback(debounce(toAttribute('google_sheets_url'), 300), []);
+
+  return (
+    <InspectorControls>
+      <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
+        <URLInput
+          label={__('Google Sheets URL', 'planet4-blocks-backend')}
+          placeholder={__('Enter URL', 'planet4-blocks-backend')}
+          help={ <RawHTML>{URLDescriptionHelp}</RawHTML> }
```
lithrel, line 35: React makes a `div` for `RawHTML` (which shouldn't be here according to [the doc](https://github.com/WordPress/gutenberg/blob/375eaa0d870282e12cd461a53463d3e10e776c41/packages/element/src/raw-html.js#L21) ?) and then complains that a `div` is in a `p` `Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.` :man_shrugging:

assets/src/blocks/Timeline/TimelineBlock.js
pablocubico: Indeed, initially we thought these would be classes but in the end it was never the case, I'll make this change here and create a separate PR to change all the other ones and remove those pointless `new BlockName()...` instantiations because I know you despise them! 🔥

assets/src/blocks/Timeline/URLDescriptionHelp.js
pablocubico: Good catch! 👁️  🔍

assets/src/blocks/Timeline/TimelineNode.js
Inwerpsel: Good catch about the compression :100: Probably makes most sense if we put it on our cloudflare, then that's one less domain to connect to.

comzeradd: cdnjs (that we already use for other libs), seems to have it

https://cdnjs.com/libraries/timelinejs

assets/src/blocks/Timeline/URLDescriptionHelp.js
```diff
@@ -0,0 +1,22 @@
+const { __ } = wp.i18n;
+
+export const URLDescriptionHelp =
```
Inwerpsel, line 3: Like mentioned in my other comment, this will probably be easier to handle if it's a component and not a string. Something we can change later too if we can live with the div-in-a-p-tag warning.

assets/src/components/useScript/useScript.js
```diff
@@ -0,0 +1,69 @@
+// useScript implementation from: https://usehooks.com/useScript/
+import { useEffect, useState } from 'react';
+
+// Hook
+const loadedScripts = [];
```
Inwerpsel, line 5: Suggestion here but we can take it up later too.

We could probably have this check the DOM if the script is already present/loaded to get this info, instead of keeping track of it here. Otherwise if the same script would already be present it would add another script tag with the same source and load it again, potentially causing issues by evaluating it twice.

Inwerpsel (approved):
Looks good :-1: We could look into the `useScript` and `useStylesheet` functions a bit further if we intend to use them more (how it keeps track of loaded resources and behavior on error), but I guess it will do for the usage introduced in this PR. Left some other suggestions but nothing blocking.

assets/src/components/useScript/useScript.js
pablocubico: I removed the caching array and now it checks for `!!document.querySelector(`script[src="${src}"]`)`. 👍

assets/src/blocks/Timeline/URLDescriptionHelp.js
pablocubico: @Inwerpsel @lithrel I created an `HTMLSidebarHelp` component and added the help outside the `help` prop, to avoid the warning.

PR #370, opened 2020-08-27
PLANET-5403 Carousel won't let edit the html code

dpivo:
changed to use frontendRendered and added deprecated  attribute

Fixes greenpeace/planet4#79

### DIFF
--- a/assets/src/blocks/Carouselheader/CarouselHeaderBlock.js
+++ b/assets/src/blocks/Carouselheader/CarouselHeaderBlock.js
@@ -1,10 +1,33 @@
 import {CarouselHeader} from './CarouselHeader.js';
+import { frontendRendered } from '../frontendRendered';
+
+const BLOCK_NAME = 'planet4-blocks/carousel-header';
 
 export class CarouselHeaderBlock {
   constructor() {
     const {registerBlockType} = wp.blocks;
+    const attributes = {
+      carousel_autoplay: {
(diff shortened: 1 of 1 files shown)


assets/src/blocks/Carouselheader/CarouselHeaderBlock.js
```diff
@@ -256,9 +268,7 @@ export class CarouselHeaderBlock {
           removeSlide={removeSlide}
         />;
       },
-      save() {
-        return null;
-      }
+      save: frontendRendered( BLOCK_NAME )
```
pablocubico, line 271: Hey @dpivo this is preventing the Carousel from rendering in the frontend, as it wasn't converted to WYSIWYG yet. 

Maybe the ticket needs refinement, blocks returning `null` are not expected to have any HTML representation in the Editor. I think the goal is to fix this:

https://jira.greenpeace.org/browse/PLANET-5463

But the "Edit HTML" option only makes sense for purely static blocks. I think we need to review what is expected and what is plausible. 

cc @comzeradd

pablocubico (changes requested):

assets/src/blocks/Carouselheader/CarouselHeaderBlock.js
comzeradd: We could wait till we transform this block too into WYSIWYG

PR #373, opened 2020-08-31
Planet 4800 - Gallery block WYSIWYG

mleray:
See https://jira.greenpeace.org/browse/PLANET-4800


assets/src/blocks/Gallery/useGalleryImages.js
```diff
@@ -0,0 +1,39 @@
...
+const { apiFetch } = wp;
+const { addQueryArgs } = wp.url;
+
+const GALLERY_IMAGE_SIZES = {
+  'slider': 'retina-large',
+  'three-columns': 'medium_large',
+  'grid': 'large'
+};
+
+export const useGalleryImages = ({ multiple_image, gallery_block_focus_points }, layout) => {
```
lithrel, line 12: Still reviewing, but I think you can drop the new endpoint and just use `getMediaItems()`, I've made a poc there https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/compare/planet-4800...planet-4800-woendpoint

Inwerpsel: I'm also for avoiding creating a new endpoint if that's possible. However in this case we probably can't use `getMediaItems` on the frontend. Locally I could only get that to work if I add the `wp-editor` script as a dependency of `frontendIndex.js`, and then only when logged in :grimacing:

lithrel: Oh, makes sense then. So that's something I'd like to be more clear in our next phase; in the split two columns block I saved every image src statically (and text etc.) so that there's no need for more requests for the visitor to get the block, I understand now that it's not resilient to domain change or image infos modifications; but there could be a middle ground where those data are prepared on the backend ? No additional request/endpoint and still resilient to change ?

Inwerpsel (commented):
Just leaving a few comments and some comment responses which I hadn't intended to add to the review :)

mleray:
Opened [another PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/381) for this issue, to use React hooks for the GalleryCarousel instead of the jquery library. Still in progress but worth noting 🙂

assets/src/blocks/Gallery/useGalleryImages.js
pablocubico: @lithrel @Inwerpsel Just ran into this old conversation, I migrated the Media block to a static-render version, but had to add a temporary endpoint to support existing content without running a migration. Static means resilient to change, but we need to discuss if that's what we actually want, maybe with a "pros & cons" table, considering all these cases. Maybe an old image caption should stay as it was in some old post even if it changes in the Media Library... not sure. Also, a static HTML block is easy to edit in the "Edit HTML" tab of the editor... but, do we want that? There's currently an open ticket about it. Having a listener to update all posts that use certain image seems overkill... but I'm not sure. Not sure about anything. Life is strange.

comzeradd:
@mleray do you think this can have a visual way of reordering the images? Maybe something like the [native gallery](https://k8s.p4.greenpeace.org/gutenberg/wp-admin/post.php?post=25693&action=edit). Although this can be challenging with our different styles.

assets/src/blocks/Gallery/GalleryBlock.js
```diff
@@ -1,207 +1,91 @@
...
-                  type: 'string',
-                  shortcode: ({named: {gallery_block_focus_points = ''}}) => gallery_block_focus_points,
-                },
-              },
-            },
-          ]
+    registerBlockType(BLOCK_NAME, {
+      title: __('Gallery', 'planet4-blocks-backend'),
+      icon: 'format-gallery',
+      category: 'planet4-blocks',
```
comzeradd, line 46: Can you also disable the "HTML edit" button?

Example: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/398/files

mleray:
> @mleray do you think this can have a visual way of reordering the images? Maybe something like the [native gallery](https://k8s.p4.greenpeace.org/gutenberg/wp-admin/post.php?post=25693&action=edit). Although this can be challenging with our different styles.

There is already a way to do this! After adding photos to the gallery, you can then choose to "Edit gallery" and there you have drag and drop possibilities:
Screenshot 2020-09-28 at 11 24 45

assets/src/blocks/Gallery/useGalleryImages.js
mleray: @lithrel @Inwerpsel I'm not sure what the conclusion is after this conversation, are there any changes needed in the end? Or is this good for now?

lithrel: Good for me at least :) I think we can review our gutenberg wysiwg methods after the first full conversion is done, and come back to it with more certainties and experience in a following phase :ok_hand:

assets/src/blocks/Gallery/GalleryBlock.js
```diff
@@ -1,207 +1,96 @@
...
-            type: 'string',
-          },
-          image_data: {
-            type: 'array',
-            default: []
-          },
+        {
+          name: 'three-columns',
+          label: getStyleLabel(
+            '3 Column',
```
lithrel, line 74: Column**s** ? (same next line)

assets/src/blocks/Gallery/GalleryFrontend.js
```diff
@@ -0,0 +1,36 @@
...
+  const layout = getGalleryLayout(className, gallery_block_style);
+  const postType = document.body.getAttribute('data-post-type');
+
+  const { images } = useGalleryImages({ multiple_image, gallery_block_focus_points }, layout);
+
+  return (
+    <section className={`block ${GALLERY_BLOCK_CLASSES[layout]}`}>
+      {gallery_block_title &&
+        <header>
+          <h2 className="page-section-header">{gallery_block_title}</h2>
```
lithrel, line 24: You have the same problem here we had with split 2 columns in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/401 , I don't have any specific fix in mind (I also need to check how the RichText does its own escape at some point, there's an [`escapeAmpersand`](https://github.com/WordPress/gutenberg/blob/4857ad58c1241b3d63d21a6880c989b85746c3dc/packages/escape-html/src/index.js) over there), either RichText, dangerouslySetInnerHTML or other :shrug:

assets/src/blocks/Gallery/GalleryBlock.js
mleray: I've kept the old copy here but I think you're right indeed, there should be an "s" 😄

assets/src/blocks/Gallery/GalleryFrontend.js
mleray: I'm confused, I thought that if the component used `RichText` in the editor (which it does here) then it's not a problem? 🤔

classes/blocks/class-gallery.php
```diff
@@ -144,7 +130,7 @@ public function prepare_data( $fields ): array {
 		];
 
 		foreach ( $exploded_images as $image_id ) {
-			$image_size = $image_sizes[ $fields['gallery_block_style'] ];
+			$image_size = $fields['gallery_image_size'] ? $fields['gallery_image_size'] : $image_sizes[ $fields['gallery_block_style'] ];
```
lithrel, line 133: If you want you can [shorten this ternary](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary) to `$image_size = $fields['gallery_image_size'] ?: $image_sizes[ $fields['gallery_block_style'] ];`

lithrel (approved):
:+1:

PR #375, opened 2020-09-01
PLANET-5479 Add page showing posts with "classic blocks"

Inwerpsel:
Since this block can be hard to support, it would be nice to know where it's still being used. (EDIT: I said "still" but actually new ones can be created using the code editor, which probably happens.)

We can't add this to the block report with the other core block types, as a classic block is not a real block as it's not inside a block comment, but is created on the fly when wordpress finds html that is not inside a block in a post.

https://k8s.p4.greenpeace.org/test-deimos/wp-admin/admin.php?page=classic_block_usages

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -346,6 +346,7 @@ P4GBKS\Loader::get_instance(
 		// 'P4GBKS\Controllers\Blocks\NewCovers_Controller'
 		\P4GBKS\Controllers\Menu\Settings_Controller::class,
 		\P4GBKS\Controllers\Menu\Blocks_Usage_Controller::class,
+		\P4GBKS\Controllers\Menu\Classic_Blocks_Usage::class,
 		\P4GBKS\Controllers\Menu\Reusable_Blocks_Controller::class,
 		\P4GBKS\Controllers\Menu\Archive_Import::class,
 		\P4GBKS\Controllers\Menu\Postmeta_Check_Controller::class,
--- a/classes/controller/menu/class-classic-blocks-usage.php
+++ b/classes/controller/menu/class-classic-blocks-usage.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Blocks Usage class
+ *
+ * @package P4BKS\Controllers\Menu
+ * @since 1.40.0
+ */
+
+namespace P4GBKS\Controllers\Menu;
+
+/**
+ * Show posts using classic blocks.
+ */
+class Classic_Blocks_Usage extends Controller {
+	/**
+	 * Create menu/submenu entry.
+	 */
+	public function create_admin_menu() {
+
+		$current_user = wp_get_current_user();
+
+		if ( in_array( 'administrator', $current_user->roles, true ) && current_user_can( 'manage_options' ) ) {
+			add_submenu_page(
+				P4GBKS_PLUGIN_SLUG_NAME,
+				__( 'Classic block usage', 'planet4-blocks-backend' ),
+				__( 'Classic block usage', 'planet4-blocks-backend' ),
+				'manage_options',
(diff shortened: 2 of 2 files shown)


planet4-gutenberg-blocks.php
```diff
@@ -346,6 +346,7 @@ function ( $std_class_object ) {
 		// 'P4GBKS\Controllers\Blocks\NewCovers_Controller'
 		\P4GBKS\Controllers\Menu\Settings_Controller::class,
 		\P4GBKS\Controllers\Menu\Blocks_Usage_Controller::class,
+		\P4GBKS\Controllers\Menu\Classic_Blocks_usages::class,
```
lithrel, line 349: **U**sages (uppercase U)

classes/controller/menu/class-classic-blocks-usages.php
```diff
@@ -0,0 +1,79 @@
...
+	 * Create menu/submenu entry.
+	 */
+	public function create_admin_menu() {
+
+		$current_user = wp_get_current_user();
+
+		if ( in_array( 'administrator', $current_user->roles, true ) && current_user_can( 'manage_options' ) ) {
+			add_submenu_page(
+				P4GBKS_PLUGIN_SLUG_NAME,
+				__( 'Classic block usages', 'planet4-blocks-backend' ),
```
lithrel, line 25: I would use `usage` (singular) because it is used as singular in `Blocks > Usage`

classes/controller/menu/class-classic-blocks-usages.php
```diff
@@ -0,0 +1,79 @@
...
+		};
+
+		$with_classic_blocks = array_filter( $all_posts, $has_classic_block );
+
+		// Copied this code from https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/add-classic-block-to-report/classes/controller/menu/class-blocks-usage-controller.php#L107-L107.
+		// So disabling CS here as well as this is only a temporary dev tool.
+		//phpcs:disable
+		echo '<hr>';
+		echo '<h2>Posts using classic blocks</h2>';
+		echo '<p>Following posts are detected to use classic blocks. It is recommended to convert them to new blocks.</p>';
```
lithrel, line 66: Is there a guide to convert to new blocks ? or a simple method or a url, something short we can write here to help the user ?

Inwerpsel: Normally classic blocks should allow to click "Convert to blocks" in the block top menu, which I haven't seen failing or producing wrong output. But maybe we want to double check if this is actually safe to do in all cases before we suggest doing that.

lithrel: We could point to it, or to some documentation about [classic block](https://wordpress.org/support/article/classic-block/) and [how to convert it](https://gogutenberg.com/blocks/classic/https://gogutenberg.com/blocks/classic/), and suggest contacting us if something goes wrong ?

comzeradd: Maybe that's also relevant: https://planet4.greenpeace.org/create/manage/convert-posts-to-gutenberg-editor/

Inwerpsel: @comzeradd Great I'll add the link. How reliable is this conversion? Are there cases where it may fail, or even produce wrong content that might go unnoticed? If that's the case maybe we should warn editors to thoroughly check their content before and after?

comzeradd: For the most part it is reliable. But you can never predict what kind of html will be there. You can see the content right in the editor after it's converted, before you save the post. And you can hit "Undo" on the top if you are not happy with the result. We could add that at the bottom of this handbook page.

lithrel (approved):
👍

PR #376, opened 2020-09-02
PLANET-4810: Convert Media Block to WYSIWYG

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4810

I'm testing a "canonical save" approach, it uses the `save()` method as is intended by Wordpress. 

There is a caveat to it:  the `save()` method needs to be a pure function with no side effects or hooks, static content is saved in editing time, so we can't change the styles of a block saved using this method in the future. For example, any changes to `MediaVideoElement` will cause a validation error in the block editor.

So we have 3 options when rendering blocks:
1. Render in the client side using our `frontendRendered`, which stores a static DOM node as a placeholder, in which we inject the React component at run time on the frontend.
2. Render on the server side using Twig and ServerSideRender.
3. Save the content as static content using `save()` and a "pure" component.

Currently we are leaning towards (1), this makes perfect sense for heavily interactive blocks as the Spreadsheet. Pros: we can do a lot using React in the frontend. Cons: it may have some penalizations for SEO and no-script situations or JS errors.

This block uses (3), which is the "WP endorsed" way of dealing with static blocks. Pros: *ideally*, the content is stored in its "pure" form, with no side effects or after-render actions. But for example, in the case of MP4 files, this block adds some DOM manipulations after the initial render. This should only be used in cases where the content is *absolutely* static. Cons: we can't change the block once its saved. Any changes to the code generating the output will cause a validation error in the editor. We do those kind of changes sometimes, so maybe we don't what that resilience to change.

Method (2), the Twig templates: maybe this is not a bad option for some blocks, as it allows for server-side rendering dynamic blocks and integrates well with some WP flows. Pros: it's rendered on the server side and we have full control of what happens *before* the output, preventing any FOUC. We can change the way a block is rendered over time and it won't hurt the block in the editor. We can make retroactive changes to blocks inside content. Cons: we have to replicate the block in React to make it WYSIWYG in the Editor, duplicating the markup in JSX and Twig versions.


assets/src/blocks/Media/MediaBlock.js
```diff
@@ -1,109 +1,48 @@
...
+
+console.log("MediaEditor", MediaEditor);
+
 export class MediaBlock {
     constructor() {
       const {registerBlockType} = wp.blocks;
 
-      registerBlockType('planet4-blocks/media-video', {
+      registerBlockType(BLOCK_NAME, {
         title: __('Media block', 'p4ge'),
```
mleray, line 34: I think we're now supposed to remove those `p4ge` translation domains?

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,41 @@
...
+    video_title,
+    description,
+    video_poster_img,
+    embedHtml,
+  } = attributes;
+
+  const wrappedHtml = wrapAndRemoveSize(embedHtml);
+
+  return (
+    <section className="block articles-block">
```
mleray, line 29: Is this `articles-block` class supposed to be there? 🤔

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,41 @@
...
+      <div className="container">
+        <header>
```
mleray, line 31: We should probably hide this if `video_title` is not defined, as we do below for `description` 🙂

mleray (changes requested):
I think there is still some old code that needs to be removed:
- in `test-shortcode-converter.php`: [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/tests/unit/test-shortcode-converter.php#L39) and [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/tests/unit/test-shortcode-converter.php#L395)
- in `class-shortcode-to-gutenberg.php`: [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/command/class-shortcode-to-gutenberg.php#L38)

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+const { __ } = wp.i18n;
+const { RichText } = wp.blockEditor;
+
+const renderEdit = ({ attributes, setAttributes }) => {
+  // Using a state to prevent the input losing the cursor position, a React issue reported multiple times
+  const [ mediaURL, setMediaURL ] = useState(attributes.youtube_id);
+  const debouncedMediaURLUpdate = useCallback(debounce(value => setAttributes({ youtube_id: value }), 300), []);
+
+  return (
+    <Fragment>
```
mleray, line 21: I don't think the `Fragment` is needed here, since there is only one child...? 🤔

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+const renderView = (props, toAttribute) => {
+  const { attributes, setAttributes } = props;
+  const { video_title, description, embedHtml, youtube_id, posterURL } = attributes;
+
+  function onSelectImage({id}) {
+    setAttributes({video_poster_img: id});
+  }
+
+  function onSelectURL({url}) {
+    setAttributes({id: null});
```
mleray, line 51: Is this `null` a mistake? Shouldn't it be `url`?

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+        className="page-section-description"
+        placeholder={__('Enter description', 'planet4-blocks-backend')}
+        value={description}
+        onChange={toAttribute('description')}
+        keepPlaceholderOnFocus={true}
+        withoutInteractiveFormatting
+        characterLimit={200}
+      />
+      <MediaPlaceholder
+        labels={{ title: __('Video poster image [Optional]', 'p4ge'), instructions: __('Applicable for .mp4 video URLs only.', 'p4ge')}}
```
mleray, line 86: These `p4ge` translation domains should be replaced by `planet4-blocks-backend`

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+    setAttributes({id: null});
+  }
+
+  function onUploadError({message}) {
+    console.log(message);
+  }
+
+  return (
+    <Fragment>
+      <Tooltip text={__('Edit text', 'planet4-blocks-backend')}>
```
mleray, line 60: I'm not sure that we need a tooltip here 🤔 We added one for the Articles block due to the default value for the title (therefore it was not really clear that it was editable), here since it will be empty at first the placeholder should be enough! May be good to double check with Magali 🙂

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+        onSelect={ onSelectImage }
+        onSelectURL={ onSelectURL }
+        onError={ onUploadError }
+        accept="image/*"
+        allowedTypes={["image"]}
+      />
+      {
+        (youtube_id && !youtube_id.endsWith('.mp4')) && !embedHtml
+        ? <div className="block-edit-mode-warning components-notice is-error">
+            { __( 'The video URL could not be parsed.', 'planet4-blocks' ) }
```
mleray, line 97: The translation domain here should be `planet4-blocks-backend` 🙂

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+          </div>
+        : <Fragment>
+            {
+              youtube_id && youtube_id.endsWith('.mp4')
+              ? <MediaElementVideo videoURL={ youtube_id } videoPoster={ posterURL } />
+              : <MediaEmbedPreview html={ embedHtml ? embedHtml : null } />
```
mleray, line 103: This could be shortened to `embedHtml || null`

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+    return {
+      embedHtml: embedPreview ? embedPreview.html : null,
+      posterURL: media ? media.media_details.sizes.large.source_url : null
+    }
+  },
+  [ youtube_id, video_poster_img ]);
+
+  useEffect(() => {
+    setAttributes({
+      embedHtml: embedHtml ? embedHtml : null,
```
mleray, line 139: Same as above, these could be shortened to `embedHtml || null` & `posterUrl || null`

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,155 @@
...
+      posterURL: posterURL ? posterURL : null
+    })
+  },
+  [ embedHtml, posterURL ]);
+
+  return (
+    <div>
+      {
+        renderEdit(props)
```
mleray, line 148: Shouldn't this be rendered only when `isSelected` is true? 🤔

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,52 @@
...
+          </header>
+        }
+        {
+          description &&
+          <div className="page-section-description" dangerouslySetInnerHTML={{ __html: description }} />
+        }
+        {
+          youtube_id.endsWith('.mp4')
+          ? <MediaElementVideo videoURL={ youtube_id } videoPoster={ posterURL } />
+          : <MediaEmbedPreview html={ wrappedHtml ? wrappedHtml : null } />
```
mleray, line 47: Same as above, this could be shortened to `wrappedHtml || null`

assets/src/blocks/Media/MediaEmbedPreview.js
```diff
@@ -0,0 +1,3 @@
+export const MediaEmbedPreview = (props) => {
```
mleray, line 1: Not a big deal, but this could be a tiny bit cleaner by removing the `return` statement, a.k.a 
```
export const MediaEmbedPreview = props => <div dangerouslySetInnerHTML={{ __html: props.html }} />;
```
(same for `MediaElementVideo` above actually 🙂)

mleray (changes requested):
Looks pretty good 👍  I've added a few comments, mostly for small code clean-ups & improvements! Also the failing tests need to be fixed, by removing the [old acceptance tests](https://github.com/greenpeace/planet4-base-fork/blob/develop/tests/acceptance/MediaCept.php) (+ create follow-up ticket to add JS acceptance tests) and fix some php spacing 🙂

assets/src/blocks/Media/MediaEditor.js
pablocubico: Yep, I just ported this here and forgot about it. I just checked, and this is also present in the old Media Block, selecting a URL using the "Insert via URL" button does nothing (!). I removed this listener and the `onSelectURL` property from the component, which prevents the button from rendering at all. 

cc @comzeradd

assets/src/blocks/Media/MediaEditor.js
pablocubico: Ok, I'll check!

assets/src/blocks/Media/MediaEditor.js
pablocubico: So here's a thing which I actually asked in the React conference on Friday (I didn't get a proper answer): if we use `isSelected` here, React complains about the order of hooks changing (I hate that error!). I see `renderEdit` as a function component even if we are calling it as a function... I thought the "hooks order" error wouldn't trigger in a child component if it's rendered conditionally, but maybe it is triggered. 

So I moved the `isSelected` check to inside `renderEdit` after the hooks are set, right before the `return` call.

pablocubico: Ok, I found another solution: instead of calling `renderEdit`, I turned it into a component by using it as a JSX tag. It's the same function code but invoked with a different syntax. This fixes the "order of hooks" warning. I guess it changes depending on the way React wraps components when it founds a JSX tag instead of you invoking a function directly.

pablocubico:
@mleray comments addressed and rebased to master!

mleray (changes requested):
After testing this on local, it looks like existing blocks are not rendered at first and need to be re-saved! We could fix that by adding a `render_callback` as we did for other blocks 🙂

pablocubico:
> After testing this on local, it looks like existing blocks are not rendered at first and need to be re-saved! We could fix that by adding a `render_callback` as we did for other blocks 🙂

@mleray So I added the `render_callback` as in the other blocks but I just realized this is a static block, the output of `save()` is stored as is and the block is rendered statically, so it doesn't need it... I need to figure out what to do with existing content, hold on! 🤔

assets/src/blocks/Media/MediaEditor.js
Inwerpsel: Can't we even shorten it more to just `embedHtml`?

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,61 @@
...
+
+const wrapAndRemoveSize = embedHtml => {
+  const wrapperDiv = document.createElement('div');
+  wrapperDiv.innerHTML = embedHtml;
+  if (wrapperDiv.firstChild && wrapperDiv.firstChild.src && wrapperDiv.firstChild.src.indexOf('youtube') > -1) {
+    wrapperDiv.className = 'embed-container';
+    wrapperDiv.firstChild.removeAttribute('width');
+    wrapperDiv.firstChild.removeAttribute('height');
+    return wrapperDiv.outerHTML;
+  } else {
```
Inwerpsel, line 15: Return early here. We could add a rule to enforce that https://eslint.org/docs/rules/no-else-return

Inwerpsel: Probably better to invert the if, so that we only need to put `return embedHtml` inside it, instead of the preceding 4 lines.

Inwerpsel (changes requested):
Having the frontend wait for php, which in turns waits for youtube's (or other) oembed endpoint response so that it finally can start performing requests for the embedded content, seems like something we probably want to avoid unless there is no other way, as it will be really slow.

We can have a look at the different embed types we offer and whether any of them has content that changes over time. If none of them do then saving it statically seems to make a lot more sense.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,143 @@
...
+        allowedTypes={["image"]}
+      />
+      {
+        (youtube_id && !youtube_id.endsWith('.mp4')) && !embedHtml
+        ? <div className="block-edit-mode-warning components-notice is-error">
+            { __( 'The video URL could not be parsed.', 'planet4-blocks-backend' ) }
+          </div>
+        : <Fragment>
+            {
+              youtube_id && youtube_id.endsWith('.mp4')
```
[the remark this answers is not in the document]
Inwerpsel: But I understand if you don't want to put the burden of that on this PR :slightly_smiling_face: We could address it later, but then we should probably already map the attribute to a variable with an appropriate name as early as possible, and not use it directly unless unavoidable.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,143 @@
...
+import { MediaEmbedPreview } from "./MediaEmbedPreview";
+import { MediaElementVideo } from './MediaElementVideo';
+import { useSelect } from '@wordpress/data';
+import { debounce } from 'lodash';
+
+const { __ } = wp.i18n;
+const { RichText } = wp.blockEditor;
+
+const MediaInspectorOptions = ({ attributes, setAttributes }) => {
+  // Using a state to prevent the input losing the cursor position, a React issue reported multiple times
```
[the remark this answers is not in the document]
pablocubico: Using a cached state locally (instead of the attribute directly) prevented this issue: 
https://github.com/facebook/react/issues/14904

I haven't tried turning it into an uncontrolled component and using `defaultValue` as stated in the last comment of the issue, it is a bit of a hack. Uncontrolled components are not totally ruled out by React, but I'd say they are mildly discouraged. 

I couldn't trace why it happens when you set the attribute directly, using a local state fixed it.

assets/src/blocks/Media/MediaEditor.js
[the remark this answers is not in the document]
pablocubico: Yeah, this variable name is vestigial, from old versions of the block, let's map it to a proper name earlier.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,143 @@
...
+    const media = select('core').getMedia(video_poster_img)
+    return {
+      embedHtml: embedPreview ? embedPreview.html : null,
+      posterURL: media ? media.media_details.sizes.large.source_url : null
+    }
+  },
+  [ youtube_id, video_poster_img ]);
+
+  useEffect(() => {
+    setAttributes({
```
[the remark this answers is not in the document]
pablocubico: This effect is part of the "static render" path:

In order to render the block statically with no endpoint calls we need to store the embed's HTML and the poster image URL.

As that content is fetched when the Media URL or the image ID changes, the effect itself is more or less a change listener for ` embedHTML` and `posterURL`, which are retrieved in the previous `useSelect` call right before this effect. Once we know the content of `embedHTML` and `posterURL`, we store them as attributes for the block to render statically. 

I'll add this as a comment to the code.

assets/src/blocks/Media/MediaEditor.js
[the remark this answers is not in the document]
pablocubico: See also: https://github.com/facebook/react/issues/955#issuecomment-469352730

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,143 @@
...
+          </Fragment>
+      }
+    </Fragment>
+  )
+}
+
+export const MediaEditor = (props) => {
+  const { attributes, setAttributes, isSelected } = props;
+  const { youtube_id, video_poster_img } = attributes;
+  let mediaURL = youtube_id;
```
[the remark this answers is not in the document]
pablocubico: Yes!

assets/src/blocks/Media/MediaElementPlugin.js
```diff
@@ -0,0 +1,51 @@
+function addScripts(scriptTagSources, onFinish){
```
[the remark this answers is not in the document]
pablocubico: I guess... provided we merge the Timeline first 🤔

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,61 @@
...
+    return embedHtml;
+  }
+}
+
+export const MediaFrontend = (props) => {
+  const mediaData = useMediaDataFetch(props);
```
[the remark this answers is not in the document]
pablocubico: We have this block without the endpoint, I think we already chatted about this, the endpoint is only here to support the rendering of old content. New content will be saved statically.

That said, we still need to think about the pros & cons of using static blocks... the "resilient to change" vs "easy to change" dilemma. 🤔

pablocubico:
> Having the frontend wait for php, which in turns waits for youtube's (or other) oembed endpoint response so that it finally can start performing requests for the embedded content, seems like something we probably want to avoid unless there is no other way, as it will be really slow.
> 
> We can have a look at the different embed types we offer and whether any of them has content that changes over time. If none of them do then saving it statically seems to make a lot more sense.

As said in one of the comments, the endpoint is only for old content, new content is statically saved and rendered directly here: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/376/files#diff-ea623fe817268e0262e2bec08bdf6098R39

This is just to support old content without a migration, we can remove it if we migrate old Media blocks to static blocks in the near future.

tests/unit/test-shortcode-converter.php
```diff
@@ -34,7 +34,11 @@ public function setUp() {
 	 * @dataProvider carousel_header_shortcodes_provider
 	 * @dataProvider columns_shortcodes_provider
 	 * @dataProvider covers_shortcodes_provider
+<<<<<<< HEAD
```
sagarsdeshmukh, line 37: looks like leftover while rebase.

pablocubico: Got it, thanks!

classes/blocks/class-media.php
```diff
@@ -23,41 +23,22 @@ class Media extends Base_Block {
...
 		register_block_type(
 			'planet4-blocks/media-video',
 			[
 				'editor_script'   => 'planet4-blocks',
-				'render_callback' => [ $this, 'render' ],
+				'render_callback' => static function ( $attributes, $content ) {
+					if ( '' === trim( $content ) ) {
+						$json = wp_json_encode( [ 'attributes' => $attributes ] );
+						return '<div data-render="planet4-blocks/media-video" data-attributes="' . htmlspecialchars( $json ) . '"></div>';
+					} else {
```
Inwerpsel, line 38: No need for the else here

Inwerpsel: A bit more straightforward to invert the condition here, as less will be nested.
```
if ( '' !== trim($content)) {
    return $content;
}

return '<div data-render="planet4-blocks/media-video" data-attributes="' . htmlspecialchars( $json ) . '"></div>';
```

classes/blocks/class-media.php
```diff
@@ -23,41 +23,22 @@ class Media extends Base_Block {
...
-		add_shortcode( 'shortcake_media_video', [ $this, 'add_block_shortcode' ] );
-
 		register_block_type(
 			'planet4-blocks/media-video',
 			[
 				'editor_script'   => 'planet4-blocks',
-				'render_callback' => [ $this, 'render' ],
+				'render_callback' => static function ( $attributes, $content ) {
+					if ( '' === trim( $content ) ) {
+						$json = wp_json_encode( [ 'attributes' => $attributes ] );
```
Inwerpsel, line 36: Instead of setting up the block to be frontend rendered here, we could also fetch the dynamic data here already. This will not add to the page's response time as the result is cached along with the rest of the page, and will periodically update as the cache gets invalidated, which should be more than sufficient as this data changes infrequently.

assets/src/blocks/Media/MediaBlock.js
```diff
@@ -1,109 +1,54 @@
...
-                },
-                video_poster_img: {
-                  type: 'integer',
-                  shortcode: ({named: {video_poster_img = ''}}) => video_poster_img,
-                }
-              },
-            },
-          ]
-        },
+        deprecated: [{
```
Inwerpsel, line 31: Can you move this `deprecated` key to be the last one? It grows over time as we add more entries, that way we don't have it push the more relevant config down.

classes/blocks/class-media.php
Inwerpsel:
```
json = wp_json_encode( 
    [ 'attributes' => array_merge($attributes, Media::get_media_data($attributes) ] 
);
```

Inwerpsel: Then the only thing we need to do is remove the `useMediaData` hook call, as the resulting attributes will be there already.
```javascript
export const MediaFrontend = (props) => {
  // const mediaData = useMediaDataFetch(props);

  // const staticProps = { ...props, ...mediaData };

  const staticProps = props;

  return <StaticMediaFrontend { ...staticProps } />
}
```

Inwerpsel (changes requested):
We can easily avoid introducing another endpoint call for existing content if we just put the props in place in the backend render call, instead of having the backend render call set up front end rendering to fetch the same props.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,143 @@
+import { Fragment, useEffect, useState, useCallback } from "@wordpress/element";
+import { PanelBody } from '@wordpress/components';
+import { MediaPlaceholder } from "@wordpress/editor";
```
lithrel, line 3: This import is deprecated, we have to import it from `wordpress/block-editor`

lithrel (commented):
Shouldn't the _Video post image_ selector be on the side panel, so that the appearance of the block doesn't differ too much from the live one ?

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,143 @@
...
+import { useSelect } from '@wordpress/data';
+import { debounce } from 'lodash';
+
+const { __ } = wp.i18n;
+const { RichText } = wp.blockEditor;
+
+const MediaInspectorOptions = ({ attributes, setAttributes }) => {
+  // Using a state to prevent the input losing the cursor position, a React issue reported multiple times
+  const [ mediaURL, setMediaURL ] = useState(attributes.youtube_id);
+  const debouncedMediaURLUpdate = useCallback(debounce(value => setAttributes({ youtube_id: value }), 300), []);
```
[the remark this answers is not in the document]
pablocubico: Ah, the debounce is there to prevent making a request to the oEmbed API on every keystroke, it's not related to the cursor position issue.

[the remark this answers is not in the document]
Inwerpsel: Ah ok :+1: . My suggestion still holds as it seems a simpler way to achieve the same thing. The debounce would be bit simpler too then:
```
        <URLInput
          value={ mediaURL }
          onChange={value => debounce(() => setMediaURL(value), 300) }
         // ...
        />
```

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,54 @@
...
+  const {
+    video_title,
+    description,
+    embedHtml,
+    posterURL,
+    youtube_id,
+  } = props;
+
+  // Deprecate: rename this attribute.
+  const mediaURL = youtube_id;
```
Inwerpsel, line 29: Maybe we can already find the new name before we apply it. We use other attributes with their snake case name in the code, so if we make this variable already the new snake case name, we only need to remove this line and extract the value from the props instead.

assets/src/frontendIndex.js
```diff
@@ -24,3 +28,6 @@ document.querySelectorAll( `[data-render]` ).forEach(
     wp.element.render( <BlockFrontend { ...attributes.attributes } />, blockNode );
   }
 );
+
+// Other non-react initializations
+setupMediaElementJS();
```
Inwerpsel, line 33: This will add to the render blocking time of `frontendIndex.js`. Do we have an alternative?

assets/src/blocks/Media/MediaEditor.js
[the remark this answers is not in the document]
pablocubico: I had to defer to `useCallback` because of this issue: 
https://stackoverflow.com/questions/55616536/lodash-debounce-in-react-functional-component-not-working
https://dev.to/pulkitnagpal/using-throttling-and-debouncing-with-react-hooks-57f1

I tested your snippet anyway just in case I was wrong but it actually prevents you from typing anything in the input field. 

At this point I would avoid looking for more elegant ways of doing something that's working with no issues to prevent any further delays, it's hard to determine which comments are blockers and which aren't. Maybe we should use a "traffic light" system (red, yellow, green) to flag these.

Note: this block is probably going to be deprecated and replaced by the native WP blocks in the near future (cc @comzeradd )

assets/src/frontendIndex.js
pablocubico: These scripts are not blocking (as far as I see), they are dynamically added to the body (well, the style is attached to the head), way after the first contentful paint, the blocking scripts are pretty much the ones we are enqueueing from PHP (e.g.: the files for "slick"), as seen here:

blockingscripts

The CSS/JS files for MediaElement are not shown as blocking in the Lighthouse report.

Inwerpsel: Maybe if we merge the Timeline PR first, then we could do these things with `useStylesheet` and `useScript` instead.

Inwerpsel (commented):
Just commenting these already as we might be able to leverage `useStylesheet` and `useScript` from [the Timeline PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/368) so that these resources don't need to be setup in `frontendIndex` and seems logical to solve this the same way for both blocks, so want to check on that a last time before approving.

assets/src/blocks/Media/MediaEditor.js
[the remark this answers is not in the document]
pablocubico: But what would be the benefit of adding another hook? It may look simpler in the `onChange` call, but why adding an effect when there is no real need for an effect there? There is [a comment](https://stackoverflow.com/questions/55616536/lodash-debounce-in-react-functional-component-not-working/58929988#comment109952426_58929988) to that solution actually suggesting this.

assets/src/frontendIndex.js
pablocubico: Sounds good, I'll move those two hooks outside the `Timeline` dir in the Timeline PR.

assets/src/blocks/Media/MediaEditor.js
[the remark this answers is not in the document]
Inwerpsel: Yeah maybe that would only make sense if there were multiple things calling `setMediaURL`, then we would only have to add the debounce in one place. But since that won't be the case here, feel free to ignore these comments :+1:

assets/src/blocks/Media/MediaFrontend.js
pablocubico: Ok, NOW it's 1000% done, double-checked. Resolving.

pablocubico:
@lithrel Indeed, I moved the video poster selector to the sidebar.
@Inwerpsel `youtube_id` deprecated & migrated, the MediaElementJS plugin is now loaded via `useScript` & `useStylesheet`.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,150 @@
...
+  const parsedMediaURL = media_url && media_url.indexOf('/') === -1
+    ? `https://www.youtube.com/watch?v=${media_url}`
+    : media_url;
+
+  console.log('media URL', parsedMediaURL);
+
+  const toAttribute = attributeName => value => setAttributes({ [attributeName]: value });
+
+  if ( parsedMediaURL && parsedMediaURL.indexOf('youtube.com') > -1 ) {
+    parsedMediaURL.replace('youtube.com', 'youtube-nocookie.com');
```
Inwerpsel, line 114: This won't actually replace in the url that's being used.

Inwerpsel: While you're changing it, probably best to extract the logic for which URL to use based on the `media_url` prop into a separate function, so that the complexity doesn't complicate this high level render method.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,150 @@
...
+export const MediaEditor = (props) => {
+  const { attributes, setAttributes, isSelected } = props;
+  const { media_url, video_poster_img } = attributes;
+
+  // Assume that a non-URL is a YouTube video ID.
+  const parsedMediaURL = media_url && media_url.indexOf('/') === -1
+    ? `https://www.youtube.com/watch?v=${media_url}`
+    : media_url;
+
+  console.log('media URL', parsedMediaURL);
```
Inwerpsel, line 109: Can you remove this logging? We should probably setup `eslint` to disallow these.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,150 @@
...
+
+  const { embed_html, poster_url } = useSelect((select) => {
+    const embedPreview = select('core').getEmbedPreview( parsedMediaURL );
+    const media = select('core').getMedia(video_poster_img)
+    return {
+      embed_html: embedPreview ? embedPreview.html : null,
+      poster_url: media ? media.media_details.sizes.large.source_url : null
+    }
+  },
+  [ media_url, video_poster_img ]);
```
Inwerpsel, line 125: Even though this `useSelect` is originally used to map something to props in the same way `withSelect` does, for  `useSelect` this is actually vestigial and not needed anymore.

It was previously mapped to props because the HOC paradigm works that way, but with hooks we actually don't need these intermediary props anymore since we're already inside of render. So we can avoid some duplication here:

```js
const [ embed_html, poster_url ] = useSelect(select => {
    const embedPreview = select('core').getEmbedPreview( parsedMediaURL );
    const media = select('core').getMedia(video_poster_img);

    return [
      embedPreview ? embedPreview.html : null,
      media ? media.media_details.sizes.large.source_url : null
    ]
  },
  [ media_url, video_poster_img ]
);

Comment (Inwerpsel) [reply assets/src/blocks/Media/MediaEditor.js:125]: Actually 2 separate `useSelect` calls is even more readable:
```js
  const embed_html = useSelect(select => {
    const embedPreview = select('core').getEmbedPreview(parsedMediaURL);

    return embedPreview ? embedPreview.html : null;
  }, [media_url]);

  const poster_url = useSelect(select => {
    const media = select('core').getMedia(video_poster_img);

    return media ? media.media_details.sizes.large.source_url : null;
  }, [video_poster_img]);

File: assets/src/blocks/Media/MediaElementVideo.js [hunk 32]
```diff
@@ -0,0 +1,60 @@
...
+	const [stylesLoaded, stylesError] = useStyleSheet(
+    `https://cdn.jsdelivr.net/npm/mediaelement@4.2.16/build/mediaelementplayer-legacy.min.css`
+	);
+
+	const [scriptLoaded, scriptError] = useScript(
+		`https://cdn.jsdelivr.net/npm/mediaelement@4.2.16/build/mediaelement-and-player.min.js`
+	);
+
+	const [vimeoRendererLoaded, vimeoRendererError] = useScript(
+		`https://cdn.jsdelivr.net/npm/mediaelement@4.2.16/build/renderers/vimeo.min.js`
```
Inwerpsel, line 16: Shouldn't we try to only load this if it's actually a vimeo video?

assets/src/blocks/Media/MediaElementVideo.js
```diff
@@ -0,0 +1,60 @@
...
+	);
+
+  const setupMediaElementJS = function() {
+		const meJSNodes = document.querySelectorAll( '.mejs-video-block' );
+
+		meJSNodes.forEach(node => {
```
Inwerpsel, line 22: What is the behavior here if a page has many video blocks? Does each render also setup the player for all other blocks n times?

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -0,0 +1,51 @@
+import { MediaEmbedPreview } from './MediaEmbedPreview';
+import { MediaElementVideo } from './MediaElementVideo';
+
+const { __ } = wp.i18n;
+
+const wrapAndRemoveSize = embed_html => {
+  const wrapperDiv = document.createElement('div');
```
Inwerpsel, line 7: Can't we do this before saving the embed html? Maybe here https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/376/files#diff-c6c91408622162ae2bd090b6d50b303d1de0a3ebff4e385997189643c4a18ce1R121

Inwerpsel (changes requested):
Currently turning the youtube url into a nocookie one will not work anymore, definitely one to address. Also some things we can simplify a lot.

assets/src/blocks/Media/MediaElementVideo.js
pablocubico: Indeed, but now that you mention it, this script originally added by Wordpress, because Vimeo allowed to link .directly to the .mp4 source of your video, but I'm not sure that's the case anymore, I have some videos there, I'll do a quick check.

assets/src/blocks/Media/MediaElementVideo.js
pablocubico: Oops, forgot to update this, it should use a ref to the node as the Timeline does.

assets/src/blocks/Media/MediaElementVideo.js
pablocubico: After double checking with a Vimeo video, I couldn't find any reason why this was needed in the first place. Vimeo links go through the oEmbed part of the code, you can link directly to the MP4 source of the file in Vimeo but the URL is a download link, it does not end with `.mp4` so it won't be used in this caso, and if it were, it's just a regular MP4 file. So I deleted this dependency.

assets/src/blocks/Media/MediaElementVideo.js
pablocubico: So there was a problem with using the `useScript` and `useStylesheet` hooks: we can't use hooks in the `save()` method. So I extracted the part of the code that actually creates the script/link elements and adds them to the DOM into two functions that can be used inside the hooks or without them.

assets/src/blocks/Media/MediaEditor.js
pablocubico: Ugh! Well spotted. 👁️

assets/src/blocks/Media/MediaElementVideo.js
Inwerpsel: Maybe we can use `hydrate` on the frontend and still be able to use these hooks?

pablocubico: Wouldn't `hydrate` miss the point of making the block "static" and resilient to change? Since these are not server side rendered by React but saved statically at editing time into the database, any change in the component would trigger a hydration warning.

pablocubico: That is, any change that causes the resulting markup to be different.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,153 @@
...
+          </Fragment>
+      }
+    </Fragment>
+  )
+}
+
+const resolveURL = url => {
+  // If it's a Youtube ID, turn it into a Youtube URL
+  // Youtube IDs are 11 chars long strings with the given set of chars
+  const result = /^[a-z0-9_-]{11}$/i.test(url)
```
Inwerpsel, line 103: It's easier to read if you directly return without a `const`

assets/src/blocks/Media/MediaElementVideo.js
Inwerpsel: It would indeed make the block harder to change, but if we know the markup will rarely change that might be less of an issue. It would have the advantage of not having some separate logic somewhere else that needs to be called for the block to work.

assets/src/blocks/Media/MediaFrontend.js
pablocubico: Hm, initially I thought we could but I think we'll need to keep it in the MediaFrontend file for old content.

Inwerpsel: If it's only for old content, we're already doing something for backwards compatibility in the back end, we might as well handle that there too. It only needs to remove the width and height attributes, and put it in a wrapper div. It's already doing the youtube url replacement https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/9fc1e628ca2a2bfc65d6570729afc9fac91793d2/classes/blocks/class-media.php#L124-L124

pablocubico: Well, is not only for old content, is both for old and new content, maybe we should remove the PHP part instead.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -0,0 +1,153 @@
...
+    setAttributes({video_poster_img: id});
+  }
+
+  return <InspectorControls>
+      <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
+        <TextControl
+          label={__('Media URL/ID', 'planet4-blocks-backend')}
+          placeholder={__('Enter URL', 'planet4-blocks-backend')}
+          value={ media_url }
+          onChange={
```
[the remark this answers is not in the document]
pablocubico: Looks like this is also happening with other fields in different blocks (e.g.: Spreadsheet URL in Spreadsheet, Post Types in Articles), so I guess it deserves a new ticket.

assets/src/blocks/Media/MediaFrontend.js
pablocubico: I went the other way around: removed it from PHP. Since it adds a wrapper div with a class, removes the width and height and replaces the URL, I thought it was more consistent to do it in a single place for all the cases. Using DOM methods is also a bit cleaner & safer than doing `str_replace` or `ereg_replace` in the PHP side.

pablocubico:
@Inwerpsel picking up from our latest conversations (you probably saw my comments via Toast notifications):
* undo/redo: it wasn't because of the state, it's happening on other blocks, I'll create a ticket for that
* wrapEmbedHTML: happening on the `MediaFrontend` block, not at `save()`-time nor in the PHP side
* other minor comments addressed

assets/src/blocks/Media/MediaBlock.js
```diff
@@ -1,109 +1,64 @@
...
+        type: 'string',
+        default: ''
+      },
+    },
+    save: ({ attributes }) => {
+      return <MediaFrontend { ...attributes } />
+    },
+    edit: MediaEditor,
+    deprecated: [{
+      attributes: {
```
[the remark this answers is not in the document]
Inwerpsel: The last part to preserve history can be a bit cumbersome for this PR as you need to rebase on top of those steps :stuck_out_tongue: You can do it if you feel like it but definitely not a requirement for me :smile: 

But just moving the deprecated config to a separate file in a `deprecated` directory could be an easy improvement to make the current version's config more readable. Like in the branch I made:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/5126415228fc9e44b227609c340bc302e0eaafd3/assets/src/blocks/Media/MediaBlock.js#L108-L110

Inwerpsel (commented):
I'll finish the review tomorrow. Looks good so far, left some suggestions.

If we fallback with JS we probably want to run a migration soon as the difference with static html will probably be noticeable (assumption, still need to confirm the difference). 

Thinking of that migration, it's actually going to be a bit tricky as we need to run JS to get the new block state, probably some setup involved in having that run on all posts :thinking:

assets/src/blocks/Media/MediaBlock.js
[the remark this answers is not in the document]
Inwerpsel: Looks similar to how it's done in Gutenberg: https://github.com/WordPress/gutenberg/blob/master/packages/block-library/src/audio/index.js#L31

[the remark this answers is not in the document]
pablocubico: Sounds good. I think we should start documenting some P4 "best practices" somewhere (Gitbook?) and then make the rest of the blocks match those best practices.

assets/src/blocks/Media/MediaEditor.js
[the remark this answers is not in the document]
pablocubico: Hmm no, I don't think so, can you come up with a test case?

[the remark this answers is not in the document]
Inwerpsel: Well it's effectively a component modifying its own props in render. But indeed in its current form the code won't go in a loop because the effect ensures it only modifies 2 certain props right after 2 other props have changed.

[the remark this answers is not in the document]
Inwerpsel: Shouldn't we ensure instead that the `embed_html` prop gets set whenever the `media_url` prop gets set?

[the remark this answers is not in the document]
Inwerpsel: Or is this to account for old content not having the `embed_html` prop yet?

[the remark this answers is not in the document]
pablocubico: Hmmm not sure I follow, `embed_html` and `poster_url` are actually local in this scope, so `embed_html` is not obtained from `attributes`, it's the result of the `useSelect` call above which is fired when `media_url` changes. 

This is not happening in render, it's inside an effect. And the effect's dependencies lists exactly the values it should check to avoid unnecessary re-renders.. 

What is the real risk here? We could discuss every line on a PR but we need to find closure at some point, this has 115 comments already. 💥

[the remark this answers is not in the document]
pablocubico: I mean, re-renders caused by the attribute being set over and over by the effect.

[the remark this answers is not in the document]
pablocubico: This was superseded by the merged PR. https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/423

classes/blocks/class-media.php
```diff
@@ -23,70 +23,75 @@ class Media extends Base_Block {
...
+					if ( isset( $attributes['youtube_id'] ) ) {
+						$attributes['media_url'] = $attributes['youtube_id'];
+						unset( $attributes['youtube_id'] );
+					}
+
+					$json = wp_json_encode(
+						[ 'attributes' => $attributes ]
+					);
+
+					return '<div data-render="planet4-blocks/media-video" data-attributes="' . htmlspecialchars( $json ) . '"></div>';
```
Inwerpsel, line 50: Since this will only run on old content, and so the logic needed is essentially "frozen", we might as well make it render the whole block in PHP. Then we could reduce the size of our front end bundle by not having to include the component code. Also page render time would benefit from less script execution/DOM manipulation.

We would need to write that logic in PHP anyway if we want to migrate old content to the new saved blocks.

But the current PR works so we can consider that in a follow up.

Inwerpsel (approved):
Looks good :+1: :+1: :+1: 

I recommend still making `useScript` and `useStylesheet` not retrying on error, it could result in the resource being added and removed on each render if the error persists. But probably not an issue for the usage in this PR :shrug:

assets/src/components/useStyleSheet/addLinkTag.js
```diff
@@ -0,0 +1,22 @@
...
+	onError
+}) => {
+	const link = document.createElement('link');
+	link.rel = 'stylesheet';
+	link.href = href;
+	link.media = media;
+
+	link.addEventListener('load', onLoad);
+	link.addEventListener('error', event => {
+		link.remove();
```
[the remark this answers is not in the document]
pablocubico: Removed in both cases. 👍

mleray (approved):
We need to remember to create a follow-up ticket to rewrite the acceptance tests in JS, but apart from that looks good to me 👍

PR #377, opened 2020-09-02
PLANET-5439 Search only in post types that could have blocks

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5439

Blocked by https://github.com/greenpeace/planet4-master-theme/pull/1175 as it uses a function introduced there.

Especially since we added archived posts as a post type, this extra condition is needed because we're regex matching each post's content.

~~Maybe we should double check whether any NRO has custom post types that can contain blocks. If so maybe we can invert the where to exclude certain post types instead (would mainly be archive posts, but there may be others).~~ Actually we can figure out which posts types are registered that support the block editor, so we don't need to have a fixed list.

Before (GPI staging): https://release.k8s.p4.greenpeace.org/international/wp-admin/admin.php?page=plugin_blocks_report
After (Test instance with GPI data): https://k8s.p4.greenpeace.org/test-nix/wp-admin/admin.php?page=plugin_blocks_report

### DIFF
--- a/classes/controller/menu/class-blocks-usage-controller.php
+++ b/classes/controller/menu/class-blocks-usage-controller.php
@@ -8,6 +8,7 @@
 
 namespace P4GBKS\Controllers\Menu;
 
+use P4\MasterTheme\SqlParameters;
 use WP_Block_Type_Registry;
 
 if ( ! class_exists( 'Blocks_Usage_Controller' ) ) {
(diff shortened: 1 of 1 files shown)


lithrel (approved):
👍

lithrel (approved):
:+1: :+1:

PR #378, opened 2020-09-03
PLANET-5424 Spreadsheet block: URL vanishes after page being published

dpivo:
set url value to attributes.url as opposed to state.url

### DIFF
--- a/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
+++ b/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
@@ -45,7 +45,7 @@ export class SpreadsheetEditor extends Component {
     this.state = {
       invalidSheetId: false,
       errorFetchingSpreadsheet: false,
-      url: '',
+      url: props.attributes.url,
     };
 
     this.debounceSearch = debounce(url => {


mleray (changes requested):
This will cause a delay when typing in the input, since the attributes' url is updated with a [debounce function](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L51) A solution would be to initialise `this.state.url` to `attributes.url` instead of the current `''` ([here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L48))

dpivo:
> This will cause a delay when typing in the input, since the attributes' url is updated with a [debounce function](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L51) A solution would be to initialise `this.state.url` to `attributes.url` instead of the current `''` ([here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L48))

I don't think the attributes object is visible at that point. So I set it on renderEdit(). Let me know if it's ok :ok_hand:

mleray:
> > This will cause a delay when typing in the input, since the attributes' url is updated with a [debounce function](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L51) A solution would be to initialise `this.state.url` to `attributes.url` instead of the current `''` ([here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L48))
> 
> I don't think the attributes object is visible at that point. So I set it on renderEdit(). Let me know if it's ok 👌

`props.attributes.url` should be available? 🤔

dpivo:
> > > This will cause a delay when typing in the input, since the attributes' url is updated with a [debounce function](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L51) A solution would be to initialise `this.state.url` to `attributes.url` instead of the current `''` ([here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js#L48))
> > 
> > 
> > I don't think the attributes object is visible at that point. So I set it on renderEdit(). Let me know if it's ok ok_hand
> 
> `props.attributes.url` should be available? thinking

ah right, yes. I implemented the change, thanks! :pray:

mleray (approved):
Looks good to me 👍

PR #380, opened 2020-09-07
PLANET-5426 Image block - Removed the blue caption style

sagarsdeshmukh:
Ref. [JIRA 5426](https://jira.greenpeace.org/browse/PLANET-5426)

- Removed the "Blue Caption" style option from Image block backend(block Editor option)
- Removed blue caption styles code
- Added fallback solution for existing Image blocks with "blue-caption" style

**e.g.**
Page: https://k8s.p4.greenpeace.org/test-titan/test-image-block-with-blue-captionpage/
Post: https://k8s.p4.greenpeace.org/test-titan/story/28934/test-image-block-with-blue-captionpost/
Campaign:  https://k8s.p4.greenpeace.org/test-titan/campaign/image-block-with-blue-captioncampaign/

### How to fix the existing Image blocks which are using the “blue-caption” style?
**Proposed solution:** Add a migration script which check all page, post, campaigns for Image blocks with blue caption style and update them with medium caption style value. 

**The fallback solution:** make alias of “caption-style-medium” css class to “caption-style-blue-overlay”. This way if the image block having “blue-caption” style will work same as medium caption style.

Any thoughts/suggestions on above would be welcome :)

### DIFF
--- a/assets/src/ImageBlockExtension.js
+++ b/assets/src/ImageBlockExtension.js
@@ -11,10 +11,6 @@ const targetBlocks = [
 
 const captionStyleOptions = [
   {
-  label: __( 'Blue Overlay (default)' ),
-  value: 'blue-overlay',
-  },
-  {
   label: __( 'Medium' ),
   value: 'medium',
   },
--- a/assets/src/ImageBlockExtension.js
+++ b/assets/src/ImageBlockExtension.js
@@ -1,4 +1,4 @@
-import { Button, ButtonGroup, PanelBody, SelectControl } from '@wordpress/components';
+import { Button, ButtonGroup, PanelBody } from '@wordpress/components';
 import assign from 'lodash.assign';
 
 const { addFilter } = wp.hooks;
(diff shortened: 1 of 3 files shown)


Inwerpsel:
> ### How to fix the existing Image blocks which are using the “blue-caption” style?
> **Proposed solution:** Add a migration script which check all page, post, campaigns for Image blocks with blue caption style and update them with medium caption style value.
> 
> **The fallback solution:** make alias of “caption-style-medium” css class to “caption-style-blue-overlay”. This way if the image block having “blue-caption” style will work same as medium caption style.
> 

The fallback solution seems better as that would also work on imported older content, and avoid the need to modify the data. Instead of adding an alias to the css, we could also dynamically add the `caption-style-medium` class to each block that has the `caption-style-blue-overlay` class.

Inwerpsel (approved):
Looks good :+1: The approach you chose to handle existing content with the old class name seems like the way to go here. Probably better than the javascript solution like I suggested earlier as in CSS it's less likely to stop working as intended :100:

comzeradd:
> > ### How to fix the existing Image blocks which are using the “blue-caption” style?
> > **Proposed solution:** Add a migration script which check all page, post, campaigns for Image blocks with blue caption style and update them with medium caption style value.
> > **The fallback solution:** make alias of “caption-style-medium” css class to “caption-style-blue-overlay”. This way if the image block having “blue-caption” style will work same as medium caption style.
> 
> The fallback solution seems better as that would also work on imported older content, and avoid the need to modify the data.

Agreed :+1: To unblock merging this we can create a `blue-caption` branch, so we can merge the Gallery fix too.

PR #381, opened 2020-09-08
PLANET-4800 Use React for GalleryCarousel

mleray:
This PR allows us to use React for the Gallery slider transitions, instead of relying on the bootstrap JS code (however I took a lot of inspiration from their code, specifically [this function](https://github.com/twbs/bootstrap/blob/main/js/src/carousel.js#L412))
This means that now the slider is fully functional in the editor too 🙂 which was requested during UAT!

### DIFF
--- a/assets/src/blocks/Gallery/GalleryCarousel.js
+++ b/assets/src/blocks/Gallery/GalleryCarousel.js
@@ -1,59 +1,128 @@
-import { Component } from '@wordpress/element';
+import { useState, useEffect, useRef } from '@wordpress/element';
 import _uniqueId from 'lodash/uniqueId';
 
 const { __ } = wp.i18n;
 
-export class GalleryCarousel extends Component {
-  constructor(props) {
-    super(props);
-    this.id = `gallery_${_uniqueId()}`;
+// This will trigger the browser to synchronously calculate the style and layout
+// You can find a list of examples here: https://gist.github.com/paulirish/5d52fb081b3570c81e3a
+const reflow = element => element.offsetHeight;
+
+export const GalleryCarousel = ({ images }) => {
+  const [currentSlide, setCurrentSlide] = useState(0);
+  const [sliding, setSliding] = useState(false);
+
+  const lastSlide = images.length - 1;
+
+  const getOrder = newSlide => {
+    let order = newSlide < currentSlide ? 'prev' : 'next';
+    if (newSlide === lastSlide && currentSlide === 0 && order !== 'prev') {
+      order = 'prev';
+    } else if (newSlide === 0 && currentSlide === lastSlide && order !== 'next') {
+      order = 'next';
+    }
+    return order;
   }
 
-  render() {
-    const { images } = this.props;
-    return (
-      <div id={this.id} className="carousel slide" data-ride="carousel">
-        {images.length > 1 &&
-          <ol className="carousel-indicators">
-            {images.map((image, index) =>
-              <li key={`#${this.id}-${index}`} data-target={`#${this.id}`} data-slide-to={index} className={index === 0 ? 'active' : ''} />
-            )}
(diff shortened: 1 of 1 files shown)


assets/src/blocks/Gallery/GalleryCarousel.js
```diff
@@ -1,59 +1,124 @@
...
+    if (newSlide !== currentSlide && !sliding) {
+      setSliding(true);
+      const order = getOrder(newSlide);
+      const direction = order === 'next' ? 'left' : 'right';
+      const nextElement = document.getElementById(`${idRef.current}-slide${newSlide}`);
+      const activeElement = document.getElementById(`${idRef.current}-slide${currentSlide}`);
+      const orderClassname = `carousel-item-${order}`;
+      const directionClassname = `carousel-item-${direction}`;
+
+      nextElement.classList.add(orderClassname);
```
lithrel, line 34: I have a `nextElement does not exist` or something like that when I insert a brand new gallery, before hitting "Insert gallery" but after selecting the images (it probably creates the gallery view before having a list of images ?), which stops interactivity of the gallery; fixed it with this diff, don't know if there's a better solution :man_shrugging: 
```diff
diff --git a/assets/src/blocks/Gallery/GalleryCarousel.js b/assets/src/blocks/Gallery/GalleryCarousel.js
index a55dc7c..b7ed24f 100644
--- a/assets/src/blocks/Gallery/GalleryCarousel.js
+++ b/assets/src/blocks/Gallery/GalleryCarousel.js
@@ -31,16 +31,16 @@ export const GalleryCarousel = ({ images }) => {
       const orderClassname = `carousel-item-${order}`;
       const directionClassname = `carousel-item-${direction}`;
 
-      nextElement.classList.add(orderClassname);
+      nextElement?.classList.add(orderClassname);
 
-      reflow(nextElement);
+      nextElement ? reflow(nextElement) : null;
 
-      activeElement.classList.add(directionClassname);
-      nextElement.classList.add(directionClassname);
+      activeElement?.classList.add(directionClassname);
+      nextElement?.classList.add(directionClassname);
 
       setTimeout(() => {
-        nextElement.classList.remove(directionClassname, orderClassname);
-        activeElement.classList.remove(orderClassname, directionClassname);
+        nextElement?.classList.remove(directionClassname, orderClassname);
+        activeElement?.classList.remove(orderClassname, directionClassname);
         setSliding(false);
         setCurrentSlide(newSlide);
       }, 600);

```

assets/src/blocks/Gallery/GalleryCarousel.js
```diff
@@ -1,59 +1,124 @@
...
+import { useState, useEffect, useRef } from '@wordpress/element';
 import _uniqueId from 'lodash/uniqueId';
 
 const { __ } = wp.i18n;
 
-export class GalleryCarousel extends Component {
-  constructor(props) {
-    super(props);
-    this.id = `gallery_${_uniqueId()}`;
+const reflow = element => element.offsetHeight;
```
lithrel, line 6: I would document the reflow function in here or in the PR description (maybe just a link to some doc), because it was not obvious when we looked at it :grimacing:

lithrel (changes requested):
Just a small bug to fix and it's good :)

assets/src/blocks/Gallery/GalleryCarousel.js
mleray: Good catch! I've pushed a fix for it, a bit different from yours, that I think is a bit easier to read 😉 I've added a check in the `useEffect` for autoplay to apply it only if `images.length` is more than 1. And to be extra safe I also check in `goToSlide` that `nextElement` and `activeElement` are not null!

lithrel (approved):
:+1: :tada:

PR #382, opened 2020-09-10
PLANET-5229 Hashtags Accessibility

dpivo:
added span with aria-label for hashtag accessibility in blocks

Fixes greenpeace/planet4#71


lithrel (approved):
:+1: (tested with voiceover :man_dancing:  )

PR #390, opened 2020-09-21
PLANET-5493 Fix articles block manual override JS test

sagarsdeshmukh:
Related to [JIRA #5493](https://jira.greenpeace.org/browse/PLANET-5493)

### DIFF
--- a/tests/js/__snapshots__/articles.editor.spec.js.snap
+++ b/tests/js/__snapshots__/articles.editor.spec.js.snap
@@ -6,12 +6,6 @@ exports[`Articles block is inserted into the Editor 1`] = `
 <!-- /wp:planet4-blocks/articles -->"
 `;
 
-exports[`Articles block should use only the manually chosen posts if any are entered 1`] = `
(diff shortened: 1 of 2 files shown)


lithrel (approved):
:+1: 
(php 7.2 test failed for some unrelated http response failure :man_shrugging: should pass on next run)

sagarsdeshmukh:
> 👍
> (php 7.2 test failed for some unrelated http response failure 🤷‍♂️ should pass on next run)

Hi @lithrel , Thanks for the review :) , I update the PR after your last review. Earlier the test was depends on default content post data, now we create required posts(needed to test manual override ) within test execution.
cc: @Inwerpsel

lithrel:
Still good to me :+1: 
Side question, because article titles inserted are not very specific, if other tests run in the same instance and create a post with similar title, will it mess up the test ? If it's the case I would prefix the title with the test name or something recognizable.

sagarsdeshmukh:
> Still good to me 👍
> Side question, because article titles inserted are not very specific, if other tests run in the same instance and create a post with similar title, will it mess up the test ? If it's the case I would prefix the title with the test name or something recognizable.

The post/page with same name will not affect on test result, but its a good suggestion to keep test name in post title. 👍

PR #391, opened 2020-09-22
PLANET-5502 Vertical align happypoint block iframe

comzeradd:
- Remove top padding from the wrapper
- Vertical align the iframe by placing it at 50% and using a negative top margin


mleray (approved):
Looks good to me! 👍

comzeradd:
Confirmed with Design. Just added a minor height increase on mobile.

PR #392, opened 2020-09-22
PLANET-5506 Fix articles block date translation in backend

sagarsdeshmukh:
The articles block date translation was failing in block editor because of `window.p4bk_vars.dateFormat` var not found in backend.


comzeradd (approved):
Tested locally and on the test instance, looks good :+1: 

<s>Can you do a minor change in the commit message, just to add the `PLANET-5506` prefix?</s>

PR #393, opened 2020-09-23
Add option for updating the counters in the client

oekeur:
For now this adds an eventListener to the counter which in turn executes the calculateRemaining.
This way GPNL (and other NRO's with devs) can use instantly updating counters.
Which might serve as a Proof of Concept for adding in more interactivity.

I will follow up on this PR with another which separates the functions for calling the API and calculating remaining. This improves the code a bit on single responsibility and make sure the instantly updating counter don't hit API's unneeded.

### DIFF
--- a/assets/src/blocks/Counter/CounterFrontend.js
+++ b/assets/src/blocks/Counter/CounterFrontend.js
@@ -20,6 +20,11 @@ export class CounterFrontend extends Component {
     window.addEventListener('updateCounter', function () { counter.calculateRemaining(); }, false);
   }
 
+  componentWillUnmount() {
+    let counter = this;
+    window.removeEventListener('updateCounter', function () { counter.calculateRemaining(); }, false);
+  }
+
   componentDidUpdate({ target: prevTarget, completed: prevCompleted, completed_api: prevCompletedApi }) {
     // Update completed and remaining values depending on props
     const { target, completed, completed_api } = this.props;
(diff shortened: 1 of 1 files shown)


assets/src/blocks/Counter/CounterFrontend.js
```diff
@@ -14,7 +14,9 @@ export class CounterFrontend extends Component {
 
   componentDidMount() {
     // Calculate completed and remaining values depending on props
-    this.calculateRemaining();
+    let counter = this;
+    counter.calculateRemaining();
+    window.addEventListener('updateCounter', function (e) { counter.calculateRemaining(); }, false);
```
mleray, line 19: The `e` param is unused, it can probably be removed? 🤔 Maybe it would also be handy to add a comment to explain what is happening here 🙂

oekeur: I will remove the parameter for now :)
I am going to need and re-add it later though, because through the e.detail parameter it's possible to pass the new unique_count and that the API doesn't need to be called twice (update&read).

Will add a comment :)

oekeur: Done :)

mleray (commented):
Maybe we should check if `completed_api` is set in the props before adding the listener? To avoid useless calls... Also, shouldn't we remove the event listener when unmounting the component? 🤔

oekeur:
> Maybe we should check if `completed_api` is set in the props before adding the listener? To avoid useless calls... Also, shouldn't we remove the event listener when unmounting the component?

I have incorporated your suggestions in the latest commits. :)

mleray (approved):
Looks good to me 👍

PR #395, opened 2020-09-23
PLANET-5500 Columns Block: Use full width on buttons

comzeradd:
Also removed text-decoration, because buttons already inherit that from the styleguide.

Ref: https://jira.greenpeace.org/browse/PLANET-5500


sagarsdeshmukh (approved):
LGTM 👍

PR #396, opened 2020-09-25
PLANET-5374 Update Columns header color

mleray:
See https://jira.greenpeace.org/browse/PLANET-5374

I've added a CSS variable for the Columns header color, @Inwerpsel please let me know if the name fits the rules 🙂

### DIFF
--- a/assets/src/styles/blocks/Columns.scss
+++ b/assets/src/styles/blocks/Columns.scss
@@ -16,6 +16,7 @@
     font-size: $font-size-lg;
     font-family: $roboto;
     font-weight: 700;
+    color: var(--block-columns--column-heading--color, $grey-80);
 
     @include medium-and-up {
       font-size: 1.5rem;


assets/src/styles/blocks/Columns.scss
```diff
@@ -16,6 +16,7 @@
     font-size: $font-size-lg;
     font-family: $roboto;
     font-weight: 700;
+    color: var(--columns-header--color, $grey-80);
```
Inwerpsel, line 19: I also switch these up all of the time, should be `heading` instead :smile:

The format is good, but maybe we should think about how to handle variables that apply to a block. If we consistently use a prefix for each block, then we could automatically extract all variables for a given block. Do you think it would make sense to separate this prefix too with 2 dashes? So that the resulting variable would be 
```css
color: var(--block-columns--column-heading--color, $grey-80);
```

Inwerpsel: The result is a bit long, but that wouldn't be an issue when we start using shorthand notation.
```css
--block-columns--column-heading-- {
  color: $grey-80;
}
```

mleray: Updated!

Inwerpsel (approved):
Looks good :+1:

PR #397, opened 2020-09-25
Feature/accordion block

elncetoo:


elncetoo:
Two things are not quite okey - the read-more button`s href and the multiple collapse( I left commented out some things that worked partially, but with a bit of issue)

assets/src/blocks/Accordion/AccordionBlock.js
```diff
@@ -0,0 +1,139 @@
...
+        {
+          attributes,
+          save() {
+            return null;
+          },
+        }
+      ],
+      edit: ( { isSelected, attributes, setAttributes } ) => {
+
+        function addAccRow(){
```
mleray, line 85: We could define this function (and also the following `removeAccRow`) in `AccordionEditor` file instead of here, so that we can write directly `edit: AccordionEditor` at line 83, which is much shorter 🙂

assets/src/blocks/Accordion/AccordionBlock.js
```diff
@@ -0,0 +1,139 @@
...
+      // },
+      // css_variables: CSS_VARIABLES_ATTRIBUTE,
+    };
+
+    registerBlockType( BLOCK_NAME, {
+      title: __( 'Accordion', 'planet4-blocks-backend' ),
+      icon: 'menu',
+      category: 'planet4-blocks-beta',
+      keywords: [
+        __(this.BLOCK_NAME),
```
mleray, line 71: I don't think `this` is needed here? 🤔

assets/src/blocks/Accordion/AccordionBlock.js
```diff
@@ -0,0 +1,139 @@
+/* eslint-disable no-unused-vars */
+import { AccordionEditor } from './AccordionEditor';
+import { frontendRendered } from '../frontendRendered';
+import { CSS_VARIABLES_ATTRIBUTE } from '../CssVariablesAttribute';
```
mleray, line 4: This import is not used, can be removed 🙂

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+              <RichText
+                tagName="p"
+                className="page-section-description"
+                placeholder={__('Enter the description of the block or leave empty to hide it on the front page.', 'planet4-blocks-backend')}
+                value={attributes.accordion_description}
+                onChange={this.toAttribute('accordion_description')}
+                keepPlaceholderOnFocus={true}
+                withoutInteractiveFormatting
+                characterLimit={400}
+                multiline="false"
```
mleray, line 212: Description should allow multiline, if we want to be consistent with other blocks!

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+    // accordion_rows = this.getAccordionRows(accordion_rows);
+
+    return <Fragment>
+        <section>
+          <div className="block accordion-block my-3">
+            <header>
+              <RichText
+                tagName="h2"
+                className="page-section-header mt-3"
+                placeholder={__('Enter title for the block or leave empty to hide it.', 'planet4-blocks-backend')}
```
mleray, line 195: Other WYSIWYG blocks just use `Enter title` as placeholder (for example [Articles block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesEditor.js#L112)) so we should probably use the same here, for consistency? Also, character limit is usually less for titles, from a quick search in our existing blocks I've seen either 40 or 60 (this should probably be discussed with design actually, ideally all blocks should have the same value!)

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+                onChange={this.toAttribute('accordion_title')}
+                keepPlaceholderOnFocus={true}
+                withoutInteractiveFormatting
+                characterLimit={100}
+                multiline="false"
+              />
+              <RichText
+                tagName="p"
+                className="page-section-description"
+                placeholder={__('Enter the description of the block or leave empty to hide it on the front page.', 'planet4-blocks-backend')}
```
mleray, line 206: Same here, other WYSIWYG blocks just use `Enter description` as placeholder (for example [Articles block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesEditor.js#L125)) so we should probably use the same here, for consistency? Here again, in terms of character limit from a quick search in our existing blocks I've seen a lot of different values from 200 to 400 so this should probably be discussed with design...

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+                value={attributes.accordion_description}
+                onChange={this.toAttribute('accordion_description')}
+                keepPlaceholderOnFocus={true}
+                withoutInteractiveFormatting
+                characterLimit={400}
+                multiline="false"
+              />
+            </header>
+
+            {[accordion_rows].map((accordion_row, index) => {
```
mleray, line 216: Isn't `accordion_rows` already an array? 🤔 Why is it between `[]`?

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+        })}
+      */}
+      </Fragment>
+    );
+  }
+
+
+//renders the editor view
+  renderView() {
+    const { attributes, className } = this.props;
```
mleray, line 182: I don't think that this `className` is used anywhere

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+              />
+            </header>
+
+            {[accordion_rows].map((accordion_row, index) => {
+              return (
+                <div key={index}>
+                  <div className="accordion-content my-2">
+                    <div className="accordion card-header"
+                        onClick={this.handleCollapseClick}
+                        id={attributes.accordion_id, index+1}
```
mleray, line 222: Is it wanted here to have two ids? From element inspection it seems only the first one is applied 🤔

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+                      <RichText
+                        // id={attributes.accordion_id, index+1}
+                        tagName="p"
+                        className="accordion-text"
+                        placeholder={__('Enter description here', 'planet4-blocks-backend')}
+                        value={attributes.accordion_text}
+                        onChange={this.toAttribute('accordion_text')}
+                        keepPlaceholderOnFocus={true}
+                      />
+                    <Tooltip text={__('Add button text or leave empty to hide it on the front page.', 'planet4-blocks-backend')}>
```
mleray, line 247: Again, here if we want to be consistent with the [Articles block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesEditor.js#L134) we probably need to change the copy to `Edit text` for the tooltip and `Enter text` for the input placeholder... Actually we probably don't even need the tooltip here, we added it to Articles block only because there was a default value but here that's not the case!

assets/src/blocks/Accordion/AccordionFrontend.js
```diff
@@ -0,0 +1,183 @@
...
+      //           // panel.style.visibility === "hidden" &&  panel.style.opacity === "0"
+      //           // ? panel.style.visibility === "visible" &&  panel.style.opacity === "1"
+      //           // : panel.style.visibility === "hidden" &&  panel.style.opacity === "0";
+      //   });
+      // }
+
+
+    return (
+
+      <Fragment>
```
mleray, line 145: This `Fragment` is not needed since it only has one child 🙂

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+                className="page-section-description"
+                placeholder={__('Enter the description of the block or leave empty to hide it on the front page.', 'planet4-blocks-backend')}
+                value={attributes.accordion_description}
+                onChange={this.toAttribute('accordion_description')}
+                keepPlaceholderOnFocus={true}
+                withoutInteractiveFormatting
+                characterLimit={400}
+                multiline="false"
+              />
+            </header>
```
mleray, line 214: The `header` tag in other blocks is only around the title, and not the description, we probably want to keep this markup 🙂

assets/src/blocks/Accordion/AccordionFrontend.js
```diff
@@ -0,0 +1,183 @@
...
+        <section className="block accordion-block my-0 py-0">
+            <header>
```
mleray, line 147: Same as in the editor, the `header` tag in other blocks is only around the title, and not the description, we probably want to keep this markup 🙂

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,314 @@
...
+                  >
+                    {__('Remove Row', 'planet4-blocks-backend')} <Dashicon icon={"minus"} size={12} />
+                  </button>
+                </Tooltip>
+              )}
+            </div> */}
+          </div>
+        </section>
+
+        <AccordionFrontend isEditing {...attributes} />
```
mleray, line 295: From a quick look at `AccordionFrontend` it looks like pretty much nothing is rendered if `isEditing` is true, so we probably don't need to add it here...? 🤔

classes/blocks/class-accordion.php
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * Accordion constructor.
+	 */
+	public function __construct() {
+		register_block_type(
+			'planet4-blocks-beta/accordion',
+			[
+				'editor_script'   => 'planet4-blocks-beta',
+				'render_callback' => [ $this, 'render' ],
```
mleray, line 34: I don't think that we need this `render_callback`, since this is a brand new block 🤔

mleray (commented):
Looks like a really good start! 👍 I've put some comments, most of them about consistency across WYSIWYG blocks 🙂 Also, I'm not sure whether or not the changes to `composer.json` and `package.json` should be committed...?

assets/src/blocks/Accordion/AccordionBlock.js
elncetoo: right.. I am removing it, I used it in the beginning for the multiple style options of the block :)

assets/src/blocks/Accordion/AccordionBlock.js
elncetoo: I cleaned the code more and this could actually go in general. Currently as it works if block title & description are left empty they are hidden from the front end, so a person can add as much as wanted collapsible elements without the need of an array to hold them all. (and still looks on the frontend as if they are all one block)

assets/src/blocks/Accordion/AccordionEditor.js
elncetoo: it isn't any more indeed (was used for the multiple style options), removing it

assets/src/blocks/Accordion/AccordionEditor.js
elncetoo: setting it now for 200 till it's discussed with design..

assets/src/blocks/Accordion/AccordionEditor.js
elncetoo: Sure I fixed the placeholder text;
The tooltip I added for the button so that it's clear that if there is no text the button won't show up on the front end. I understood that this is optional, also in our use case for the Nordics ( that I added accordion block as a custom html block atm) - we use this block for FAQ and we don't use button anywhere to be honest.. I think this "optionality" for the button needs to be discussed further .. :)

elncetoo:
One thing is left to be debugged on the editor`s side all items open on the state toggle event..

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -0,0 +1,170 @@
...
+        <RichText
+          tagName="p"
+          className="page-section-description"
+          placeholder={__('Enter description', 'planet4-blocks-backend')}
+          value={attributes.accordion_description}
+          onChange={this.toAttribute('accordion_description')}
+          keepPlaceholderOnFocus={true}
+          withoutInteractiveFormatting
+          characterLimit={200}
+          multiline="true"
```
mleray, line 110: I still can't enter multiline text for the description 🤔 I think you can just remove this line completely and it should work!

assets/src/blocks/Accordion/AccordionBlock.js
```diff
@@ -0,0 +1,88 @@
...
+      attributes,
+      deprecated: [
+        {
+          attributes,
+          save () {
+            return null;
+          }
+        }
+      ],
+      edit: ({ isSelected, attributes, setAttributes }) => {
```
mleray, line 78: This can now be shortened to `edit: AccordionEditor` 🙂

assets/src/blocks/Accordion/AccordionEditor.js
Inwerpsel: Not sure if that's causing the issue, but with a boolean prop you can just do
```js
<RichText
// ...
  multline>

mleray: Yeah that should also work 🙂 But I think it's the default value anyway so it can just be removed!

mleray:
I opened a new PR to explore using a `tabs` property instead of having multiple Accordion blocks in a row: https://github.com/elncetoo/planet4-plugin-gutenberg-blocks/pull/1/files

mleray:
This was solved with another PR 🙂

PR #399, opened 2020-09-29
PLANET-5497 Remove Image block caption style picker

mleray:
See https://jira.greenpeace.org/browse/PLANET-5497

The ticket specified to use CSS to hide the picker, but I think that this JS solution makes more sense 🙂 (CSS selector would be quite complicated, or we would need to add a new class just for this purpose)

### DIFF
--- a/assets/src/ImageBlockExtension.js
+++ b/assets/src/ImageBlockExtension.js
@@ -1,4 +1,4 @@
-import { Button, ButtonGroup, PanelBody, SelectControl } from '@wordpress/components';
+import { Button, ButtonGroup, PanelBody } from '@wordpress/components';
 import assign from 'lodash.assign';
 
 const { addFilter } = wp.hooks;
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (approved):
Tested on local, LGTM 👍

assets/src/ImageBlockExtension.js
```diff
@@ -95,7 +95,7 @@ const addExtraControls = function() {
               title={ __( 'Planet4 Image Options' ) }
               initialOpen={ true }
             >
-              <SelectControl
+              {/* <SelectControl
```
comzeradd, line 98: Maybe just remove this block completely, instead of commenting it.

mleray: I thought we didn't want to completely remove it, in case we need to put it back? 🤔 (hence the initial idea to hide it with CSS)

comzeradd: yes, indeed. but it will be part of the git log, if we ever want to re-add it.

mleray: Okay, sure! Removed 🙂

PR #400, opened 2020-10-01
PLANET-5468 Columns block: Huge image on the 4th column

dpivo:
Added word break for columns description so long words don't cause the column to go oversize.

### DIFF
--- a/assets/src/styles/blocks/Columns.scss
+++ b/assets/src/styles/blocks/Columns.scss
@@ -492,4 +492,8 @@
       }
     }
   }
+
+  .column-wrap {
+    word-break: break-word;
+  }
 }


mleray (changes requested):
I've tried applying this change to [the page given as example in the ticket](https://www.greenpeace.org/luxembourg/de/sich-informieren/greenpeace/team/) and it didn't fix the issue... What seemed to work however was to apply `word-break: break-word;` directly to `.column-wrap`, but I'm wondering if that might have side effects 🤔 Also, I think that `break-word` is better than `break-all` because it will keep words whole as much as possible instead of randomly cutting them at any character 🙂

dpivo:
> I've tried applying this change to [the page given as example in the ticket](https://www.greenpeace.org/luxembourg/de/sich-informieren/greenpeace/team/) and it didn't fix the issue... What seemed to work however was to apply `word-break: break-word;` directly to `.column-wrap`, but I'm wondering if that might have side effects thinking Also, I think that `break-word` is better than `break-all` because it will keep words whole as much as possible instead of randomly cutting them at any character slightly_smiling_face

Good point with the 'break-word' vs 'break-all.' For it not working on the test page however, for me it seems to work, I pinned the changes to a test instance and created a block with the same words as the page with the issue: https://k8s.p4.greenpeace.org/test-oberon/?page_id=975&preview=true

mleray (approved):
Looks good to me 👍

PR #401, opened 2020-10-06
PLANET-5562: Fix htmlentity character on split two columns titles

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5562

> An & sign is displayed as its htmlentities on the frontend version of the Split 2 Columns block, for the title (left column) and the tag (right column) of the block.

The `&` sign is automatically escaped by the RichText component on input, and saved as an html entity. To display it properly, we have to allow some html entities to be displayed as is.

## Fix

Allows titles to display html characters, the same way it is allowed in text blocks.  
Input is still limited by `allowedFormats`, this fix is only here to compensate for the automatic escaping of `&`.

## Test

- Add a new page with a Split Two Columns block.
- Add a title to this block with an `&` (ie `Foo & Bar`)
- Clicking outside the block should trigger the frontend view, 
  - On the current version, it displays a title `Foo &amp; Bar`
  - On the fixed version, it displays the proper title `Foo & Bar`

### DIFF
--- a/assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
+++ b/assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
@@ -44,7 +44,9 @@ export const SplittwocolumnsFrontend = ({
         <div className="split-two-column-item-content">
           {title && issue_link_path &&
             <h2 className="split-two-column-item-title">
-              <a href={issue_link_path} {...analytics('Category Title')}>{title}</a>
+              <a href={issue_link_path} {...analytics('Category Title')}
+                dangerouslySetInnerHTML={{__html: title}} 
+              />
             </h2>
           }
           {title && !issue_link_path &&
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh:
The fix is working 👍 
We also need the fix for button text and issue link text.
image

lithrel:
> We also need the fix for button text and issue link text.

:grimacing: Good point :+1: 
@sagarsdeshmukh Just pushed an additional fix for those

mleray:
> We also need the fix for button text and issue link text.
Could this be a problem for other blocks too? 🤔 I'm thinking in particular about the Articles block that can have [custom button text](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlesFrontend.js#L46)

lithrel:
@mleray  I think there's another way to do this, which is to use RichText component (with readonly attribute ?) instead of just html tags, it seems to handle its own escape stuff properly; apparently ArticlesBlock uses RichText for renderView(), so it's not a problem there.

sagarsdeshmukh (approved):
LGTM 👍

PR #409, opened 2020-10-13
PLANET-5341: Firebase + ControlShift PoC

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5341

Dependencies:
- Firebase https://www.npmjs.com/package/firebase
- React Firebase Hooks https://www.npmjs.com/package/react-firebase-hooks

[`CSFrequency/react-firebase-hooks`](https://github.com/CSFrequency/react-firebase-hooks) makes it quite easy to work with firebase inside React blocks, with hooks like `useCollection()`, `useDocument()` or `useCollectionOnce()` ([see doc.](https://github.com/CSFrequency/react-firebase-hooks/blob/master/firestore/README.md)).

## Demo

The demo here adds a new block **ControlShift demo** usable in the page editor (no save function, won't appear in preview), that allows to gather petition data and events data.

cs-demo

Block type `Petition` displays a petition selector with petitions gathered from the firestore collection `p4-test-petition`, and selecting one of those will display all available data. Data is only fetch once with `useCollectionOnce()`.

Block type `Events` displays all events available in the collection `p4-test-event`; it is fetched using `useCollection()` and will be updated live if more data is added to the collection (using the [firestore console](https://console.firebase.google.com/project/greenpeace-testing/firestore/data~2Fp4-test-event))

## Access

I added an app [`p4-web`](https://console.firebase.google.com/project/greenpeace-testing/settings/general/web:MzIyMDg1ODQtMmQyYS00YmEzLWE2NGUtZTE5NTdiNjVmNTFi) to the firebase project, to get api keys to access the datastore.

## Security issues

API keys are public, security has to be defined by rules in Firestore.  
Rules are easy to mess up: 
>  In the case where multiple allow expressions match a request, the access is allowed if any of the conditions is true  

which means if anybody writes a rule too permissive, all the rules are broken.  
There's a [rules playground](https://console.firebase.google.com/project/greenpeace-testing/firestore/rules) to test differents rules before applying them.

Screenshot from 2020-10-13 15-37-28

## Development

We need a dev store separated from production (with dev config and prod config in files) so that we don't break data and rules too easily.

### DIFF
--- a/assets/src/blocks/ControlShift/ControlShiftBlock.js
+++ b/assets/src/blocks/ControlShift/ControlShiftBlock.js
@@ -1,0 +1,23 @@
+import { InspectorControls } from '@wordpress/block-editor';
+
+import { ControlShiftSelector } from './ControlShiftSelector';
+
+const { registerBlockType } = wp.blocks;
+const { __ } = wp.i18n;
+
+const BLOCK_NAME = "planet4-blocks/control-shift-demo";
+
+export const ControlShiftBlock = () => {
+  return (
+    <ControlShiftSelector />
+  )
+}
+
+registerBlockType( BLOCK_NAME, {
+  title: __('ControlShift demo', 'planet4-blocks-backend'),
+  icon: 'editor-table',
+  category: 'planet4-blocks',
+  attributes: {},
+  edit: ControlShiftBlock,
+  save: () => null
+});
--- a/assets/src/blocks/ControlShift/ControlShiftEvents.js
+++ b/assets/src/blocks/ControlShift/ControlShiftEvents.js
@@ -1,0 +1,24 @@
+import { useState } from '@wordpress/element';
+import { useCollection } from 'react-firebase-hooks/firestore';
+import firebase from './firebase';
+
+export const ControlShiftEvents = () => {
+  const [eventsCollection, loading, error] = useCollection(
+    firebase.firestore().collection('p4-test-event'), {}
+  );
+
+  const events = eventsCollection ? eventsCollection.docs.map((e) => {
+    return <li key={e.id}>[{e.data().type}] {e.data().jid}</li>
+  }) : [];
+
+  return (
+    <div>
+      <h2>Events</h2>
+      {error && <p><strong>Error: {JSON.stringify(error)}</strong></p>}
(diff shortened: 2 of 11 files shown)


assets/src/blocks/ControlShift/ControlShiftEvents.js
```diff
@@ -0,0 +1,24 @@
...
+import { useCollection } from 'react-firebase-hooks/firestore';
+import firebase from './firebase';
+
+export const ControlShiftEvents = () => {
+  const [eventsCollection, loading, error] = useCollection(
+    firebase.firestore().collection('p4-test-event'), {}
+  );
+
+  const events = eventsCollection ? eventsCollection.docs.map((e) => {
+    return <li key={e.id}>[{e.data().type}] {e.data().jid}</li>
```
pablocubico, line 11: Just a note for the future, this collection seems to be mapping the Webhook events endpoints (e.g.: https://developers.controlshiftlabs.com/#webhook-endpoints-event-updated) 

I think it should map:
https://developers.controlshiftlabs.com/#authenticated-rest-api-events

Anyway, for the purpose of this PoC, it's irrelevant.

lithrel: Yeah I realized that later :grimacing: What I don't understand is there's no public api endpoint for events, like there is for petitions :shrug: This would force us to connect to the authenticated api of CSL, or pay attention to what we save in Firestore so that there's no access to private/unpublished information.

assets/src/blocks/ControlShift/ControlShiftBlock.js
```diff
@@ -0,0 +1,23 @@
...
+export const ControlShiftBlock = () => {
+  return (
+    <ControlShiftSelector />
+  )
+}
+
+registerBlockType( BLOCK_NAME, {
+  title: __('ControlShift demo', 'planet4-blocks-backend'),
+  icon: 'editor-table',
+  category: 'planet4-blocks',
```
sagarsdeshmukh, line 19: We could use `planet4-blocks-beta` category here.

comzeradd:
Closing this, but keeping the branch for now in case we need to get back to this.

PR #410, opened 2020-10-13
Add plugin for shorthand css variable syntax

Inwerpsel:
Relevant ADR: https://docs.google.com/document/d/10r5RM1h8tN6Dw8APCJCM7GeRYOaRIOc8QMm2t13AwT8/edit# (advised to read the "Challenges" section, it describes the limitations of SCSS for this use case).

Add webpack plugin that runs after SCSS is transformed to CSS, to transform the shorthand syntax into the corresponding declarations that include variables.

So it should transform any sub-selector that in SCSS starts and ends with 2 dashes into variable declarations that are added to the parent selector. So
```scss
#some .parent[selector] {
  border: 2px;
  --some-meaningful-name-- {
    color: red;
    background-color: green;
  }
}
```
into
```css
#some .parent[selector] {
  border: 2px;
  color: var(--some-meaningful-name--color, red);
  background-color: var(--some-meaningful-name--background-color, green);
}
```

I'm working on applying this shorthand in all places where it requires no other changes to do so.

For some variables the naming needs to change. I'm checking using github search across repo code to see if any variable that needs to be renamed is used in any other repo. None so far were, if I find any I'll exclude those for now.

For that search I'm using this Intellij plugin https://plugins.jetbrains.com/plugin/12678-onlinesearch2 that allows to add custom search URLs similar to "Search with Google" to perform on selected text from the context menu. Very useful in our case as we frequently need to check if a change would break child themes.

### DIFF
--- a/assets/src/styles/blocks/Covers/TakeActionCovers.scss
+++ b/assets/src/styles/blocks/Covers/TakeActionCovers.scss
@@ -146,12 +146,17 @@
 
     .btn {
       display: block;
-      background-color: var(--cover-card-btn-background-color, $orange);
-      border-color: var(--cover-card-btn-border-color, $orange);
+
+      --block-covers--card-btn-- {
+        background-color: $orange;
+        border-color: $orange;
+      }
 
       &:hover {
-        background-color: var(--cover-card-btn-hover-background-color, $orange-hover);
-        border-color: var(--cover-card-btn-hover-border-color, $orange-hover);
+        --block-covers--card-btn-hover-- {
+          background-color: $orange-hover;
+          border-color: $orange-hover;
+        }
       }
     }
   }
@@ -190,7 +195,11 @@
     padding-top: 0;
     max-width: 80%;
     transition: color 100ms linear;
-    color: var(--cover-card-heading-color, $white);
+
+    --block-covers--heading-- {
+      color: $white;
+    }
+
     text-shadow: 1px 1px 3px $black;
     display: table;
     z-index: 1;
@@ -225,7 +234,11 @@
 }
 
 .cover-card-tag {
-  color: var(--cover-card-tag-color, $yellow);
+
+  --block-covers--card-tag-- {
+    color: $yellow;
+  }
+
   display: inline-block;
   margin-bottom: 8px;
   text-decoration: none;
@@ -237,7 +250,10 @@
 
   &:hover {
     text-decoration: underline;
-    color: var(--cover-card-tag-hover-color, $yellow);
+
+    --block-covers--card-tag-hover-- {
+      color: $yellow;
+    }
   }
 
   @include large-and-up {
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -64,7 +64,9 @@
     "selector-pseudo-class-case": "lower",
     "selector-pseudo-element-case": "lower",
     "selector-type-case": "lower",
-    "selector-type-no-unknown": true,
+    "selector-type-no-unknown": [true, {
+	  "ignoreTypes": ["/^--[a-z]\\w*(--?[a-z0-9]+)*--$/"]
+	}],
     "selector-max-empty-lines": 0,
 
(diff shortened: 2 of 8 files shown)


Inwerpsel:
Discussed this a bit already with @pablocubico on Slack, and found the `ignoreTypes` option of `selector-type-no-unknown` which makes our CI take peace with the syntax, and also seems to be recognized in both PHPStorm and VSCode, provided those IDEs are set up to use the `.stylelintrc` in the repo.

I think we're agreeing on using the `--` prefix on the selector, as it corresponds to how the variable will look in the transformed CSS. We're not sure about suffixing with `--` as well. There's some pros and cons that I can think of, but feel free to add more:
**pro**:
- Easier to differentiate at a glance from regular selectors and custom property declarations.
```
.btn-primary {
  .sub-selector {
    border-size: 1px;
  }
  --primary-button-- {
    color: green;
  }
  --some-prop: blue;
}
```
vs
.btn-primary {
  .sub-selector {
    border-size: 1px;
  }
  --primary-button {
    color: green;
  }
  --some-prop: blue;
}
- These dashes will be there in the final variable names too to separate the prefix from the property names (`var(--some-meaningful-name**--**background-color, green)`). So a bit more recognizable and explicit.
**con**:
- More tedious to type.

lithrel:
> We're not sure about suffixing with `--` as well. There's some pros and cons that I can think of, but feel free to add more

I like the version without suffix better, but just because it looks prettier to me, nothing technical :) I don't feel like one version is easier to differentiate quickly than the other, it's all dashes to my old eyes, something with a keyword and different characters like `--var(some-meaningful-name) {}` might be more visible, but in the end it's syntax highlighting that would help the most; so I'm fine with any solution :ok_hand:

Inwerpsel:
I just found out that Gutenberg is working on something eerily similar to what we want to do with campaign themes :open_mouth: 
https://developer.wordpress.org/block-editor/developers/themes/theme-json/#css-custom-properties

It looks like a very recent addition to the docs, and wasn't aware of it until I stumbled on this line https://github.com/WordPress/gutenberg/blob/520957775fa7ac8f4da0a219cb76e85b1f63770e/packages/blocks/src/api/constants.js#L16-L16

Mentioning it here because they have chosen pretty much the same naming scheme for the variables, where different parts are separated with 2 dashes, just like this PR does with block prefixes. Interestingly in the line mentioned above they have matched their name with the UI grouping the variable will be in. Maybe with the idea of eventually extracting the UI entirely from the variable names, without needing a separate config file?

But it's the only such variable, so their format will probably still change.

sagarsdeshmukh:
> We're not sure about suffixing with -- as well. There's some pros and cons that I can think of, but feel free to add more:

+1 for 
> Easier to differentiate at a glance from regular selectors and custom property declarations.

comzeradd:
Both options are equal technically, but visually I agree the `--` suffix is slightly better.

assets/src/styles/blocks/Covers/TakeActionCovers.scss
```diff
@@ -146,12 +146,17 @@
 
     .btn {
       display: block;
-      background-color: var(--cover-card-btn-background-color, $orange);
-      border-color: var(--cover-card-btn-border-color, $orange);
+
+      --block-covers--card-btn-- {
```
Inwerpsel, line 150: I changed the name of this variable (and some others), but only after checking that there is no usage of it among any other repo of the Greenpeace Github organisation. (e.g. https://github.com/search?q=org%3Agreenpeace+var(--cover-card-heading-color&type=code)

Inwerpsel, line 150: Applying the syntax here meant changing the variable name. Did this in a few places, but only after checking that the previous name wasn't used anywhere else (e.g. https://github.com/search?q=org%3Agreenpeace+%22--cover-card-btn-border-color%22&type=code).

lithrel (approved):
:+1:

PR #412, opened 2020-10-14
PLANET-5388 Fix related articles issue on Post

sagarsdeshmukh:
[JIRA 5388](https://jira.greenpeace.org/browse/PLANET-5388)

Issue -
- The articles block on Post page not showing the current page tags and categories related articles

Before fix - https://www.greenpeace.org/international/story/44543/fading-memories-of-atomic-bombing-hiroshima-75-years-later/
After fix    - https://k8s.p4.greenpeace.org/test-jupiter/story/44543/fading-memories-of-atomic-bombing-hiroshima-75-years-later/
(Please check the articles block at the bottom of page, The articles display in block should belongs to current post tags and categories)

### DIFF
--- a/assets/src/blocks/Articles/ArticlesFrontend.js
+++ b/assets/src/blocks/Articles/ArticlesFrontend.js
@@ -18,7 +18,9 @@ export const ArticlesFrontend = (props) => {
 
   const postId = !postIdClass ? null : postIdClass.split('-')[1];
 
-  const { posts, loadNextPage, hasMorePages, loading } = useArticlesFetch(props, postType, postId, document.body.dataset.nro);
+  const postCategories = props.post_categories || [];
+
+  const { posts, loadNextPage, hasMorePages, loading } = useArticlesFetch(props, postType, postId, document.body.dataset.nro, postCategories);
 
   return (
     <section className="block articles-block">
--- a/assets/src/blocks/Articles/useArticlesFetch.js
+++ b/assets/src/blocks/Articles/useArticlesFetch.js
@@ -8,7 +8,7 @@ const fetchJson = async(url) => {
   return response.json();
 };
 
-export const useArticlesFetch = (attributes, postType, postId, baseUrl = null) => {
+export const useArticlesFetch = (attributes, postType, postId, baseUrl = null, postCategories = []) => {
   const { article_count, post_types, posts, tags, ignore_categories } = attributes;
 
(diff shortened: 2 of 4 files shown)


assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -18,7 +18,9 @@ export const ArticlesFrontend = (props) => {
 
   const postId = !postIdClass ? null : postIdClass.split('-')[1];
 
-  const { posts, loadNextPage, hasMorePages, loading } = useArticlesFetch(props, postType, postId, document.body.dataset.nro);
+  const postCategorie = props.post_categorie || [];
```
lithrel, line 21: Hey I feel a bit lost in the review, I don't see any mention of `post_categorie` in the code (and because categor~y~ie is the singular for categories but in french, I assumed someone made a small mistake somewhere else that was copied here), do you know where it comes from/where it appears ?  
Also I have a bug in `PostTypeSelector` when I try to add a post type (`this.props.postTypes.find is undefined l.17`), but I don't think it's related to this PR.

sagarsdeshmukh: Hi @lithrel , 
Sorry for the trouble, I forgot to commit the [master theme changes](https://github.com/greenpeace/planet4-master-theme/pull/1204) :( 

Earlier we had server side render function to load articles block, but now we have api endpoints, so the server side code fails at [this point](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-articles.php#L281) to fetch post categories and tags.

The workaround I added here is, pass post tags and categories as block attributes from [here](https://github.com/greenpeace/planet4-master-theme/pull/1204/files#diff-8e039a17b2aadb723e862bced701672b6e20a79e19ca6426b49084fc76d52624R87).

lithrel (approved):
I would change `post_categorie` and  `postCategorie` to plural `post_categories` `postCategories` (same in https://github.com/greenpeace/planet4-master-theme/pull/1204).

:+1:

sagarsdeshmukh:
Hi @lithrel , Thanks for the review :) , I added the test case to check articles block on post page. please have a look.

lithrel:
Still looks good to me :+1: 
CI fails because of some timeout on `apt install`, should pass if you re-run the process :man_shrugging:

PR #414, opened 2020-10-19
PLANET-5324: Implement lightbox solution for images

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-5324

This PR creates a thin React wrapper around the Photoswipe lib. Photoswipe requires certain markup (see "Step 2" in [their guide](https://photoswipe.com/documentation/getting-started.html)) to be present and then "hydrates" it (as the React folks would say) to add all the necessary JS magic.

The wrapper is a Lightbox component we can use to pass some images to it and open it on demand. It integrates with the current galleries with just a few lines of code.

As we needed some custom markup for the bottom indicators (Photoswipe has no indicators) I delegated the markup creation to the React side so we can just loop through the images array to build the indicators, and set an `active` class in one of them easily.

The lightbox uses a Portal to add the element to the document's `body` instead of the Gallery's React root. 

For image blocks, the component is rendered on the fly upon image click, it shows a lightbox with a single image. 

In the case of images, as React's `render()` needs a target element, and that target can't be `body` (because that will make React to erase everything), I create a `div` on the fly with `position: fixed`, which removes the element from the [document flow](https://developer.mozilla.org/en-US/docs/Web/CSS/position) so it causes no stress to the layout (i.e.: no gaps). This `div` ends up having nothing inside, as the Portal in Lightbox makes React append it to the document's `body`.

React's [diffing/reconciliating algorithm ](https://reactjs.org/docs/reconciliation.html)will do its best to reuse existing DOM elements and avoid unnecessary remounts and re-renders,. 

In the case of the Carousel Gallery, as the Gallery's state is updated on each slide switch, I used `memo()` to avoid re-renders of the Lightbox, which is pretty much as using `shouldComponentUpdate` but in a functional component (see: https://reactjs.org/docs/optimizing-performance.html#avoid-reconciliation).


assets/src/blocks/Gallery/GalleryGrid.js
```diff
@@ -1,5 +1,11 @@
-export const GalleryGrid = ({ images }) => (
-  <div className="container">
+import { useEffect } from '@wordpress/element';
```
Inwerpsel, line 1: Maybe this is a good opportunity to look into a better React version for the frontend. Since we only render simple things on the frontend, we could probably do with a very lightweight React implementation. The libs included in WordPress are quite the opposite, and they have a lot of dependencies which are not optimized to reduce bundle size, as they're intended for the post editor.

Inwerpsel: Screenshot from 2020-10-20 11-12-01
Screenshot from 2020-10-20 11-11-46

Inwerpsel: The only code difference between both screenshots is the line enqueuing  `frontendIndex` (and its dependencies) is not executed in the second (I put a check for a query param around it).

Inwerpsel: Not posting this here to say that we shouldn't use an effect for the lightbox. But if we do so we probably want to look into setting up a lightweight standalone React lib for the frontend which doesn't add so many dependencies.

pablocubico: @Inwerpsel That's a good point, when we started the move to React we actually thought of using Preact (3kb) in the frontend to minimize the impact, but as at first we didn't ship any React blocks to the frontend (we used Twig instead), I think the idea got buried for a few months. I used Preact a lot in the past and I would say 90% of the times the code is compatible  (sometimes requiring `preact-compat` which is just 2kb), so we could probably do a separate bundle using the same components. I guess it has some risk of not being compatible but maybe it's worth giving it a try. 

https://preactjs.com/guide/v10/differences-to-react

Otherwise I guess we would be fine just using react/reactDOM. 

Apart from all this, for the case of the Gallery, we should probably consider making it a static block...

mleray:
@pablocubico It seems that the test instance still has the old blue captions for the Gallery block (e.g. [here](https://k8s.p4.greenpeace.org/test-tavros/story/29431/birthday-healthy-forest-cameroon-deforestation-biodiversity-carbon/), you probably need to rebase in order to test with the new caption style? Since the caption is sometimes cut off now, lightbox would be the only way to read the whole thing...

assets/src/blocks/Gallery/GalleryGrid.js
pablocubico: I think we are not going to switch to a different package in this PR so I'll resolve this comment.

package.json
```diff
@@ -31,13 +34,13 @@
     "postcss-loader": "^3.0.0",
     "remove-files-webpack-plugin": "^1.1.3",
     "sass-loader": "^7.3.1",
+    "style-loader": "^2.0.0",
```
pablocubico, line 37: This is actually not needed as we are using custom styles for Photoswipe, I'll remove it.

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,146 @@
...
+  const imageBlocks = document.querySelectorAll('.wp-block-image');
+  imageBlocks.forEach((imageBlock, index) => {   
+    const caption = imageBlock.querySelector('figcaption');
+    if (caption) {
+      items[index].title = caption.innerHTML;
+    }
+
+    imageBlock.querySelector('img').addEventListener('click', () => wp.element.render(
+        <Lightbox isOpen={true} index={index} items={items} onClose={() => {}} />,
+        document.body.appendChild(document.createElement('div'))
```
pablocubico, line 23: Seems hacky? it is... I did this because the `createPortal` call is now inside the LightBox component, but I'll extract that into a `<Modal />` component, wrap the LightBox with `Modal` in the Gallery components, and use `createPortal` here instead of `render`.

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,154 @@
...
+  const options = {
+    index: props.index || 0,
+    closeOnScroll: false,
+    history: false
+  };
+
+  const [currentIndex, setCurrentIndex] = useState(options.index);
+
+  useEffect(() => {
+    const photoSwipe = new PhotoSwipe(pswpElement, PhotoSwipeUI_Default, props.items, options);
```
pablocubico, line 59: We don't want this to happen more than once per opened lightbox. 

Currently, when using a Carousel/Slider Gallery, this effect is ran each time the Carousel goes to the next slide.

Inwerpsel: I think having a complex dependency is always a bad thing with effects, they should be as simple a value as possible, relating to a specific effect. So definitely passing the entire props of the component seems like a bad idea, and causes the exact situation you want to avoid, i.e. no control over when things change.

For instance here you could have a single effect that only creates the `PhotoSwipe` on the initial render. If the instantiation of `PhotoSwipe` isn't slow/expensive we could do it synchronously as well using state.
```js
const [photoSwipe] = useState(()=>(new PhotoSwipe(pswpElement, PhotoSwipeUI_Default, props.items, options))
```

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,154 @@
...
+
+    if (photoSwipe) {
```
Inwerpsel, line 61: Early return here would make this a lot more readable

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,154 @@
...
+        });
+
+        // photoSwipe.listen('afterChange', () => {
+        //   setCurrentIndex(photoSwipe.currItem);
+        // });
+
+        photoSwipe.init();
+      }
+      if (!props.isOpen) {
+        props.onClose();
```
Inwerpsel, line 90: I guess this is supposed to call `props.onClose` when `isOpen` changed from true to false. But it actually runs on any change to either `prop` or `options`, which is probably causing some issues.

assets/src/components/Lightbox/Lightbox.js
Inwerpsel: Actually since we only ever display a single lightbox, we don't need to create it in react, we could include the initial markup in the page, then use a hook on each image block that will add the image to the lightbox, and exposes a function to open the lightbox at that image.

pablocubico: Hey monsieur! Not creating the HTML in React would be easy as it's just delivering the needed markup from a different place, but I see a few downsides. I guess you mean detecting whether a page/post has image or gallery blocks in the server side (probably using `has_block`) and include the HTML for the Lightbox depending on that. To include the markup in the page I guess we'll have to do that in the master theme, not in the blocks plugin. We'll probably need some condition in a Twig template and pass some variable to the Timber context. Is that right?

I think the benefit of the current approach is that the code is entirely encapsulated in the Lightbox component file instead of having logic scattered through the repos (or between PHP & JS in the same repo). There's also the advantage of having the HTML right there in case we need to make further customizations. What do you think?

Inwerpsel: > I guess you mean detecting whether a page/post has image or gallery blocks in the server side (probably using has_block) and include the HTML for the Lightbox depending on that.

It might not even be worth checking if a post has an image, the vast majority of them will do. And it's a tiny amount of markup anyway, it doesn't include the actually markup for the images. It's just a starting point Photoswipe needs once per page load, then in JS we just need to "hydrate" it and tell it to render certain images.

> To include the markup in the page I guess we'll have to do that in the master theme, not in the blocks plugin.

We could just as well include this in the blocks plugin if we want to keep it together with the JS, I don't see any reason why we couldn't.

> We'll probably need some condition in a Twig template and pass some variable to the Timber context. Is that right?

I think we can get away with the simplicity of including it each time, as the vast majority of posts will use it anyway. But if we do want to make it conditional that seems like a very minimal change. We also don't need to pass any variables, it's 100% static markup. It doesn't even need to be a twig file, you can include any file in a twig template.

```twig
<body>
...
{% include "photoswipe-skeleton.html" %}
</body>
```

pablocubico: Alright, that does make it simpler, but it'll be a bit more difficult to render some customizations as the indicators (currently rendered here: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/414/files#diff-7603fc2cbf7cc1eec26221206dffa1f05c92e5b18163c0ab259a2b83443d1767R139).

pablocubico: Ok, so I fixed the original issue by using `memo()` as a functional version of `shouldComponentUpdate`. It seems to be working as expected now.

Inwerpsel: Are you sure that's a solid solution? React seems to recommend against relying on it to prevent a render.

> This method only exists as a performance optimization. Do not rely on it to “prevent” a render, as this can lead to bugs.

from https://reactjs.org/docs/react-api.html#reactmemo

Inwerpsel (commented):
Leaving these comments already so they don't bunch up too much :slightly_smiling_face:

assets/src/blocks/Gallery/GalleryGrid.js
```diff
@@ -1,16 +1,37 @@
...
             style={{ objectPosition: image.focus_image }}
             alt={image.alt_text}
+            onClick={() => {
+              onImageClick(index);
+            }}
           />
         </div>
       ))}
     </div>
+    <Lightbox isOpen={lightboxIsOpen} index={lightboxIndex} items={items} onClose={onLightboxClose} />
```
Inwerpsel, line 35: When the lightbox calls `onClose`, the only thing that happens is the state variable `lightboxIsOpen` is set to false. But `lightboxIsOpen` is then only used to pass back into the lightbox.

Would it be possible to handle that internally in `Lightbox`?

assets/src/components/Lightbox/Lightbox.js
pablocubico: Yes, I read that while investigating `memo()`, but they don't shed much light on *why* it would cause bugs, I guess it's the same reason `shouldComponentUpdate` [should be used sparingly](https://reactjs.org/docs/react-component.html#shouldcomponentupdate). It's the same warning, and I think it's because the API may change, or you may have to update the comparison function if it's too complex and you change the props, or you can actually slow down the app if you are doing a really deep equality check in the comparison function. As this is not dealing with thousands of elements, I don't think the latter is an issue.

But if you look at the community, `shouldComponentUpdate` is still being used a lot to prevent re-renders, specially since most React examples in the docs use classes, not functional compos. I guess there's a reason it has not been deprecated or hidden behind some `__unsafe` prefix.

The problem with functional components is that they will always re-render their childs. There are a lot of community posts about using `memo()` as a workaround when the component is re-rendered using the same props ([example](https://dmitripavlutin.com/use-react-memo-wisely/)). It's pretty much the same as using `shouldComponentUpdate`, hence the same warnings apply.

So yes, I knew the warning, but I think it's safe as a workaround, the arguments against it seem too defensive to me, if I have to change the props of the Lightbox, I'll probably be changing some of its implementation too, and I can take care of any possible issues or change the implementation in that moment.

assets/src/components/Lightbox/Lightbox.js
pablocubico: I missed this one!

assets/src/blocks/Gallery/GalleryCarousel.js
```diff
@@ -1,16 +1,33 @@
...
 import _uniqueId from 'lodash/uniqueId';
+import { Lightbox, wpImagesToItems } from '../../components/Lightbox/Lightbox';
 
 const { __ } = wp.i18n;
 
 // This will trigger the browser to synchronously calculate the style and layout
 // You can find a list of examples here: https://gist.github.com/paulirish/5d52fb081b3570c81e3a
 const reflow = element => element.offsetHeight;
 
+const MemoLightbox = memo(Lightbox, (prevProps, nextProps) => {
```
Inwerpsel, line 11: Is there a reason why `memo` is only used for this lightbox, and not for the others?

assets/src/blocks/Gallery/GalleryCarousel.js
```diff
@@ -1,16 +1,33 @@
...
+  const onImageClick = lightboxIndex => {
+    setLightboxIsOpen(true);
+    setLightboxIndex(lightboxIndex);
+  };
+
+  const onLightboxClose = () => {
+    setLightboxIsOpen(false);
+  };
 
+  const items = wpImagesToItems(images);
```
Inwerpsel, line 30: I guess we could assume that these images will be the same structure in all places? Or if not we could make them so. As they basically come from the same data structure, in theory they should all have the same fields.

Then we could handle this mapping inside of the lightbox in one place, and can just pass the images to it instead.

assets/src/blocks/Gallery/GalleryCarousel.js
pablocubico: Yes, it's because the Carousel Gallery has a lot of state changes (which is another issue we could revisit in the future), because of the autoplay feature. So it gets a lot of re-renders on an interval. The others don't change.

assets/src/components/Lightbox/Lightbox.js
Inwerpsel: > they don't shed much light on why it would cause bugs

Yeah unfortunately they don't go into detail in the those docs, but I wouldn't assume that is because they don't have a good reason to say that, they probably forgot or didn't care to add it the docs. But it's a pretty clear and unambiguous statement "**Do not** rely on it to prevent a render...".

They do go into a little more detail as to why you can't rely on it actually preventing the render in all cases in [the `useMemo` hook docs](https://reactjs.org/docs/hooks-reference.html#usememo).

> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. **Write your code so that it still works without useMemo** — and then add it to optimize performance.

assets/src/blocks/Gallery/GalleryCarousel.js
Inwerpsel: > Carousel Gallery has a lot of state changes (which is another issue we could revisit in the future)

Can you elaborate what the issue is?

Inwerpsel: From the code it seems like it's doing only the necessary state changes at the right moment, but maybe I'm missing something.

pablocubico: Sometimes when you leave the page and go back to it (for example by switching tabs), the carousel is all white, and then a slide appears. I didn't debugged it yet, but I guess it could have something to do with the timers used there, maybe we can rely on `transitionend` events instead of using timers, which also have the problem of having to match the time of the transitions defined in the CSS side.

Inwerpsel: Why not add the `LightBox` component in `GalleryFrontend` instead of in the 3 different variants? :thinking: Then we would not have this issue, as the front end components' render only gets called once per page load.

Inwerpsel: It would solve another issue, that the editor is currently also including the lightbox, which probably won't play well.

pablocubico: Indeed! I'll try that. The three-columns gallery won't have a Lightbox so I added it to the other ones, but that can be easily controlled in GalleryFrontend.

pablocubico: Ah wait, there is an issue: we need the onClick event for the images, but I guess we can pass a handler as prop.

pablocubico: Update: moving the Lightbox component to GalleryFrontend definitely still fixes the issue and does not need `memo` hacks.

assets/src/blocks/Gallery/GalleryCarousel.js
pablocubico: Not sure, the caption for the images can come from different places. In the case of image blocks, it comes from the DOM. I'll take a look.

assets/src/components/Lightbox/Lightbox.js
pablocubico: I think there was a comment about GalleryFrontend around here... 

Anyway: as suggested, I moved the Lightbox component to GalleryFrontend and it definitely fixes the issue, no need for `memo()` then.

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,116 @@
...
+  const [currentIndex, setCurrentIndex] = useState(options.index);
+
+  useEffect(() => {
+    const photoSwipe = new PhotoSwipe(photoSwipeElement, PhotoSwipeUI_Default, props.items, options);
+
+    if (!photoSwipe) {
+      return;
+    }
+
+    if (!props.isOpen) {
```
Inwerpsel, line 24: This currently won't be run if  `props.isOpen` changes, it's not in the deps. But not suggesting to add it to the same effect, it should probably be a separate one, if we still need to use an effect for that.

assets/src/blocks/Gallery/GalleryGrid.js
pablocubico: So I went around this and we actually need that callback in the parent to toggle `isOpen` in the parent, so it reopens the Lightbox even if  `items` or `index` stays the same (e.g.: you close it and click the same image).

assets/src/blocks/Gallery/GalleryCarousel.js
pablocubico: I moved this function to `GalleryFrontend`, so it's  in a single place for galleries, but as said in the previous comment, building the items for image blocks is different, and it could be different for other use cases of the Lightbox, so I'd rather make the Lightbox expect the `items` collections it needs without doing any transforms internally.

assets/src/blocks/Gallery/GalleryFrontend.js
```diff
@@ -13,6 +23,20 @@ export const GalleryFrontend = ({
   const layout = getGalleryLayout(className, gallery_block_style);
   const postType = document.body.getAttribute('data-post-type');
 
+  const [lightboxIsOpen, setLightboxIsOpen] = useState(false);
+  const [lightboxIndex, setLightboxIndex] = useState(0);
+
+  const onImageClick = lightboxIndex => {
```
Inwerpsel, line 29: I guess this should be named after what the function does, rather than `onImageClick`. If for example there is another way we could open the lightbox at the image, it could also use the exact same function. So maybe something like `openLightBox = (index) =>` would be more useful.

Especially if you see it in JSX:
```
      {layout === 'slider' && <GalleryCarousel onImageClick={openLightbox} images={images || []} />}
```

assets/src/components/Lightbox/Lightbox.js
pablocubico: Moved it outside of the effect.

assets/src/blocks/Gallery/GalleryGrid.js
Inwerpsel: There's maybe a way around that, but maybe it's not worth digging in further, as there's not too many blocks using a lightbox. So I'm ok with the current version.

But it feels like there must be a way to create a Lightbox component that can be used without duplicating [this entire block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/076c55d83ecb2a8c50bb09c160d4b123937cc42c/assets/src/blocks/Gallery/GalleryFrontend.js#L26-L36) in each place.

Inwerpsel: Maybe a `useLightbox` hook could encapsulate that block of code?

Inwerpsel: Then `setLightboxIsOpen` and `setLightboxIndex` would be just internal of that hook, it only would need to expose the other things.

pablocubico: The Lightbox itself doesn't need that block: 
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/414/files#diff-5807ab99eebdcdb5559e04f7ee7d3fcff716a376e2b33ca0664fefc9fc19dfc0R28

That's just what GalleryFrontend does, not a hard requirement to use the Lightbox, it just needs the props. How you manage those props outside of it it's up to you, it can be different depending on the case.

Inwerpsel: Wouldn't it need all the same things if another block also creates a gallery? Besides I don't think it hurts to add the hook even if there's only a single usage, as it greatly improves readability imo.

pablocubico: The custom hook sounds good. As for your question, I think it does not need _exactly_ that, it could be handled differently. The Lightbox doesn't care.

Inwerpsel: Indeed the lightbox doesn't require it, but I rather mean if another block would use it the same way as Gallery, it would probably be needing the same things.

pablocubico: Right, so I extracted the block as a custom hook in a separate file, so it's reusable in other places.

Inwerpsel: :100:

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,119 @@
...
+  const [currentIndex, setCurrentIndex] = useState(photoSwipeOptions.index);
+
+  if (!isOpen) {
+    onClose();
+  }
+
+  useEffect(() => {
+    const photoSwipe = new PhotoSwipe(photoSwipeElement, PhotoSwipeUI_Default, items, photoSwipeOptions);
+
+    if (!photoSwipe) {
```
Inwerpsel, line 27: I don't think this can be the case? Since it's the result of a constructor, `photoSwipe` will always be truthy.

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,119 @@
...
+      return;
+    }
+
+    if (isOpen) {
```
Inwerpsel, line 31: `if (!isOpen) { return; }`

assets/src/components/Lightbox/Lightbox.js
pablocubico: True, removed.

assets/src/components/Lightbox/Lightbox.js
pablocubico: I changed it and moved it to the top.

assets/src/components/Lightbox/setupLightboxForImageBlocks.js
```diff
@@ -0,0 +1,34 @@
...
+export const setupLightboxForImageBlocks = function() {
+    // We can't use `createPortal` outside `render()`,
+    // `React.render()` needs a node, even if it ends up being empty.
+    // See: https://github.com/facebook/react/issues/12653#issuecomment-382851495
+    const lightBoxNode = document.createElement('div');
+    lightBoxNode.style.position = 'fixed';
+  
+    document.body.appendChild(lightBoxNode);
+  
+    const imageBlocks = document.querySelectorAll('.wp-block-image');
```
Inwerpsel, line 12: Maybe we can add something like `:not(.force-no-lightbox)` to this selector, as a quick way to implement excluding images by adding a class in the sidebar?

pablocubico: Nice, added it. 👍

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,118 @@
...
+  const photoSwipeOptions = {
+    index: index || 0,
+    closeOnScroll: false,
+    history: false
+  };
+
+  const [currentIndex, setCurrentIndex] = useState(photoSwipeOptions.index);
+
+  if (!isOpen) {
+    onClose();
```
Inwerpsel, line 21: I don't think we need to call this here. `onClose` was already called [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/414/files#diff-7603fc2cbf7cc1eec26221206dffa1f05c92e5b18163c0ab259a2b83443d1767R44) and/or [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/414/files#diff-7603fc2cbf7cc1eec26221206dffa1f05c92e5b18163c0ab259a2b83443d1767R48) at this point, right?

Inwerpsel: Hmm actually `onClose` just sets `isOpen` to false :smiley: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/a21971d3f325a396b5c360a342d5a234183d4fa2/assets/src/components/Lightbox/useLightbox.js#L13

assets/src/components/Lightbox/setupLightboxForImageBlocks.js
```diff
@@ -0,0 +1,34 @@
...
+        h: 0
+      };
+    
+      const caption = imageBlock.querySelector('figcaption');
+      if (caption) {
+        item.title = caption.innerHTML;
+      }
+  
+      imageBlock.querySelector('img').addEventListener('click', () => wp.element.render(
+          <Lightbox isOpen={true} index={index} items={[item]} onClose={() => {}} />,
```
Inwerpsel, line 28: You could also set onClose defaulting to a noop in the components arguments, then you could exclude `onClose` prop here.
```js
export const Lightbox = ({ index, isOpen, items, onClose = () => {} }) => {
...

File: assets/src/styles/vendors/photoswipe/photoswipe.scss [hunk 17]
```diff
@@ -0,0 +1,207 @@
...
+	font-size: 14px;
+	line-height: 16px;
+	margin-top: -8px;
+	color: $pswp__error-text-color;
+}
+
+.pswp__error-msg a {
+	color: $pswp__error-text-color;
+	text-decoration: underline;
+}
```
Inwerpsel, line 207: Maybe we should add `cursor: pointer` or something other than the regular cursor when hovering an image that would open the lightbox on click.

pablocubico: Indeed, that was actually a requirement! Good catch.

pablocubico:
Hey @Inwerpsel , all pending items were addressed, care to do another check?

assets/src/styles/blocks/Gallery_carousel.scss
```diff
@@ -57,6 +57,7 @@
     img {
       width: 100%;
       object-fit: cover;
+      cursor: pointer;
```
Inwerpsel, line 60: I guess we also need to add `cursor: pointer` for `.wp-block-image:not(force-no-lightbox)`?

assets/src/styles/vendors/photoswipe/photoswipe.scss
```diff
@@ -0,0 +1,207 @@
+/*! PhotoSwipe main CSS by Dmitry Semenov | photoswipe.com | MIT license */
```
Inwerpsel, line 1: This file is based on a file the photoswipe lib provides as a starting point, if I'm correct? Does it have any modifications compared to that file?

 Maybe you can link to the file this is based on? Ideally if they host it in Github we can link to the commit you used.

Inwerpsel (commented):
Got a few small remarks, nothing too crucial except for the pointer cursor on image blocks, if I'm correct it's only on Gallery images currently. Will approve after that's cleared out :+1:

assets/src/components/Lightbox/setupLightboxForImageBlocks.js
```diff
@@ -0,0 +1,34 @@
...
+      }
+  
+      imageBlock.querySelector('img').addEventListener('click', () => wp.element.render(
+          <Lightbox isOpen={true} index={index} items={[item]} />,
+          lightBoxNode
+        )
+      );
+    })
+  }
+
```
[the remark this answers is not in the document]
pablocubico: - Ah, indeed, that's probably like this since I had to switched to a PC, my Mac is in the doctor... 👨‍⚕️

assets/src/styles/vendors/photoswipe/photoswipe.scss
```diff
@@ -0,0 +1,207 @@
...
+
+/*
+	Styles for basic PhotoSwipe functionality (sliding area, open/close transitions)
+*/
+
+// PhotoSwipe uses Autoprefixer, so vendor prefixed are added automatically when needed.
```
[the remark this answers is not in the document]
pablocubico: That's from the author's setup, I removed these comments from the files.

pablocubico:
Hey @Inwerpsel , the cursor is now updated in image blocks and the unnecessary `return` removed. 

As for the Photoswipe files, I added comments to the tagged versions in Github, but also renamed them so they have the same names they had in the Photoswipe repo, so it's absolutely clear where they came from.

assets/src/styles/vendors/photoswipe/photoswipe.scss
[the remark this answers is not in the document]
pablocubico: These comments were from the author's setup, I removed them.

comzeradd:
@pablocubico can you also rebase the branch? Just to make Acceptance tests pass

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -0,0 +1,114 @@
+import React, { useEffect, useRef, useState } from 'react';
+
+import PhotoSwipe from '../../../../node_modules/photoswipe/dist/photoswipe.js';
+import PhotoSwipeUI_Default from '../../../../node_modules/photoswipe/dist/photoswipe-ui-default.js';
+
+// `items` should be an array of object with this shape:
+// [{ src, w, h, title }, ...]
```
Inwerpsel, line 7: [a reflection rather than a remark] Maybe we can look into TypeScript at some point for cases like this to make it easier to define the expected shape of props.

public/images/vendors/photoswipe/default-skin.svg
```diff
@@ -0,0 +1 @@
...5v2h-3v3h-2v-5zM31 15v5h-2v-3h-3v-2h5zM31 29h-5v-2h3v-3h2v5z" id="Shape"/></g><g fill="#111"><path d="M62 24v5h-2v-3h-3v-2h5zM62 20h-5v-2h3v-3h2v5zM70 20v-5h2v3h3v2h-5zM70 24h5v2h-3v3h-2v-5z"/></g><path d="M20.586 66l-5.656-5.656 1.414-1.414L22 64.586l5.656-5.656 1.414 1.414L23.414 66l5.656 5.656-1.414 1.414L22 67.414l-5.656 5.656-1.414-1.414L20.586 66z" fill="#111"/><path d="M111.785 65.03L110 63.5l3-3.5h-10v-2h10l-3-3.5 1.785-1.468L117 59l-5.215 6.03z" fill="#111"/><path d="M152.215 65.03L154 63.5l-3-3.5h10v-2h-10l3-3.5-1.785-1.468L147 59l5.215 6.03z" fill="#111"/><g><path id="Rectangle-11" fill="#111" d="M160.957 28.543l-3.25-3.25-1.413 1.414 3.25 3.25z"/><path d="M152.5 27c3.038 0 5.5-2.462 5.5-5.5s-2.462-5.5-5.5-5.5-5.5 2.462-5.5 5.5 2.462 5.5 5.5 5.5z" id="Oval-1" stroke="#111" stroke-width="1.5"/><path fill="#111" d="M150 21h5v1h-5z"/></g><g><path d="M116.957 28.543l-1.414 1.414-3.25-3.25 1.414-1.414 3.25 3.25z" fill="#111"/><path d="M108.5 27c3.038 0 5.5-2.462 5.5-5.5s-2.462-5.5-5.5-5.5-5.5 2.462-5.5 5.5 2.462 5.5 5.5 5.5z" stroke="#111" stroke-width="1.5"/><path fill="#111" d="M106 21h5v1h-5z"/><path fill="#111" d="M109.043 19.008l-.085 5-1-.017.085-5z"/></g></g></g></svg>
```
Inwerpsel, line 1: line break

assets/src/styles/style.scss
```diff
@@ -6,3 +6,6 @@
 @import "styleguide/src/base/fonts";
 
 @import "blocks";
+
+// Third-party vendor overrides / customizations
+@import "vendors/photoswipe/planet4-photoswipe";
```
Inwerpsel, line 11: This seems like a good candidate to load at the body end, since we already have way to much blocking CSS, and I guess for now we don't have a way to open the lightbox that quickly :smile:

Inwerpsel (commented):
All looks good, one thing only now occurred to me. We already have way to much blocking CSS, and the lightbox CSS could be a good candidate to load at the body end. So seems like a small change with a lot of benefit to serve it in a separate file at the end, what do you think?

assets/src/styles/style.scss
pablocubico: Indeed, I was hesitant about adding more CSS (and JS) to the bundle, I considered loading them using the hooks from the Media PR (useStylesheet...), but as we need it for image blocks and that is pretty much on all pages, I left them in the bundle. 

I'll move the CSS away and enqueue separately. As for the JS stuff, in the future we should consider unbundling the blocks and loading only the ones we need on each page.

pablocubico:
Hey @Inwerpsel @mleray , unless I'm missing something, this should be ready to go!

assets/src/components/Lightbox/setupLightboxForImageBlocks.js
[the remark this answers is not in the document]
Inwerpsel: Still a few left :wink:

mleray (changes requested):
While checking the test instance, I found two cases of images that don't have the lightbox and I think should have it:
- Gallery 3 columns style (I see in the code that this seems to be the wanted behaviour, but I don't see why?)
- Images that are added via classic WP editor (instead of Image block)
You can find examples of both of these on [this page](https://www-dev.greenpeace.org/test-tavros/explore/ships/arctic-sunrise/)

pablocubico:
Hey @mleray ! I talked with Mag and the lightbox is now enabled in all the Gallery styles. 

I added support for other image containers like the ones inside the Classic blocks. You can test again in the page you provided: 
https://www-dev.greenpeace.org/test-tavros/explore/ships/arctic-sunrise/

Cheers!

pablocubico:
@mleray Ah, forgot to pin the [related master branch](https://github.com/greenpeace/planet4-master-theme/pull/1280), which removes the old lightbox.

mleray:
> Hey @mleray ! I talked with Mag and the lightbox is now enabled in all the Gallery styles.
> 
> I added support for other image containers like the ones inside the Classic blocks. You can test again in the page you provided:
> https://www-dev.greenpeace.org/test-tavros/explore/ships/arctic-sunrise/
> 
> Cheers!

@pablocubico It works great now, thanks! I just noticed that the Gallery 3-image and the Classic blocks' images don't show a pointer cursor on hover, but maybe that's just because the test instance isn't up to date?

pablocubico:
@mleray Fixed! It was missing a few `cursor: pointer`s.

mleray (approved):
Looks good to me 👍

PR #416, opened 2020-10-21
PLANET-5607 No apiFetch on frontend, it breaks wpml and nonces

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5607

`apiFetch` fails if the page is cached with an expired nonce, then it fails to fetch a new nonce. The fix is to not use `apiFetch` on the frontend but a regular `fetch` instead. This probably also prevents an issue with wpml, as we applied the same solution for such an issue in the Articles block previously, and now that block is not experiencing the nonce issue.

### DIFF
--- a/assets/src/functions/fetchJson.js
+++ b/assets/src/functions/fetchJson.js
@@ -1,0 +1,18 @@
+/**
+ * Function with a similar signature as WordPress's apiFetch, but doesn't do a bunch of things we don't need and cause
+ * issues. You could as well use what is inside this function directly, but having this in a single function makes it
+ * easier to use in a ternary assignment, as in the editor it still needs to use apiFetch in the same place.
+ *
+ * 1) It will pass a nonce, which is on the page but could have been cached and expired. Even when an endpoint
+ * doesn't require a nonce, passing an invalid one will make it fail. Passing nonces on public endpoints is pretty
+ * pointless regardless.
+ *
+ * 2) It doesn't use the correct language for the API when using WPML. An alternative is to use what we already have in
+ * document.body.dataset.nro, which has the language suffix, so will cause the API to use that language.
+ *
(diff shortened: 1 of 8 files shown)


assets/src/functions/fetchJson.js
```diff
@@ -0,0 +1,4 @@
+export const fetchJson = async(url) => {
```
lithrel, line 1: Can you add an explanation for the existence of this function here (nonce cached problem) ? There's a [PR on gutenberg that seems to add an option `withNonce` in `apiFetch`](https://github.com/WordPress/gutenberg/pull/24623/files#diff-89239b7bd7836db63dd0ddb87c358ceb6fcfc8227469ffe796e85501338805d9R24), we might not need this anymore if it is merged at some point.

pablocubico (approved):
LGTM. Tested the affected blocks locally (using `https://www.planet4.test/` as NRO).
Adding a comment to `fetchJson` as @lithrel suggested sounds good. Maybe a reference to the JIRA issue.

assets/src/functions/fetchJson.js
Inwerpsel: `apiFetch` doesn't really do anything except adding the nonce (and allowing to use [some middleware we probably don't need](https://github.com/WordPress/gutenberg/blob/master/packages/api-fetch/README.md#built-in-middlewares)). The nonce makes no sense to use on a public endpoint, and we're starting to cache these endpoint responses in Cloudflare.

Inwerpsel: And also using `apiFetch` means loading yet another resource before `frontendIndex.js`. But maybe I'm missing something about `apiFetch` that could still make it useful :slightly_smiling_face:

PR #418, opened 2020-10-22
PLANET-5345 - Firebase + ControlShift PoC - Connect to an emulator instance

pablocubico:
This hijacks the DB config on the initialized App object to connect to a local emulator instance. This way we can play with the data without touching the production instances. It also enables us to run CI workflows using the emulators with test data.

See: https://firebase.google.com/docs/emulator-suite/connect_firestore#instrument_your_app_to_talk_to_the_emulators

Ref: https://jira.greenpeace.org/browse/PLANET-5345

### DIFF
--- a/assets/src/blocks/ControlShift/firebase.js
+++ b/assets/src/blocks/ControlShift/firebase.js
@@ -16,4 +16,12 @@ const devConfig = {
 firebase.initializeApp(devConfig);
 firebase.analytics();
 
+export const db = firebase.firestore();
+if (location.host === 'www.planet4.test') {
+  db.settings({
+    host: "localhost:8081",
+    ssl: false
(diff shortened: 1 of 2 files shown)


assets/src/blocks/ControlShift/firebase.js
```diff
@@ -16,4 +16,12 @@ const devConfig = {
 firebase.initializeApp(devConfig);
 firebase.analytics();
 
+export const db = firebase.firestore();
+if (location.host === 'www.planet4.test') {
```
pablocubico, line 20: These host and port values should probably come from ENV vars or somewhere else.

comzeradd:
Closing, but keeping the branch for now as a reference

PR #421, opened 2020-10-27
PLANET-5501 Update EN form button styles

mleray:
When updating the button styles for [PLANET-5501](https://jira.greenpeace.org/browse/PLANET-5501) we forgot some overrides in the EN forms 😬


lithrel (approved):
:+1:

PR #423, opened 2020-11-03
Media change listener

Inwerpsel:
This makes the component set the embed html in the input change listener so that we don't need to set the attributes in an intermediary invalid state until the effect runs, but instead sets both at the same time. This ensures that no render can unintentionally trigger a change to the post content.

Since the backend will still render any block that has no content**, we can make the block return `null` in save whenever it's missing either (or both) `embed_html` or `poster_url`, but does have the corresponding "identifier" prop (`media_url` or `video_poster_img`). Then it uses [`useSelect` in the edit component](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/423/files#diff-c6c91408622162ae2bd090b6d50b303d1de0a3ebff4e385997189643c4a18ce1R124) to get the missing html or url.

We can in a follow up make a migration script to migrate existing content, then after running it we only need to remove a few lines from `MediaEditor` and `save`.

_** well technically it adds the missing attributes for the frontend to render_

### DIFF
--- a/assets/src/blocks/Media/MediaBlock.js
+++ b/assets/src/blocks/Media/MediaBlock.js
@@ -6,27 +6,14 @@ const {__} = wp.i18n;
 
 const BLOCK_NAME = 'planet4-blocks/media-video';
 
-const attributes = {
-  video_title: {
-    type: 'string'
-  },
-  description: {
-    type: 'string'
-  },
-  video_poster_img: {
-    type: 'integer'
-  },
-  embed_html: {
-    type: 'string',
-    default: ''
-  },
-  media_url: {
-    type: 'string'
-  },
-  poster_url: {
-    type: 'string',
-    default: ''
-  },
+// Old content didn't store the poster url and embed html in the props. For the frontend this is caught in the block's
+// backend render method, where we setup a div to be frontend rendered by the same MediaFrontend component as is used for
+// save.
+export const lacksAttributes = attributes => {
+  const lacksEmbedHtml = attributes.media_url && !attributes.media_url.endsWith('.mp4') && !attributes.embed_html;
+  const lacksPosterUrl = attributes.video_poster_img && !attributes.poster_url;
+
+  return lacksEmbedHtml || lacksPosterUrl;
 };
 
 export const registerMediaBlock = () => {
@@ -36,9 +23,38 @@ export const registerMediaBlock = () => {
     title: __('Media block', 'planet4-blocks-backend'),
     icon: 'format-video',
     category: 'planet4-blocks',
-    attributes,
-    save: MediaFrontend,
+    attributes: {
+      video_title: {
+        type: 'string'
+      },
+      description: {
+        type: 'string'
+      },
+      video_poster_img: {
+        type: 'integer'
+      },
+      embed_html: {
+        type: 'string',
+        default: ''
+      },
+      media_url: {
+        type: 'string'
+      },
+      poster_url: {
+        type: 'string',
+        default: ''
+      },
+    },
+    save: ({ attributes }) => {
+      if (lacksAttributes(attributes)) {
+        return null;
+      }
+
+      return <MediaFrontend attributes={attributes} />;
+    },
     edit: MediaEditor,
-    mediaV1,
+    deprecated: [
+      mediaV1,
+    ],
   });
 };
--- a/assets/src/blocks/Media/MediaBlock.js
+++ b/assets/src/blocks/Media/MediaBlock.js
@@ -50,7 +50,7 @@ export const registerMediaBlock = () => {
         return null;
       }
 
-      return <MediaFrontend attributes={attributes} />;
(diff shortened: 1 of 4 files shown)


assets/src/blocks/Media/MediaEditor.js
```diff
@@ -17,10 +17,24 @@ const MediaInspectorOptions = ({ attributes, setAttributes }) => {
   // https://github.com/facebook/react/issues/14904
   // https://github.com/facebook/react/issues/955#issuecomment-469352730
   const [ media_url, setMediaURL ] = useState(attributes.media_url);
-  const debouncedMediaURLUpdate = useCallback(debounce(value => setAttributes({ media_url: value }), 300), []);
+  const debouncedMediaURLUpdate = useCallback(debounce(value => {
+    setAttributes({ embed_html: null });
```
pablocubico, line 21: Why the need to set `null` here?

Inwerpsel: We probably don't, I added it so that the previous embed html is already cleared before the api request to get the embed is completed, but I guess we can just leave the previous html in place until the new arrives.

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -17,10 +17,24 @@ const MediaInspectorOptions = ({ attributes, setAttributes }) => {
...
 
-  function onSelectImage({id}) {
-    setAttributes({video_poster_img: id});
+    const embedPreview = wp.data.select('core').getEmbedPreview(resolveURL(value));
+
+    const embed_html = embedPreview ? embedPreview.html : null;
+
+    setMediaURL(value);
```
pablocubico, line 27: As discussed by chat, this is already set in the onChange listener, it has to be there so the text is immediately updated in the input, otherwise there'll be a 300ms delay while typing.

assets/src/blocks/Media/MediaBlock.js
```diff
@@ -17,6 +17,16 @@ const attributes = {
   },
 };
 
+// Old content didn't store the poster url and embed html in the props. For the frontend this is caught in the block's
+// backend render method, where we setup a div to be frontend rendered by the same MediaFrontend component as is used for
+// save.
+export const lacksAttributes = attributes => {
```
pablocubico, line 23: Isn't this (and `patchLegacyAttributes` below) introducing more backwards compatibilty code than the current version? It looks pretty version-specific too, I think it's even harder to adapt or remove.

Inwerpsel: It may be more lines, but they're much easier to remove than the effect:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/compare/media-change-listener...media-change-listener-remove-bc

Inwerpsel: Also with this PR the BC code only affects content that actually needs it, while keeping our render code clean and not performing updates of the block's attributes, which is not a render concern.

Inwerpsel: > It looks pretty version-specific too

Isn't BC code by nature version specific? :smile:

pablocubico: Ha, I guess, but some can be more specific than others. So what's the next step here? Remove the WIP label?

Inwerpsel: Sure, I'll just rebase it :+1:

assets/src/blocks/Media/MediaBlock.js
```diff
@@ -36,9 +23,38 @@ export const registerMediaBlock = () => {
...
+    save: ({ attributes }) => {
+      if (lacksAttributes(attributes)) {
+        return null;
+      }
+
+      return <MediaFrontend { ...attributes } />;
+    },
     edit: MediaEditor,
-    mediaV1,
+    deprecated: [
```
Inwerpsel, line 56: I don't think the previous code worked as `mediaV1` was only an array, not an object with the deprecated key.

assets/src/blocks/Media/MediaEditor.js
Inwerpsel: We ended up using the `defaultValue` value prop instead to resolve the issue

assets/src/blocks/Media/MediaFrontend.js
```diff
@@ -19,7 +17,7 @@ const wrapEmbedHTML = embed_html => {
   return wrapperDiv.outerHTML;
 }
 
-export const MediaFrontend = ({ attributes }) => {
+export const MediaFrontend = ( attributes ) => {
```
Inwerpsel, line 20: Had to undo this change I suggested earlier on your PR, my bad! It needs to be matched in the frontend rendering code too, but then we would need to change the other frontend components, which is a bit too much for a PR that's already big.

pablocubico (approved):
Looks good, discussed by chat and tested locally.

PR #425, opened 2020-11-04
PLANET-5633: Gallery block 3 Columns style overly zoom on images

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5633

> On both the front and back ends the 3-column style of the gallery block is not working as expected. 
> The images are overly zoomed and the centre function in the wysiwyg is not functioning.

## Fix 

- Switching object-fit property from none to cover, to give each image more visibility and more moving space with focus.
- Fixing a small notice `Undefined index` on dev env that would break the ajax request when gallery_block_style is not specified.

Before:
Screenshot from 2020-11-04 15-28-33

After:
Screenshot from 2020-11-04 15-27-35

## Test

- Insert a Gallery block in a page, add at least 3 images
- Select the style "3 columns"
- You should see more parts of the image, and the focus point should have an effect on the part of the image visible.

### DIFF
--- a/assets/src/styles/blocks/Gallery_content_three_column.scss
+++ b/assets/src/styles/blocks/Gallery_content_three_column.scss
@@ -25,7 +25,7 @@
       position: relative;
       left: -35%;
       max-width: inherit;
-      object-fit: none;
+      object-fit: cover;
 
       @include mobile-only {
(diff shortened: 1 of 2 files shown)


mleray (approved):

PR #427, opened 2020-11-06
PLANET-5652: Articles block: Tags containing an ampersand are displayed escaped

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5652

> In the Articles block, tags listed on top of each article are escaped and will display an ampersand as &amp;

## Fix 

Using `unescape()` function to fix tag names

Before fix:
Screenshot from 2020-11-06 10-50-24

After fix:
Screenshot from 2020-11-06 10-49-31

## Test

- Find an Articles block (Related articles) listing some articles with tags
- Edit one of those tags to add a character `&` in its title
- It should appear as `&` in the tag list of the related articles


sagarsdeshmukh (approved):
Tested on local, LGTM 👍
It would be nice if we also add same fix for Post type. :)

PR #430, opened 2020-11-10
Less libs

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5503

I started this out as a branch to test a few things while going through the pagespeed report, I don't think these actually require further changes, so it seems safe to merge these already in the context of the analysis ticket.

* Format date in the back end for Articles so we don't need to include an expensive lib on the frontend. (Actually only a small change in the PHP file, but I also made it use `array_map`, hence all the line changes)
* `wp-url` was only needed for `addQueryArgs`. For our use case on the front end this is easily replaceable with a custom function.

### DIFF
--- a/assets/src/blocks/Articles/useArticlesFetch.js
+++ b/assets/src/blocks/Articles/useArticlesFetch.js
@@ -1,8 +1,8 @@
 import { useState, useEffect } from '@wordpress/element';
 import { fetchJson } from '../../functions/fetchJson';
+import { addQueryArgs } from '../../functions/addQueryArgs';
 
 const { apiFetch } = wp;
-const { addQueryArgs } = wp.url;
 
 export const useArticlesFetch = (attributes, postType, postId, baseUrl = null, postCategories = []) => {
   const { article_count, post_types, posts, tags, ignore_categories } = attributes;
@@ -34,11 +34,12 @@ export const useArticlesFetch = (attributes, postType, postId, baseUrl = null, p
       args.categories = postCategories;
     }
 
+    const path = addQueryArgs('planet4/v1/get-posts', args);
 
     try {
       const response = baseUrl
-        ? await fetchJson(`${ baseUrl }/wp-json/${ addQueryArgs('planet4/v1/get-posts', args) }`)
-        : await apiFetch({ path: addQueryArgs('planet4/v1/get-posts', args) });
+        ? await fetchJson(`${ baseUrl }/wp-json/${ path }`)
+        : await apiFetch({ path });
 
       const newPosts = [...prevPosts, ...response.recent_posts];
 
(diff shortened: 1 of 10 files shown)


assets/src/blocks/Articles/useArticlesFetch.js
```diff
@@ -1,8 +1,8 @@
 import { useState, useEffect } from '@wordpress/element';
 import { fetchJson } from '../../functions/fetchJson';
+import { addQueryArgs } from '../../functions/addQueryArgs';
```
pablocubico, line 3: Wouldn't this actually increase the bundle size given that `wp` is an external and is not included in the pipeline?

Inwerpsel: No, because we can dequeue `wp.url` for the front end, it's the only thing we use from it. https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/430/commits/9c606f0b3b14c9afd274e90f997071b155aa6408#diff-4baaf0cfa9f65dafcc2bed33ae43cbc221e00364aaa33ec6b6c71b49e56ea5c0L402

pablocubico: Ah, got it, didn't see it had its own package 👍

Inwerpsel: There's also `apiFetch` in files we use on the front end, but we actually don't use it there already, because of an issue with WPML.

Inwerpsel: Like here https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/9c606f0b3b14c9afd274e90f997071b155aa6408/assets/src/blocks/Articles/useArticlesFetch.js#L41-L41

pablocubico: So it does seem to increase the bundle size (by some totally negligible few bytes), in some files. Of course it doesn't matter but maybe we could add some bundle comparison tool in the future for before/after checks, some can be integrated to the CI pipeline.

Inwerpsel: I made up for those bytes by removing some unused imports :smile:

Inwerpsel: And of course `wp.url` is much bigger, 4.7K compressed

sagarsdeshmukh:
The articles date formatting change looks good 👍 
Regarding `addQueryArgs` function change, While testing the articles block "Load more" button - It adds `undefined` value to some params.
image

Inwerpsel:
> The articles date formatting change looks good 
> Regarding `addQueryArgs` function change, While testing the articles block "Load more" button - It adds `undefined` value to some params.
> image

Right, overlooked that `addQueryArgs` handles these this way, added something to filter those out, as well as empty strings like `wp.url.addQueryArgs` does.

Also I needed to add something for arrays, as these need to be added to the url using multiple params `path?array_param[0]=foo&array_param[1]=bar`

sagarsdeshmukh (approved):
Tested on local. LGTM 👍

PR #431, opened 2020-11-10
Only add mediaelement script when needed

Inwerpsel:


assets/src/blocks/Media/setupMediaElementJS.js
```diff
@@ -3,17 +3,20 @@ import { addLinkTag } from '../../components/useStyleSheet/addLinkTag';
 
 export const setupMediaElementJS = function() {
 	addLinkTag({ href: 'https://cdn.jsdelivr.net/npm/mediaelement@4.2.16/build/mediaelementplayer-legacy.min.css' });
+  const meJSNodes = document.querySelectorAll( '.mejs-video-block' );
```
pablocubico, line 6: How about moving the early return block to the top, before `addLInkTag`?

Inwerpsel: Ah right missed that one :+1:

pablocubico (approved):
LGTM! Tested locally.

PR #432, opened 2020-11-11
PLANET-5503 Add slick on blocks that need it

Inwerpsel:
* That way we don't need to add it globally.


Inwerpsel:
Turns out it was only the Covers block that has a usage of slick. I mistakenly thought that Carousel Header also needs it, but turns out that was caused by the `frontendIndex.js` depending on the `'slick'` handle.

Additionally found that the only usage in Covers is actually not working atm. We can remove the script now (and not include `pubslider.js` as it's also not doing anything), and check if we need to add it again or find an alternative solution in https://jira.greenpeace.org/browse/PLANET-5687

PR #434, opened 2020-11-13
PLANET-5503 Remove slick dep and pubslider

Inwerpsel:
* Slick was added as a dependency of the plugin's styles, even though
it's not needed.
* It turns out the pubslider.js script is currently not doing anything
as it uses a wrong selector, so doesn't match anything. It's also the
only usage of slick slider. I'll keep the script itself in as we might
fix it in a follow up ticket.

### DIFF
--- a/classes/blocks/class-covers.php
+++ b/classes/blocks/class-covers.php
@@ -123,7 +123,6 @@ class Covers extends Base_Block {
 		// Enqueue js for the frontend.
 		if ( ! $this->is_rest_request() ) {
 			\P4GBKS\Loader::enqueue_local_script( 'covers', 'public/js/load_more.js', [ 'jquery' ] );
-			\P4GBKS\Loader::enqueue_local_script( 'pubslider', 'public/js/pubslider.js', [ 'jquery' ] );
 		}
 
 		$data = [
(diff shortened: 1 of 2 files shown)


classes/blocks/class-covers.php
```diff
@@ -123,7 +123,6 @@ public function prepare_data( $fields ): array {
 		// Enqueue js for the frontend.
 		if ( ! $this->is_rest_request() ) {
 			\P4GBKS\Loader::enqueue_local_script( 'covers', 'public/js/load_more.js', [ 'jquery' ] );
-			\P4GBKS\Loader::enqueue_local_script( 'pubslider', 'public/js/pubslider.js', [ 'jquery' ] );
```
pablocubico, line 126: Shouldn't we remove the `pubslider.js` file in this PR?

Inwerpsel: It's possible that we want to re-add it in the [follow up ticket](https://jira.greenpeace.org/browse/PLANET-5687) if we decide to just fix the bug.

pablocubico: Ahhh ok.

pablocubico (approved):
LGTM! Less libs 🏃

PR #435, opened 2020-11-13
Optimized version of overlay 83K->13K

Inwerpsel:
We may also be able to replace this with pure CSS, but a quick fix is to already replace it with a significantly smaller optimized version.


pablocubico (approved):
Looks good. Didn't know this was 83K... 🤦

PR #437, opened 2020-11-16
Include gallery data in page, add missing srcset

Inwerpsel:
That way it doesn't require an endpoint call to render.

I also noticed that a lot of our pages were getting the "properly size images" warning in pagespeed tests. Turns out that for the Gallery Grid view, which is the smallest way to show the image, it wasn't specifying srcset, even though it's available at that point. 

As a result it was using an image that is 1200px wide while only 222px is displayed (the actual display size is a bit bigger as we cut off a part with the focal point, but not that much).

Screenshot from 2020-11-16 16-53-57


mleray (approved):
Looks good to me 👍

PR #438, opened 2020-11-18
Scrape spreadsheet from html on the front end

Inwerpsel:
* Google for whatever reason started redirecting to a login page for the
csv format of Google Sheets publications, which are supposed to be
completely public. Probably an issue on their end, either a bug caused
the auth to fire unintentionally, or they needed to shut it down because
of an attack?

### DIFF
--- a/assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
+++ b/assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
@@ -2,10 +2,7 @@ import { Component, Fragment } from '@wordpress/element';
 import { ArrowIcon } from './ArrowIcon';
 import { toDeclarations } from '../toDeclarations';
 import { HighlightMatches } from '../HighlightMatches';
-import { fetchJson } from '../../functions/fetchJson';
 
-const { apiFetch } = wp;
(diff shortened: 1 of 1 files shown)


lithrel (approved):
We're extracting sheet id to reinsert it in the same url :grin: and skipping the php cache mechanism that's currently in the local endpoint, but that's something for another ticket (if still useful), it works locally and seems a bit urgent :+1: 
Do we have a good NRO to use for test instance with multiple examples ?

Inwerpsel:
> We're extracting sheet id to reinsert it in the same url  and skipping the php cache mechanism that's currently in the local endpoint, but that's something for another ticket (if still useful), it works locally and seems a bit urgent 
> Do we have a good NRO to use for test instance with multiple examples ?

Yeah I specifically chose to call the url from the front end because we're in need of a quick solution and that seems the simplest, didn't seem to make sense to put a lot of extra time in refactoring the back end. But luckily @comzeradd found an issue report that has a way to fix it on the affected spreadsheets, so we don't need this fix. So we can instead check all unique Google sheets that are used and ask owners to update the sharing setting. https://github.com/NUKnightLab/TimelineJS3/issues/667#issuecomment-729623444

PR #439, opened 2020-11-23
PLANET-5706: Escaped html entities are displayed in the post excerpt in Articles block

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5706

> Escaped html entities are displayed in the articles text (here, `&nbsp;`).
> The template currently only replaces the entity `&hellip;`

As for some other fields, the post excerpt has to be filtered (unescaped) so that the html entities appear as expected (`&hellip;` as `…`, `&nbsp;` as a non-breakable space, etc.).

image_from_ios

## Fix 

Same fix used for similar cases, by unescaping the content of `post_excerpt` on the frontend.

## Test

- Find or create  a page with an Articles block
- Edit the excerpt of one of those articles, add a `&` sign
- The Articles block on the frontend should show a `&` sign, not a `&amp;`

### DIFF
--- a/assets/src/blocks/Articles/ArticlePreview.js
+++ b/assets/src/blocks/Articles/ArticlePreview.js
@@ -12,7 +12,7 @@ export class ArticlePreview extends Component {
     const { isCampaign } = this.props;
     const className = `tag-item tag-item--main page-type page-type-${pageType.toLowerCase().replace(' ', '_')}`;
     if (isCampaign) {
-      return <span className={className}>{pageType}</span>;
+      return <span className={className}>{unescape(pageType)}</span>;
     }
     return <a
               className={className}
@@ -159,7 +159,7 @@ export class ArticlePreview extends Component {
 
           {post_excerpt &&
             <p className="article-list-item-content">
-              {post_excerpt.replace('&hellip;', '...')}
+              {unescape(post_excerpt)}
             </p>
           }
         </div>


mleray (commented):
Looks good to me, however I noticed a few things while checking the file!
- The [campaign page type tags](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlePreview.js#L15) are unescaped, unlike the [non-campaign ones](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlePreview.js#L23), probably a mistake?
- The [post title](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlePreview.js#L146) is unescaped but I feel like the issue could happen there too?

lithrel:
>   The [campaign page type tags](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlePreview.js#L15) are unescaped, unlike the [non-campaign ones](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlePreview.js#L23), probably a mistake?

Good point, missed it on last fix :grimacing: Done.

>   The [post title](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Articles/ArticlePreview.js#L146) is unescaped but I feel like the issue could happen there too?

From my tests it doesn't happen with title, and I'm not sure why. Might be some attributes [in the title definition](https://github.com/WordPress/gutenberg/blob/390402e4541e35b01b9258c8c20f7dec56e04303/packages/block-library/src/post-title/block.json#L33) or a specific save function.

mleray (approved):
👍

PR #442, opened 2020-11-25
PLANET-5442: Unordered list font size bigger than paragraph in the editor

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5442

> In the editor the unordered list font size is bigger than paragraphs (at least on large screens, smaller screens seem fine), see screenshot. It should be the same font size for both. Ordered lists seem fine too but it's worth double checking. In the frontend however it's all good!

_Some work was done on lists in editor in https://github.com/greenpeace/planet4-styleguide/pull/72, so the difference left was less than reported at the time._

Lists were a little bit bigger on large screen, and smaller on small screens.

| before | after |
| --- | --- |
| Screenshot from 2020-11-25 17-40-46 | Screenshot from 2020-11-25 17-39-32 |

## Fix

Unset font-size modifications set [by the styleguide](https://github.com/greenpeace/planet4-styleguide/blob/b27c77400a9a94398cf6ec263859de7ec2df1597/src/base/_typography.scss#L21).

## Test

- Write a paragraph and an unordered list in the post editor.
- Put the editor in mobile size and in large screen size (>1200px)
- Text in paragraph and list should have the same size.

### DIFF
--- a/assets/src/styles/editorOverrides.scss
+++ b/assets/src/styles/editorOverrides.scss
@@ -13,8 +13,14 @@
   }
 }
 
-.edit-post-visual-editor .wp-block {
-  max-width: 90%;
+.edit-post-visual-editor {
+  .wp-block {
+    max-width: 90%;
+  }
+
+  li {
+    font-size: unset;
+  }
 }
 
 .wp-block-image {


mleray:
Actually, I don't think that we want to unset the modifications from the styleguide, I think we want them to apply in the editor too since it should look like in the frontend (WYSIWYG) 🤔

lithrel:
Good point, but larger topic :hammer_and_wrench: 
Paragraph font size is controlled by an `inherit` editor normalization style in wordpress editor-styles.css, and I wonder if we could apply our own typography css file without rewriting our selectors by using [editor styles support](https://github.com/WordPress/gutenberg/blob/master/docs/designers-developers/developers/themes/theme-support.md#editor-styles).
I don't see mention of this in relevant tickets like https://jira.greenpeace.org/browse/PLANET-4117, or mention of `add_editor_style()` in our code, it looks like we integrate our styles to the editor through imports in editorStyle.scss, but this method doesn't automagically append selectors ? Maybe we should close this ticket and open a different one to discuss this ? (or there's one I've missed, or I'm mistaking :grimacing: )

mleray:
Good point @lithrel, maybe we can merge these changes for now to fix the inconsistencies in the editor and open a new ticket to investigate using [editor styles support](https://github.com/WordPress/gutenberg/blob/master/docs/designers-developers/developers/themes/theme-support.md#editor-styles)?

lithrel:
@mleray  yep, ok let's go like this :ok_hand:

mleray (approved):

PR #443, opened 2020-11-26
PLANET-5728: Carousel Gallery: Slider not working

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-5728

Alright, so apparently this was caused by the removal of the `useGalleryImages` effect here: 

https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/commit/735d0d10f7bab6a745cec6df0d684638258094b1

As that was causing aditional renders, `idRef.current` was available on subsequent renders, now that the endpoint is gone, the ref is initialized as `null` and that's what goes into the markup, so `document.getElementById` can't find it:

null-slides

So the change was to avoid relying on an ID ref and DOM methods and use a collection of refs instead.

### DIFF
--- a/assets/src/blocks/Gallery/GalleryCarousel.js
+++ b/assets/src/blocks/Gallery/GalleryCarousel.js
@@ -1,5 +1,4 @@
 import { useState, useEffect, useRef } from '@wordpress/element';
-import _uniqueId from 'lodash/uniqueId';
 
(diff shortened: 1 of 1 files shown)


lithrel (approved):
:+1:

Inwerpsel (approved):
Looks good :100:

PR #447, opened 2020-12-02
PLANET-5726 Hide password protected post on articles block

sagarsdeshmukh:
[JIRA 5726](https://jira.greenpeace.org/browse/PLANET-5726)

**Task :**
- Hide password protected posts on articles block


lithrel (approved):
:+1:

PR #448, opened 2020-12-02
Revert "PLANET-5142 Add twig filter(cf_img_url) on carousal header

comzeradd:
This reverts commit 143a2c31d7b458ca4bbac4bbf64000117e7cf5df.

Since [PLANET-5584](https://jira.greenpeace.org/browse/PLANET-5584) we no longer need that. Need to revert the code before disabling it on Cloudflare, in case some re-enabled the feature.


lithrel (approved):
:+1:

PR #456, opened 2020-12-10
PLANET-5642 Update ContentCovers heading and links

mleray:
See https://jira.greenpeace.org/browse/PLANET-5642

I've also removed the custom color that we had in this block for links, as content links should all follow the color defined in the styleguide.

### DIFF
--- a/assets/src/styles/blocks/Covers/ContentCovers.scss
+++ b/assets/src/styles/blocks/Covers/ContentCovers.scss
@@ -81,42 +81,20 @@
         text-decoration: underline;
       }
 
-      a {
-        color: $blue;
-      }
-
-      h4 {
-        font-size: $font-size-xxs;
-        line-height: 1.2;
-        font-weight: 500;
+      h5 {
         padding: 0;
         margin: 0;
 
-        @include medium-and-up {
-          font-size: 1.125rem;
-          font-weight: 500;
-        }
-
-        @include large-and-up {
-          font-size: $font-size-sm;
-        }
-
-        @include large-and-up {
-          font-size: 1.125rem;
+        > a {
+          color: $grey-80;
         }
       }
 
       p {
-        font-size: $font-size-sm;
         margin-bottom: 0;
 
-        @include large-and-up {
-          font-size: $font-size-xxs;
-          line-height: 1.6;
-        }
-
         @include x-large-and-up {
-          font-size: $font-size-xs;
+          font-size: $font-size-sm;
         }
 
         &.publication-date {
@@ -135,10 +113,6 @@
             font-size: $font-size-xs;
           }
         }
-      }
-
-      .blue-button {
-        width: 100%;
       }
     }
   }
--- a/templates/blocks/content_covers.twig
+++ b/templates/blocks/content_covers.twig
@@ -39,13 +39,16 @@
 									</div>
 									<div class="content-covers-block-information">
 										{% if ( post.post_title ) %}
-											<a
-												href="{{ post.link }}"
-												data-ga-category="Content Covers"
-												data-ga-action="Title"
-												data-ga-label="n/a">
-													<h4>{{ post.post_title|e('wp_kses_post')|raw }}</h4>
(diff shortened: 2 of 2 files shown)


assets/src/styles/blocks/Covers/ContentCovers.scss
```diff
@@ -81,16 +81,17 @@
...
-        color: $blue;
-      }
-
       h4 {
         font-size: $font-size-xxs;
         line-height: 1.2;
         font-weight: 500;
         padding: 0;
         margin: 0;
+        color: $grey-80;
```
Inwerpsel, line 90: In the Articles block the same is achieved by using inherit. Maybe makes sense here too, so we don't have to specify the text color again here? Also noticed that for articles we use `<h4><a>...</a></h4>`, while covers uses `<a><h4>...</h4></a>` :thinking: 
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/e5519762ac8d0d3f8759ed2de834e033f06ee66e/assets/src/styles/blocks/Articles.scss#L27-L27

mleray: `color: inherit;` doesn't work, it just applies the link color (also when inverting the tags)... In the Articles block it's not what changes the heading color, it's [this line](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/Articles.scss#L91) so it looks like we do need to specify the color again :/

mleray: However for Columns block we do also use this order `<h3><a>...</a></h3>` and then in the css we use the selector `h3 > a` to change the color, I think that should work here too and would add more consistency in the code 🤔

mleray: Just pushed a comment with those changes, please have a look @Inwerpsel 🙂 It also allows us to get rid of adding the hover styles to `h4`

Inwerpsel: Yeah there is in fact a significant difference between having the heading on the inside or outside of the link. If it's on the inside then the clickable area will span the whole width, while on the outside the clickable area will be only the text. So having the heading on the outside seems preferable in this case, to avoid unintentional clicks of the link :+1:

Inwerpsel: https://css-tricks.com/link-header-header-link/

mleray:
I've added a new commit changing the heading tag and the font sizes after Magali added new recommendations to the ticket

assets/src/styles/blocks/Covers/ContentCovers.scss
```diff
@@ -81,42 +81,20 @@
...
 
-      a {
-        color: $blue;
-      }
-
-      h4 {
-        font-size: $font-size-xxs;
-        line-height: 1.2;
-        font-weight: 500;
+      h5 {
```
Inwerpsel, line 84: Nice that we can achieve consistency by just removing this rules :+1: This does have potential to break child themes, so we probably should set up the test instance with CH, NL and Nordics child themes so we can check if they need to do changes for this.

mleray: Good point @Inwerpsel ! I'll check that out

comzeradd:
I opened a PR to fix the failing test, due to the markup change

https://github.com/greenpeace/planet4-base-fork/pull/109

assets/src/styles/blocks/Covers/ContentCovers.scss
mleray: Checked with NL and Nordics themes, seems fine 👌 However CH theme might need small tweaks so I've messaged Stefan about it!

Inwerpsel (approved):
Looks good :100: Maybe we can also mention the small improvement in the release notes that now only the actual text of the link is clickable and not the white space next to it.

PR #457, opened 2020-12-10
PLANET-5593 Hide Articles block if no articles to display

mleray:
See https://jira.greenpeace.org/browse/PLANET-5593


sagarsdeshmukh (approved):
Tested on local. LGTM 👍

PR #460, opened 2020-12-10
Planet 5616 - Fix block reports for Post with no title

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5616

---

- For Posts with no title it will just display `(no title)`, similar to what the native post list does.
- Links reversed. ID links to editing the Post. Title links to the frontend.
- Removed the custom query for Carousel Header, since we only have one style now.


sagarsdeshmukh (approved):
Tested on local, LGTM 👍

lithrel (approved):
:+1:

PR #461, opened 2020-12-10
PLANET-4590: Image block: Photo credit/copyright is not visible in the editor

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4590
Linked to: https://github.com/greenpeace/planet4-master-theme/pull/1259

> In the editor, when an Image is added to post, the caption is there but the credit/copyright text is missing.
> It's visible though in the frontend. That can be confusing for editors.

## Fix 

- Adding a hook to BlockEdit allows to append the image caption with the credits.
- Another PR on master-theme adds the `credit_text` in the response of the rest API.

### Why don't you use other filters

...like [`blocks.getBlockAttributes` or `blocks.getSaveElement`](https://github.com/WordPress/gutenberg/blob/master/docs/designers-developers/developers/filters/block-filters.md#filters) ? 

This fix uses `useSelect` to fetch image details and get the credits. 
Hooks in those filters gives an error `Error: Invalid hook call. Hooks can only be called inside of the body of a function component.`, I couldn't find another suitable filter for this that would accept hooks.

## Test

Both PR are testable on https://www-dev.greenpeace.org/test-pandora/
- Add an image block, the image should have some text on the "Credit" field of the Attachment details.
- Credit should appear in the image caption, in the editor and on the frontend.
- It is not possible to remove the credits

Screenshot from 2020-12-10 20-12-49
_Credits in the editor_

Screenshot from 2020-12-10 20-11-15
_Credits on the site_

### DIFF
--- a/assets/src/BlockFilters.js
+++ b/assets/src/BlockFilters.js
@@ -2,11 +2,13 @@ const { addFilter } = wp.hooks;
 
 import P4ButtonEdit from './components/p4_button/edit';
 import P4TableEdit from './components/p4_table/edit';
+import { ImageBlockEdit } from './components/Image/ImageBlockEdit';
 
 export const addBlockFilters = () => {
   addFileBlockFilter();
   addButtonBlockFilter();
   addTableBlockFilter();
+  addImageBlockFilter();
 };
 
 const addFileBlockFilter = () => {
@@ -63,3 +65,5 @@ const addTableBlockFilter = () => {
 
   addFilter('blocks.registerBlockType', 'planet4-blocks/filters/table', updateTableBlockEditMethod);
 };
+
+const addImageBlockFilter = () => addFilter('editor.BlockEdit', 'core/image/edit', ImageBlockEdit);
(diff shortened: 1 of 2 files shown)


sagarsdeshmukh (approved):
Tested on local. Looks good to me 👍

assets/src/BlockFilters.js
```diff
@@ -63,3 +65,7 @@ const addTableBlockFilter = () => {
 
   addFilter('blocks.registerBlockType', 'planet4-blocks/filters/table', updateTableBlockEditMethod);
 };
+
+const addImageBlockFilter = () => {
+  addFilter('editor.BlockEdit', 'core/image/edit', ImageBlockEdit);
+}
```
mleray, line 71: VSCode signals me a missing semicolon here 😉 Actually you can just write
```
const addImageBlockFilter = () => addFilter('editor.BlockEdit', 'core/image/edit', ImageBlockEdit);
```

mleray (approved):
Tested on local, looks good to me 👍 (just a tiny punctuation issue left to solve)

PR #462, opened 2020-12-10
Use a percentile width to show a bit of the third card

pablocubico:
Ref: https://github.com/greenpeace/planet4/issues/128

This change is to show a bit of the third card in a row of content covers as shown here:

image

The trick is to set the width to something under 50%, like 45%, so the first two cards would take 90% of the screen width.

Note:
The content covers styles are affecting some Bootstrap styles, this is something I would like to change in the Grid System issues, for now I'll leave it as it is. 
Also, this block still needs to be converted to WYSIWYG so we could revisit the layout there.


lithrel (approved):
:+1:

PR #463, opened 2020-12-11
PLANET-5522 Fix Columns Tasks block styles

mleray:
### Description
See https://jira.greenpeace.org/browse/PLANET-5522 (including the comments where Magali added more specs)

#### Notes
- After discussion with Magali I've also removed the links on images, since it's not something that we do in other P4 blocks and we already have title and CTA as links.
- I've also cleaned up the code a bit from unused classes (`blue-button`, `steps-action`) and unnecessary ones (`has-cta`)

### DIFF
--- a/assets/src/styles/blocks/Columns.scss
+++ b/assets/src/styles/blocks/Columns.scss
@@ -254,67 +254,25 @@
       }
 
       .step-info {
-        h5 {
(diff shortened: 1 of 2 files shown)


comzeradd (approved):
:100:

PR #464, opened 2020-12-11
Planet 5733 Move deprecated version to its own file, some tweaks

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5733
---
I started making a few changes to make the code a bit more manageable and in line with our other code before making further changes to it. All these changes could already be merged, as they should all be equivalent to the previous version. That way we can get these out of the way already and the next changes will be easier to review in isolation.


assets/src/blocks/Splittwocolumns/SplittwocolumnsEditor.js
```diff
@@ -30,20 +35,20 @@ export const SplittwocolumnsEditor = ({ attributes, setAttributes, isSelected })
...
+const renderView = ({attributes}) => <SplittwocolumnsFrontend {...attributes} />
 const renderEdit = ({attributes}, setAttributes) => {
   const charLimit = { title: 40, description: 400 };
   const params = {attributes, charLimit, setAttributes};
 
   return (
     <Fragment>
-      <InPlaceEdit {...params} />
-      <SidebarSettings  {...params} />
+      <SplittwocolumnsInPlaceEdit {...params} />
```
Inwerpsel, line 50: I removed the alias here as they make the code significantly harder to read and navigate. Another option would be to just call the component `InPlaceEdit` in the first place.

lithrel (approved):
a bit cleaner :grin: :+1: 
Tested locally on old and new blocks

PR #465, opened 2020-12-11
PLANET-5791 Fix last issues for Take Action boxout A/B test

mleray:
### Description
See https://jira.greenpeace.org/browse/PLANET-5791

- Move `cover-card-more` element outside of `cover-card-content` to fix the absolute positioning
- Use `bottom` instead of `top` positioning to adapt to potential longer translation strings for "Want to do more?"
- Add alt text to the image

### DIFF
--- a/templates/blocks/take_action_boxout.twig
+++ b/templates/blocks/take_action_boxout.twig
@@ -13,9 +13,9 @@
 				class="cover-card-overlay"
 				href="{{ boxout.link|default('#') }}" {{ boxout.new_tab and boxout.link ? 'target="_blank"' }}
 			></a>
-			<img src={{ boxout.image }} />
+			<img src={{ boxout.image }} alt="{{ boxout.image_alt }}" />
+			<div class="cover-card-more">{{ __( 'Want to do more?', 'planet4-blocks' ) }}</div>
 			<div class="cover-card-content">
-				<div class="cover-card-more">{{ __( 'Want to do more?', 'planet4-blocks' ) }}</div>
 				{% if ( boxout.campaigns ) %}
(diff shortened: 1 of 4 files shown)


templates/blocks/take_action_boxout.twig
```diff
@@ -14,8 +14,8 @@
 				href="{{ boxout.link|default('#') }}" {{ boxout.new_tab and boxout.link ? 'target="_blank"' }}
 			></a>
 			<img src={{ boxout.image }} />
```
Inwerpsel, line 16: We should probably add the `alt` attribute here, and provide an `srcset` so we don't have to serve oversized images.

Inwerpsel: Not necessarily here in this PR, but maybe it's nice to have out of the way before doing the AB test.

mleray: Just added the alt text 🙂

Inwerpsel:
> * Move `cover-card-more` element outside of `cover-card-content` to fix the absolute positioning
> * Use `bottom` instead of `top` positioning to adapt to potential longer translation strings for "Want to do more?"
> * Add alt text to the image

Can you add this info to the commit message too?

mleray:
> Can you add this info to the commit message too?
Done! 😄

Inwerpsel (approved):
Looks good form code POV :+1: but be sure to check during UAT if the new bottom values work well on all screen sizes.

mleray:
> Looks good form code POV 👍 but be sure to check during UAT if the new bottom values work well on all screen sizes.

I've checked all sizes on local and Mag checked them on the Optimize experiments too 🙂

PR #469, opened 2020-12-15
PLANET-5733 Provide srcset for split two columns

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5733
Deployed on https://www-dev.greenpeace.org/test-deimos/vaer-med/verdenshavene/

**Some context on srcset and sizes** can be found in the description of https://jira.greenpeace.org/browse/PLANET-5810. We'll probably do something similar like is done in this PR for all other images, though most will probably not be as straightforward as this PR. I used https://github.com/ausi/respimagelint for diagnosing the images, which seems to work reliably in most cases.

Before we made Split Two Columns wysiwyg, it was using an srcset in the twig template. In the newer version it doesn't use the srcset when saving in JS, though the backend does set it if an older version of the block is handled.

After investigating the sizes attribute, I found that it was relatively straightforward to provide it for this block, so I added it too [in the last commit](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/469/commits/4b130f2e410e2844c5cf75096a62f112c9e75f24).

I also converted ImageOrButton into a function component, since it doesn't use state this was a quick win ([ignore whitespace for review](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/469/files?diff=split&w=1#diff-8f3145db65a33fc48087cfc31177789516dbbc5ece88e95e450c05e3cbff134d) advised).

I also made `update_data` into a method of Base_Block and added a type hint so that the usage gets detected. (realized now that these last 2 changes were probably better done in a separate PR, but at least they're in separate commits :sweat_smile: ).

### DIFF
--- a/assets/src/blocks/Splittwocolumns/imageSizes.js
+++ b/assets/src/blocks/Splittwocolumns/imageSizes.js
@@ -1,0 +1,10 @@
+export default {
+  columnLeft: `
+  (min-width: 580px) 80vw,
+  100vw
+  `,
+  columnRight: `
+  (min-width: 580px) 75vw,
+  1vw
+  `,
+};
--- a/assets/src/components/ImageOrButton/ImageOrButton.js
+++ b/assets/src/components/ImageOrButton/ImageOrButton.js
@@ -1,56 +1,56 @@
-import {Component} from '@wordpress/element';
 import {Button} from '@wordpress/components';
 import {MediaUpload,MediaUploadCheck} from '@wordpress/block-editor';
 
-export class ImageOrButton extends Component {
-  constructor(props) {
-    super(props);
-  }
+export const ImageOrButton = props => {
+  const {__} = wp.i18n;
+  const {
+    disabled = false,
+    imageId,
+    imageUrl,
+    imgClass,
+    buttonLabel,
+    help,
+    title,
+    onSelectImage,
+  } = props;
 
-  render() {
-    const {__} = wp.i18n;
-    const disabled = this.props.disabled ?? false;
+  const getImageOrButton = (openEvent) => {
+    if (imageId) {
 
-    const getImageOrButton = (openEvent) => {
-      if ( this.props.imageId ) {
-
-        return (
-
-          <img
-            src={ this.props.imageUrl }
-            onClick={ openEvent }
-            className={ this.props.imgClass }
-          />
-
-        );
-      }
-      else {
-        return (
-          <div className='button-container'>
-            <Button
-              onClick={ openEvent }
-              className='button'
-              disabled={ disabled }>
-              { this.props.buttonLabel }
-            </Button>
-
-            <div>{ this.props.help }</div>
-          </div>
-        );
-      }
-    };
-
-    return <div className='ImageOrButton'>
-      <MediaUploadCheck>
-        <MediaUpload
-          title={this.props.title}
-          type='image'
-          onSelect={this.props.onSelectImage}
-          value={this.props.imageId}
-          allowedTypes={['image']}
-          render={ ({ open }) => getImageOrButton(open) }
+      return (
+        <img
+          src={ imageUrl }
+          onClick={ openEvent }
+          className={ imgClass }
         />
-      </MediaUploadCheck>
-    </div>;
-  }
+
+      );
+    }
+
+    return (
+      <div className='button-container'>
+        <Button
+          onClick={ openEvent }
+          className='button'
+          disabled={ disabled }>
+          { buttonLabel }
+        </Button>
+
+        <div>{ help }</div>
+      </div>
+    );
+  };
+
+  return <div className='ImageOrButton'>
+    <MediaUploadCheck>
+      <MediaUpload
+        title={title}
+        type='image'
+        onSelect={onSelectImage}
+        value={imageId}
+        allowedTypes={['image']}
+        render={ ({ open }) => getImageOrButton(open) }
+      />
+    </MediaUploadCheck>
+  </div>;
 }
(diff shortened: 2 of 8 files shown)


assets/src/components/ImageOrButton/ImageOrButton.js
```diff
@@ -1,56 +1,57 @@
...
+    onSelectImage,
+  } = props;
+
+  const getImageOrButton = (openEvent) => {
+    if (imageId) {
+
+      return (
+        <img
+          src={ imageUrl }
+          style={ { border: '1px solid grey', minWidth: '100px', minHeight: '100px', } }
```
Inwerpsel, line 23: Ah left this one in, but it's probably something we should tackle. If the ID doesn't exist, the image currently is not visible or clickable. We should probably catch this case and show an appropriate error message.

classes/blocks/class-splittwocolumns.php
```diff
@@ -78,10 +80,10 @@ public function __construct() {
 			self::BLOCK_NAME,
 			[
 				'editor_script'   => 'planet4-blocks',
-				'attributes'      => static::$attributes,
+				'attributes'      => self::ATTRIBUTES,
```
Inwerpsel, line 83: If you use `self` then this is treated as the string literal of the class name. `static` is only useful on a parent class if you want to use the member of a child class, and is evaluated runtime.

classes/blocks/class-splittwocolumns.php
```diff
@@ -128,20 +130,25 @@ public static function update_data( array $fields ): array {
...
-		$fields                      = array_filter( $fields );
-		$fields['issue_image_src']   = $issue_image_id ? wp_get_attachment_url( $issue_image_id ) : '';
-		$fields['issue_img_srcset']  = $issue_image_id ? wp_get_attachment_image_srcset( $issue_image_id, 'large' ) : '';
-		$fields['issue_image_title'] = $issue_image_id ? get_post_meta( $issue_image_id, '_wp_attachment_image_alt', true ) : '';
-		$fields['tag_image_src']     = $tag_image_id ? wp_get_attachment_url( $tag_image_id ) : '';
-		$fields['tag_img_srcset']    = $tag_image_id ? wp_get_attachment_image_srcset( $tag_image_id, 'large' ) : '';
-		$fields['tag_image_title']   = $tag_image_id ? get_post_meta( $tag_image_id, '_wp_attachment_image_alt', true ) : '';
+		$fields = array_filter( $fields );
+
+		$fields['issue_image_src']    = $issue_image_id ? wp_get_attachment_url( $issue_image_id ) : '';
```
Inwerpsel, line 139: Thank you WP alignment rule :roll_eyes:

classes/blocks/class-splittwocolumns.php
Inwerpsel: https://www.php.net/manual/en/language.oop5.late-static-bindings.php

classes/rest/class-rest-api.php
```diff
@@ -170,17 +173,26 @@ public static function endpoints(): void {
...
+						 */
 						$block_class = $blocks[ $block_name ] ?? null;
 
-						return rest_ensure_response(
-							$block_class && method_exists( $block_class, 'update_data' )
-								? $block_class::update_data( $request->get_params() )
-								: new WP_Error( 'error', 'Unknown block ' . ( $block_name ?? 'unspecified' ) )
-						);
+						if ( null === $block_class ) {
+							return new WP_Error( 'error', 'Unknown block ' . ( $block_name ?? 'unspecified' ) );
```
lithrel, line 186: We don't want  `rest_ensure_response()` on this one ? Just asking because we use it on the other WP_Error via `$response`; `rest_ensure_response` triggers one more action through `is_wp_error()`, but in the end returns the error as well :shrug:

Inwerpsel: It's not needed, as the first thing line of `rest_ensure_response()` checks if the argument is a `WP_Error` and returns the argument itself if so https://github.com/WordPress/WordPress/blob/master/wp-includes/rest-api.php#L563-L565

But it's probably more straightforward to have the catch clause return the error as well. :+1:

assets/src/blocks/Splittwocolumns/imageSizes.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  columnLeft: `
+  (min-width: 580px) 80vw,
+  100vw
+  `,
+  columnRight: `
+  (min-width: 580px) 75vw,
+  1vw
```
Inwerpsel, line 8: The right column is hidden under `580px`, so actually the image is not needed in that case. But the browser still loads it, and I didn't see an easy way to prevent it totally. I tried using `0` and `0vw`, but it still tried to load it. Setting it to `1vw` at least only makes it load the smallest possible image. Could be I'm overlooking a much better solution.

lithrel: Tried with lazy loading (which we should add anyway I think ?), still doesn't work :/ From [the old stuff](https://timkadlec.com/2012/04/media-query-asset-downloading-results/) I'm reading, the solutions would be to put the image as a css background, or condition the img tag insertion to the block width using [`useEffect`](https://stackoverflow.com/a/56011277)

Inwerpsel: Maybe we can review the block's specs first, since it's a bit weird that the right column is simply hidden below a certain screen size, I think most editors don't expect or realize this. Both columns seem equivalent in importance.

Inwerpsel: Lazy loading wouldn't help for this, as that only checks whether a given element is within or close to the viewport, not whether it's actually visible. We can still add it, though maybe better after we make the block static.

lithrel (approved):
:+1:

PR #471, opened 2020-12-16
PLANET-5489: Happypoint mailing list iframe toggle doesn't work

pablocubico:
Ref: https://github.com/greenpeace/planet4/issues/129

Hide the IFrame URL option if Mailing List IFrame is not set.

* Fix the logic in the frontend side.
* Add a migration to fix older blocks upon edition.
* Add a pre-publish checklist in case there is no URL and the option is set to On


assets/src/BlockEditorValidation.js
```diff
@@ -77,10 +77,13 @@ const PrePublishCheckList = () => {
     if ( !validations ) {
       return {block, invalids: []};
     }
+
+    console.log(block.name, block.attributes);
```
mleray, line 81: I'm pretty sure this can be removed 😛

mleray (dismissed):
I've tested this on local, it seems that there is a small issue when removing the iframe url. Instead of reverting to the default Subscribe Form URL, it keeps showing the iframe url that was previously entered and only when toggling the mailing list iframe then does the trick.
Also, I'm not sure that it is 100% clear what the help text refers to... Maybe we should ask Mag to have a look at the UX?

assets/src/blocks/Happypoint/MailingListIframeHelp.js
```diff
@@ -0,0 +1,23 @@
+import { HTMLSidebarHelp } from '../../components/HTMLSidebarHelp/HTMLSidebarHelp';
+
+const { __ } = wp.i18n;
+
+const MailingListIframeHelpTranslation =
+	__(
+		'By default this block uses the Happy point Subscribe Form URL from. ',
```
mleray, line 7: The dot after `from` should be removed

pablocubico: Nice catch! 🔍

assets/src/blocks/Happypoint/HappypointV1.js
```diff
@@ -0,0 +1,35 @@
...
+    id: {
+      type: 'number',
+    },
+    load_iframe: {
+      type: 'boolean',
+      default: false
+    }
+  },
+  isEligible({mailing_list_iframe, iframe_url}) {
+    return mailing_list_iframe === false && iframe_url != '';
```
mleray, line 25: `!=` should be `!==`, same below 🙂

pablocubico:
> I've tested this on local, it seems that there is a small issue when removing the iframe url. Instead of reverting to the default Subscribe Form URL, it keeps showing the iframe url that was previously entered and only when toggling the mailing list iframe then does the trick.
> Also, I'm not sure that it is 100% clear what the help text refers to... Maybe we should ask Mag to have a look at the UX?

Indeed, added a `UAT Needed` tag to the PR.

assets/src/blocks/Happypoint/HappypointBlock.js
```diff
@@ -26,6 +27,12 @@ export class HappypointBlock {
         },
         iframe_url: {
           type: 'string',
+          validation: ({ iframe_url, mailing_list_iframe }) => {
+            const isValid = mailing_list_iframe === false || iframe_url !== '';
+            const messages = ['Happypoint: "Use mailing list iframe" is set to "On" but there is no IFrame URL'];
```
mleray, line 32: Should we add translations to this message? 🤔

pablocubico: Yes! I'll do.

pablocubico:
> I've tested this on local, it seems that there is a small issue when removing the iframe url. Instead of reverting to the default Subscribe Form URL, it keeps showing the iframe url that was previously entered and only when toggling the mailing list iframe then does the trick.

@mleray good catch, I did some tests and I found out that entering an invalidad URL (thus causing a 404) won't update the iframe to show the 404 error. It still shows the last one that worked. So I'm working on a fix to have more control over the iframe.

planet-4:

:new_moon: [uranus](https://www-dev.greenpeace.org/test-uranus)

:watch: 2021.05.04 08:05:06

sagarsdeshmukh:
If we amend the commit message with ticket number would be nice. :)

Not specifically related to this ticket, but when we delete the existing "happy point background" image it won’t fall back to default setting from "Planet4 >> Default  Content  >> Default Happy Point Background Image"

I tested the PR changes on local and [test instance](https://www-dev.greenpeace.org/test-uranus/test-happy-point-block/), PR looks good to me 👍

sagarsdeshmukh (approved):
👍

sagarsdeshmukh:
> If we amend the commit message with ticket number would be nice. :)
> 
> Not specifically related to this ticket, but when we delete the existing "happy point background" image it won’t fall back to default setting from "Planet4 >> Default Content >> Default Happy Point Background Image"
> 
> I tested the PR changes on local and [test instance](https://www-dev.greenpeace.org/test-uranus/test-happy-point-block/), PR looks good to me 👍

Thanks @mleray, The "happy point background" image issue is fixed with latest change, bravo 🙌

PR #472, opened 2020-12-17
PLANET-5779 fix non-numeric error in carousel header

dpivo:
Ref: https://jira.greenpeace.org/browse/PLANET-5779

---

added is_numeric check to carousel header focus point setter to avoid non-numeric error


sagarsdeshmukh (approved):
👍

PR #478, opened 2021-01-14
Update README for WYSIWYG blocks

mleray:
See https://jira.greenpeace.org/browse/PLANET-5867

This is a draft to update the README file with the new process for building blocks (WYSIWYG). It's partly inspired by what is written in [Gitbook](https://support.greenpeace.org/planet4/tech/blocks), since the two reference each other. Speaking of which, maybe actually this whole section of the README should be moved there? Then we could just link it 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,26 +42,22 @@ Wordpress provides a single dependency for the whole build setup including:
 
 1. Create a new class that extends `Base_Block` ( `P4GBKS\Blocks\Base_Block` ) inside directory _classes/blocks_. The class name should follow naming convention, for example **Blockname** and its file name should be class-**blockname**.php.
 
-1. Implement its parent's class abstract method. In block's **constructor** you need to define the block's details (fields, schema) using `register_block_type` and in method **prepare_data()** you need to prepare the data which will be used for rendering.
+1. Implement its parent's class abstract method. In block's **constructor** you need to define the block's details (fields, schema) using `register_block_type` and add (required) empty method **prepare_data()**.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,26 +42,22 @@ Wordpress provides a single dependency for the whole build setup including:
...
 
-1. Add your new class name to the array inside Loader's ( `P4GBKS\Loader` ) constructor.
+    - **BlocknameBlock.js** should be a class that uses WordPress [registerBlockType](https://developer.wordpress.org/block-editor/developers/block-api/block-registration/) to define the block's attributes, schema, styles and `edit()`/`save()` functions. `edit()` function should return a React component that will be used for rendering the block in the editor. The
+    `save()` function should return a React component that will be used for rendering the block in the frontend.
 
-1. Create a new folder inside _react-blocks/src/blocks_ named after your block **Blockname** (first letter capital - rest lowercase). Create two new files inside that folder named **Blockname.js**  and **BlocknameBlock.js**.
+    - **BlocknameEditor.js** should be a class that defines a React component that implements `renderEdit()` and `renderView()`.
```
Inwerpsel, line 52: These methods are not strictly necessary, as we only use them internally. It's also technically not accurate anymore now don't use class components, the component cannot "implement" functions, since it's only a function and not a class.

I guess the only requirement is that it is a function component that gets the `attributes, setAttributes, isSelected props`?

mleray: Yeah I know, this was a first draft, definitely needs to be rewritten and these docs will very probably be moved somewhere else anyway when working on https://jira.greenpeace.org/browse/PLANET-5867

PR #482, opened 2021-01-19
PLANET-5883 Add missing commit message extract

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5883

Forgot to add this to the `create-zip` job. It's a bit weird that it happens in that job, but otherwise `request-instance` would need to check out the code too, which would add some time to the job.


lithrel (approved):
:+1:

PR #419, opened 2020-10-24
PLANET-4809: Convert Carousel Header block to WYSIWYG

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4809

As this is a rewrite of the previous block, the diff is not so useful. For reviewing, check the contents of `src/blocks/Carouselheader`.

PR #483, opened 2021-01-20
PLANET-4809: WYSIWYG Carousel Header 

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-4809
This PR also solves the following tickets:
- [PLANET-5198](https://jira.greenpeace.org/browse/PLANET-5198) (`https://` URL error message not showing)
- [PLANET-5770](https://jira.greenpeace.org/browse/PLANET-5770) (remove jQuery from CarouselHeader)

## 🆘  Feedback welcome! This introduces a few new issues/features, so please read and give the code a look. 

---

# Carousel Header

This is a refactor so the diff won't make much sense. Better look at the files in `blocks/CarouselHeader` directly.

It introduces a few new techniques:

### Per-block CSS & JS entry points

Extracted to: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/508

### Hydration

The initial state for the Carousel Header is now saved as static content. The `attributes` are passed through a `data-attributes` attribute.

To achieve this, we need to make sure the path to render the initial content is absolutely static (as you may know, you can't have side effects on `save()`). That's the reason most of the code for the Frontend ends up being in the top `CarouselHeaderFrontend` component. Very little behavior is delegated to child components.

### Server Side Rendering

There is an extra entry point for Server Side Rendering of a Carousel Header's static content. This file is built for NodeJS, there is an extra Webpack file and a `package.json` command for it, to create the SSR file run:

`npm run buildserver`

In essence, we have a Node script to render a static block, so where/when do we run it?

There is a sample inside the Carousel's PHP class using `exec()` to render the static content for old Carousels where the content is `null`. Of course this only works if Node is available in the environment. 

We could create a migration to update the current Carousels using that script. 

We could of course make React *render* (not *hydrate*) the frontend content in case is not static (i.e., is `null`), but in the case of the Carousel Header, being the first element in the page, I guess we should migrate it to a static version.

-----

### Some issues needing feedback

When we build WYSIWYG blocks, it's quite often to run into something that has an Editor version and a Frontend version, for example:

Frontend:
```jsx
<div class="some-fancy-border">
  <div class="another-wrapper">
     <h1>The Title</h1>
   </div>
</div>
```

Editor:
```jsx
<div class="some-fancy-border">
  <div class="another-wrapper">
     <RichText { ...etc } />
   </div>
</div>
```

In those cases I used very simple components as wrappers to be able to reuse the HTML:

Editor:
```jsx
<MyFancyBorder>
  <RichText { ...etc } />
</MyFancyBorder>
```

As this creates a few more levels in the component tree, some issues arise:

### Passing down behavior we need on the top:

`goToSlide`, `goToNextSlide`, and `goToPrevSlide` are defined in the Frontend class and passed down to the "dumb" childs as we need the content to be side-effect-free in most of the tree.

Maybe we could use the Context API to avoid passing down props.

### These new wrappers actually require something inside

Some of these wrappers (if not all of them) require having some kind of child.
We can't enforce a check on that right now, we could use `PropTypes` or `Typescript` to do so in the future.

In the case of the Carousel, these dumb wrapping components are specific to it and are not likely to be reused anywhere, but as we move forward with having a Design System, we'll probably need to adopt either `PropTypes` or `Typescript` in the near future.


webpack.config.js
```diff
@@ -6,12 +6,20 @@ const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
 const RemovePlugin = require('remove-files-webpack-plugin');
 const dashDash = require('@greenpeace/dashdash');
 
+const entryPoints = blockName => {
+  return {
+    [`${blockName}EditorScript`]: `./assets/src/blocks/${blockName}/${blockName}EditorScript.js`,
```
pablocubico, line 11: I made these names match the WP standard: `editor-script`, `editor-style`, `script`, and `style`.  But CamelCased.

assets/src/blocks/CarouselHeader/CarouselHeaderScript.js
```diff
@@ -0,0 +1,11 @@
+// Hydratable components
+import { CarouselHeaderFrontend } from './CarouselHeaderFrontend';
+
+const BLOCK_NAME = 'planet4-blocks/carousel-header';
+
+document.querySelectorAll( `[data-hydrate="${BLOCK_NAME}"]` ).forEach(
```
pablocubico, line 6: This won't be looping as the bundled one does (the `data-render` loop). But we could probably extract this into a function and import it on each frontend block script.

classes/class-loader.php
```diff
@@ -355,23 +355,23 @@ public function enqueue_editor_scripts( $hook ) {
...
-		wp_localize_script( 'planet4-blocks-script', 'p4ge_vars', $reflection_vars );
+		wp_localize_script( 'planet4-blocks-editor-script', 'p4ge_vars', $reflection_vars );
 
 		$reflection_vars = [
 			'home'  => P4GBKS_PLUGIN_URL . '/public/',
 			'pages' => $this->get_en_pages(),
 			'forms' => $this->get_en_forms(),
 		];
-		wp_localize_script( 'planet4-blocks-script', 'p4en_vars', $reflection_vars );
+		wp_localize_script( 'planet4-blocks-editor-script', 'p4en_vars', $reflection_vars );
```
pablocubico, line 365: I made this change to match WP convention.

classes/class-loader.php
```diff
@@ -391,7 +391,7 @@ public function enqueue_public_assets() {
 
 		// Include React in the Frontend.
 		self::enqueue_local_script(
-			'planet4-blocks-frontend',
+			'planet4-blocks-script',
```
pablocubico, line 394: Same as above, this change matches WP convention.

assets/src/blocks/CarouselHeader/migrate.js
```diff
@@ -0,0 +1,13 @@
...
+const { CarouselHeaderStaticContent } = require('./CarouselHeaderStaticContent');
+
+const attributes = JSON.parse(process.argv[2]);
+
+const renderedCarousel = ReactDOMServer.renderToString(React.createElement(
+  CarouselHeaderStaticContent,
+  attributes
+));
+
+console.log(renderedCarousel);
```
pablocubico, line 13: Strictly, I guess this should be `process.stdout.write` but for the PoC purpose `console` is ok, we still have to define what to do with all this.

classes/blocks/class-carouselheader.php
```diff
@@ -76,70 +105,59 @@ public function __construct() {
...
-	 * Enqueue required scripts for the editor.
-	 */
-	public function enqueue_editor_scripts() {
-		wp_enqueue_script( 'hammer', 'https://cdnjs.cloudflare.com/ajax/libs/hammer.js/2.0.8/hammer.min.js', [], '2.0.8', true );
+	public static function convert_to_static_block( $attributes ) {
+		$node_script = 'assets/build/CarouselHeaderMigrate-server.js';
+		$blocks_dir = P4GBKS_PLUGIN_DIR;
+		$attributes_json = json_encode( $attributes );
+
+		exec("cd ${blocks_dir} && node ${node_script} '${attributes_json}' 2>&1", $out, $err);
```
Inwerpsel, line 133: This probably works on local development because we add node to the php container there, but production php containers don't have node. Maybe we can create a temporary container which only runs the migration and does have node?

pablocubico: Right, it won't work unless the environment has node, This was just done as a PoC, but I think it's probably better to run a Migration and convert the old content instead of converting the block "in-place" here 🤔...

assets/src/blocks/CarouselHeader/ArrowsAndIndicators.js
```diff
@@ -0,0 +1,24 @@
+import { Arrows } from './Arrows';
+import { Indicators } from './Indicators';
+
+export function ArrowsAndIndicators({
```
Inwerpsel, line 4: Seems more straightforward to use `Arrows` and `Indicators` directly? That's already a bit less props to pass on.

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const slidesRef = useRef([]);
+  const containerRef = useRef(null);
+  const { currentSlide, sliding, goToSlide } = useSlides(slides, slidesRef);
+
+  const lastSlide = slides.length - 1;
+  const goToNextSlide = () => goToSlide(currentSlide === lastSlide ? 0 : currentSlide + 1);
+  const goToPrevSlide = () => goToSlide(currentSlide === 0 ? lastSlide : currentSlide - 1);
+
+  useEffect(() => {
```
Inwerpsel, line 18: Big effects really make it hard to see what a component is doing. Can this either be broken up, or converted to one or more custom hooks?

assets/src/blocks/CarouselHeader/CarouselHeaderScript.js
Inwerpsel: Indeed seems better to extract already :+1:

assets/src/blocks/CarouselHeader/EditableBackground.js
```diff
@@ -0,0 +1,53 @@
...
+  setAttributes,
+  focalPoints,
+  children,
+}) => {
+  function onImageChange(index, value) {
+    let slidesCopy = JSON.parse(JSON.stringify(slides));
+    if (null !== value) {
+      slidesCopy[index].image = value;
+    } else {
+      slidesCopy[index].image = null;
```
Inwerpsel, line 19: Set it to the value, unless the value is `null`, then set it to `null` instead of `null` :thinking:

assets/src/blocks/CarouselHeader/EditableBackgroundActions.js
```diff
@@ -0,0 +1,21 @@
...
+  Dashicon,
+  Button,
+} from '@wordpress/components';
+
+export const EditableBackgroundActions = ({ image_url, onRemove, onChange }) => (
+  <div className="carousel-header-top-image-actions">
+    {image_url && onRemove && (
+      <Button isSecondary className="carousel-header-image-upload-remove" onClick={ev => {
+          onRemove();
+          ev.stopPropagation()
```
Inwerpsel, line 11: Why do we need `stopPropagation`?

Inwerpsel (commented):
Already dumping these various comment before they bunch up too much :smile: Not too sure about the hammer + react mix, that might get finicky.

assets/src/blocks/CarouselHeader/useSlides.js
```diff
@@ -0,0 +1,65 @@
+import { useState } from '@wordpress/element';
+
+export const useSlides = (slides, slidesRef) => {
```
Inwerpsel, line 3: The logic of this hook seems to be copied from https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/b2ad5d4448aa089a887faa2b7d85af95e781812e/assets/src/blocks/Gallery/GalleryCarousel.js#L10-L14

Probably better to make that component use the hook too while creating it, then we won't have duplicate logic and can actually improve things in one place.

Inwerpsel: If both components can't use the same hook, that probably means the hook has too many responsibilities.

classes/blocks/class-base-block.php
```diff
@@ -94,4 +94,70 @@ protected function is_rest_request() {
 	public static function update_data( array $fields ) {
 		throw new NotImplemented( 'Method update_data is not implemented for ' . static::class );
 	}
+
+	public function as_hydratable_block( $attributes, $content ) {
```
Inwerpsel, line 98: Should be `static function`

classes/blocks/class-carouselheader.php
```diff
@@ -22,17 +22,34 @@ class CarouselHeader extends Base_Block {
...
 	 */
-	const BLOCK_NAME = 'carousel_header';
+	const BLOCK_NAME = 'carousel-header';
+
+	/**
+	 * Block category.
+	 *
+	 * @const string BLOCK_CATEGORY.
+	 */
+	const BLOCK_CATEGORY = 'planet4-blocks';
```
Inwerpsel, line 32: I don't think there's a reason to split this across 2 constants, as the block name is always Namespace + Name, we never need to use these separately. Also "category" is confusing, since we also have a category on the front end.
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/b2ad5d4448aa089a887faa2b7d85af95e781812e/assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js#L46

Inwerpsel: This category happens to be the same string as the Namespace, but it could be another one like here https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/b2ad5d4448aa089a887faa2b7d85af95e781812e/assets/src/blocks/Accordion/AccordionBlock.js#L40-L40

assets/src/blocks/CarouselHeader/ArrowsAndIndicators.js
pablocubico: Indeed. Sometimes you don't know where to stop, my biggest fear in JSX is creating a `<Div>` component. 😛

classes/blocks/class-carouselheader.php
pablocubico: Ah, I just needed to split the name so I could automatically map it into something like "CarouselHeader" in the filesystem. I could split it when I need to, but I also thought separating them would be more accurate to not add a prefix to the name, as the "name" of the block is `carousel-header` and not `planet4-blocks/carousel-header`.

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
Inwerpsel: Or it might not be needed if we look into https://github.com/pmndrs/react-use-gesture instead.

classes/blocks/class-carouselheader.php
Inwerpsel: I previously glossed over the fact that the name wasn't prefixed before, thought you had split it, my bad :facepalm:

Inwerpsel: Isn't this currently breaking some twig templates? https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/6e7af53bffbd965f1ecf32906a606900224f752f/classes/blocks/class-base-block.php#L42

assets/src/blocks/CarouselHeader/Caption.js
```diff
@@ -0,0 +1,64 @@
...
+    setAttributes({slides: slidesCopy});
+  }
+
+  return <div className="carousel-caption">
+    <div className="caption-overlay"></div>
+    <div className="container main-header">
+      <div className="row">
+        <div className="col">
+          <div className="carousel-captions-wrapper">
+            <RichText
```
mleray, line 17: Can you please remove the formatting options from the RichText components, by adding `allowedFormats={[]}`? In preparation for https://jira.greenpeace.org/browse/PLANET-5780 🙂

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
pablocubico: Indeed Hammer it's a bit duct-taped in this PR. 

Last week we talked about implementing a gesture lib and an animations lib. As suggested before, we could use `react-use-gesture` to replace Hammer and `react-spring` to provide some "spring" animations. Another option would be to use `framer-motion`, which combines a bit of both, it's mainly an animation library but it provides support for some gestures.

Now after some tests, I think maybe adding other libs could be overkill unless there is a real need for it. I think both libraries are aimed at a different type of app, mainly SPAs with tons of animations and page-to-page transitions. In that case, using some lib to orchestrate things would make a lot sense.

Currently, our only real need would be to replace Hammer with something, we can just use our own custom hook with some event listeners to add swipe recognition to it. With no extra dependencies/lock-in.

As for the options, I found a few quirks in both of them that made them look overkill for the current use case.

With `useGesture`, I found the drag/swipe to be a bit too specific, as it expects the swipe to happen at some given velocity and distance. I tried a few different thresholds for speed and distance, but in the end, the vanilla version responds faster.

I made two PoC for the `Gallery/Carousel` block, this the [useGesture version](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/488) and this is a [vanilla swipe](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/489). 

As for `framer-motion`, it's a bit big, 29kb (though they claim it can be reduced to ~23kb), and switching to it now would mean converting a lot of CSS into motion variants without any performance improvements, in fact, I tested demos of Framer Motion in a real device and some were quite laggy. So I think it's a bit out of scope for this ticket.

One thing I didn't like about the `framer-motion` documentation/samples/community was that it was pretty tied to their prototyping product (Framer). I sensed a bit of a focus issue there...

Spring animations and gesture recognition can get a bit heavy on CPU very quickly, depending on the kind of animation you're trying to build. It depends a bit on the properties you're trying to animate, some are GPU accelerated (e.g.: the transforms, translate, scale, etc.) and some not. Add dampening, stiffness and other variables to the mix and it can get quite complex very quickly.

So, in conclusion:
* I think it's good to avoid mixing DOM-based libs with React ones. So I would vote for replacing Hammer with something else.
* As for animations, a lot can be done with CSS, with no extra deps, no performance hit. So I think unless we get a spec asking for something that screams for an animation library, CSS is a great option.
* By default, I would refrain from adding extra dependencies as much as we can, not just because of the build size (we can load things only when we need them) but also because it adds more lock-in, and once you adopted something like Motion, it's going to be there forever, is not trivial to change afterwards.

Note: I was initially in favor of adding these libs in the [old PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/419) (see version history for the description), changed my mind at some point when we started to trim libs & optimize performance.

pablocubico: ccing @comzeradd to keep track of the choices here

Inwerpsel: 29KB is indeed not great, but to put it into perspective, this is for the home page of international
Screenshot from 2021-02-03 19-13-03

comzeradd: @pablocubico do you think we can reach a minimum functional swipe support with just css? If so, that would be sufficient for now. It was already a bit of overkill to have a dedicated library (hammer) just for swiping a carousel :grin:

Also have you checked how bootstrap 5 has implemented [their Carousel component](https://getbootstrap.com/docs/5.0/components/carousel/), now that they dropped jquery?

pablocubico: > @pablocubico do you think we can reach a minimum functional swipe support with just css? If so, that would be sufficient for now. It was already a bit of overkill to have a dedicated library (hammer) just for swiping a carousel 😁
> 
> Also have you checked how bootstrap 5 has implemented [their Carousel component](https://getbootstrap.com/docs/5.0/components/carousel/), now that they dropped jquery?

@comzeradd yes, I made a PR with a vanilla swipe implementation and isolated the Hammer effect here to replace it. I haven't checked the new Bootstrap Carousel deeply, I'll take a look as it could affect the Gallery Carousel, which is still using styles from Bootstrap. After this PR is merged, I think we'll be able to reuse the slides logic in the Carousel Gallery and remove the Bootstrap Carousel from the build. 


> 29kB is indeed not great, but to put it into perspective, this is for the all scripts on the home page of international
> Screenshot from 2021-02-03 19-13-03

@Inwerpsel well it's not just downloading 29kb, it's also parsing & executing those 29kb... it comes with a CPU cost specially in mobile. I think the pure CSS solution will probably perform better. After some thought, I think using Framer in a non-animation-intensive product feels a bit overkill, like "shooting a fly with a cannon".

Inwerpsel: @pablocubico Makes sense, we can see how far we get with pure CSS for the limited animations and gestures we have on the front end.

> well it's not just downloading 29kb, it's also parsing & executing those 29kb

My point was not that it's only the data transfer. All 457kB we already have also needs to be parsed and executed.

I guess we do have a use case for an animation library in the back end, where it can improve the user feedback when interacting with components. There the size of the lib definitely pales in comparison with what's already loaded.

Hopefully we'll soon have a reliable way to measure the actual impact on performance :crossed_fingers:

comzeradd:
Not sure if the test instance is up to date with latest code, but two minor issues I see:
1. the previous slide arrow doesn't get a pointer cursor.
2. when I click next slide it adds a hash on the url.

The page is also not editable. I'm getting an exception and the page stays blank.

assets/src/blocks/CarouselHeader/useCarouselHeaderImages.js
```diff
@@ -0,0 +1,17 @@
+import { useSelect } from '@wordpress/data';
+
+export const useCarouselHeaderImages = slides => {
+  return useSelect(select => {
+    let slidesCopy = slides ? JSON.parse(JSON.stringify(slides)) : [];
```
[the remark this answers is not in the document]
pablocubico: Adopted with a few minor tweaks. 👍

classes/blocks/class-carouselheader.php
pablocubico: This was updated in the [split blocks PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/508), in a "let's call a spade a spade" basis: `BLOCK_NAME` is just the name, `BLOCK_NAMESPACE_PREFIX` is the namespace. I needed to separate them to be able to use the name for a few defaults without needing to split it.

pablocubico:
@Inwerpsel @mleray @comzeradd I rebased this branch to the latest master.

> the previous slide arrow doesn't get a pointer cursor.
> when I click next slide it adds a hash on the url.
The issue with the previous slide arrow seems to be related with the External Link script, I can see the external link icon in the arrow and the `target` set to `_blank`. We probably want those arrows to be `<button>`s instead of `<a>`nchors, that would also fix the hash issue.

Other than that I addressed most of the comments (if not all?), and the most important issues are still the ones I mentioned in the PR's description about using hydration, migrating old content and removing the temporary Node call.

classes/blocks/class-carouselheader.php
```diff
@@ -48,6 +65,18 @@ public function __construct() {
 								'image'            => [
 									'type' => 'integer',
 								],
+								'image_src'        => [
+									'type' => 'integer',
+								],
+								'image_srcset'     => [
+									'type' => 'integer',
+								],
+								'image_sizes'      => [
```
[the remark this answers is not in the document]
mleray: I'm actually not sure that we need these image properties at all, since we are using the CSS `background-image` 🤔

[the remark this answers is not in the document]
Inwerpsel: Wouldn't that mean we'll serve either massively oversized images to mobile, or pixelated images to desktop? :thinking:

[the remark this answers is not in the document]
mleray: Yeah I'm not sure how that impacts performance... I think we are doing this because of the focal point

planet4-gutenberg-blocks.php
```diff
@@ -107,7 +107,7 @@
 */
 
 const POST_BLOCK_TYPES = [
-	'planet4-blocks/accordion',
+	'planet4-blocks-beta/accordion',
```
Inwerpsel, line 110: Probably not intended to change this?

classes/class-loader.php
```diff
@@ -405,7 +405,7 @@ public function enqueue_public_assets() {
 		// Sets translated strings for a JS script.
 		wp_set_script_translations( 'planet4-blocks-script', 'planet4-blocks', P4GBKS_PLUGIN_DIR . '/languages' );
 
-		if ( self::can_include_theme_editor() ) {
+		if ( false && self::can_include_theme_editor() ) {
```
Inwerpsel, line 408: You can also turn off the option :grin:

mleray:
Just added the WYSIWYG version to beta blocks; you can check out both on [this page](https://www-dev.greenpeace.org/test-rhea/carousel-header-new-wysiwyg/) for instance 🙂

classes/blocks/class-carouselheader.php
```diff
@@ -22,17 +22,29 @@ class CarouselHeader extends Base_Block {
...
 	 */
 	public function __construct() {
+		add_action( 'init', [ $this, 'register_carouselheader_block' ] );
+	}
+
+	/**
+	 * Register CarouselHeader block.
+	 */
+	public function register_carouselheader_block() {
+		wp_enqueue_script( 'hammer', 'https://cdnjs.cloudflare.com/ajax/libs/hammer.js/2.0.8/hammer.min.js', [], '2.0.8', true );
```
Inwerpsel, line 38: This enqueue was previously in `enqueue_editor_scripts` function. If we add it here, it will be added to each page on the front end.

templates/blocks/carousel_header_full-width-classic.twig
```diff
@@ -4,7 +4,7 @@
 {% endmacro %}
 
 <section class="block block-header block-wide carousel-header_full-width-classic">
-	<div id="carousel-wrapper-header"
+	<div id="old-carousel-wrapper-header"
```
Inwerpsel, line 7: Didn't check everything, but I didn't see something in the new version that would collide with this. I guess we can preserve the `id`?

mleray: The issue is that we use the id to add some things in the old version (indicators, arrow events, etc) so if it's the same id in the new version then there will be issues and the block won't behave as expected. Of course that would happen only if both versions of the block are on the same page which should never be the case, but I still think we should fix it just in case since it is technically possible. However maybe I can change the new block's id instead of the old one? 🤔

Inwerpsel: Does the new version still use this id for something? I couldn't locate the same code (that used the id) as the previous version in the new version, or new code that uses it. But I might have missed it, this PR is massive :smile:

mleray: That's a good point actually, the id is removed in the new version 😅 I guess something else fixed the issues I was seeing, I'll double-check!

mleray: Removed the id change! 😄

planet-4:

:new_moon: [rhea](https://www-dev.greenpeace.org/test-rhea)

:watch: 2021.04.28 10:48:10

Inwerpsel (approved):
Tested locally and on https://www-dev.greenpeace.org/test-rhea/carousel-beta-test/ :100: So much work went into this, great to finally merge it :raised_hands:  cc @pablocubico

PR #484, opened 2021-01-22
PLANET-5901: Lazy load gallery images

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5901

This ensures a lot less images are loaded if this block is present below the fold.


mleray (approved):
Looks good to me 👍

PR #486, opened 2021-01-27
PLANET-5523 - Covers block WYSIWYG conversion

mleray:
## Description

- Remove `load_more.js` (jQuery): [PLANET-5764](https://jira.greenpeace.org/browse/PLANET-5764)
- Convert Covers block to WYSIWYG: [PLANET-5523](https://jira.greenpeace.org/browse/PLANET-5523)

Can be tested on [this page](https://www-dev.greenpeace.org/test-iocaste/all-the-covers/)

Corresponding base-fork PR to update acceptance tests: https://github.com/greenpeace/planet4-base-fork/pull/119


classes/rest/class-rest-api.php
```diff
@@ -252,6 +253,26 @@ public static function endpoints(): void {
...
+			'/get-covers',
+			[
+				[
+					'permission_callback' => static function () {
+						return true;
+					},
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => static function ( $fields ) {
+						$covers = Covers::get_covers( $fields );
+						return rest_ensure_response( $covers );
```
Inwerpsel, line 271: Can you set this data the same way we did for the Gallery block? https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/45ab1881ae40ccb5e88ae62c8ca9c7a6e63ef6e6/classes/blocks/class-gallery.php#L46-L52

mleray: Done!

assets/src/blocks/Covers/CoversBlock.js
```diff
@@ -1,130 +1,99 @@
...
+        name: 'campaign',
+        label: getStyleLabel(
+          __('Campaign covers', 'planet4-blocks-backend'),
+          __('Campaign covers pull the associated image and hashtag from the system tag definitions', 'planet4-blocks-backend'),
+        ),
+      }
+    ],
+    deprecated: [
+      {
+        attributes,
```
Inwerpsel, line 91: If we add the same `attributes` as the current version then removing, changing or adding any attribute in the current version will cause the deprecated entry to stop working as intended. If you move each deprecated entry to its own file, you can copy the attributes and they're still out of view.

Inwerpsel: For example like we did for Media https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/wysiwyg-vars/assets/src/blocks/Media/MediaBlock.js#L56-L58

admin/js/editor.js
```diff
@@ -1,19 +1,28 @@
...
     name: 'donate',
     label: 'Donate'
   });
-  wp.blocks.unregisterBlockStyle('core/button', 'outline');
-  wp.blocks.unregisterBlockStyle('core/button', 'fill');
+  unregisterBlockStyle('core/button', 'outline');
+  unregisterBlockStyle('core/button', 'fill');
+
+  // Remove Take Action and Campaign covers styles for Covers block in campaigns
+  const postType = document.querySelector('form.metabox-base-form input#post_type').value;
```
Inwerpsel, line 23: I think you can also do `wp.data.select('core').getPostType()`

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -0,0 +1,45 @@
+const { __ } = wp.i18n;
+
+export const CampaignCovers = ({ covers, covers_view, row, loadMoreCovers }) => {
+  const rowAmount = 3;
+  const showLoadMore = (covers.length > rowAmount && covers_view === '1') ||
+    (covers.length > (rowAmount * 2) && covers_view === '2');
```
Inwerpsel, line 6: We should probably replace these numbers with an actual descriptive name, I don't see why it would need to be a number. It's a bit more work to do it now, since we would need to write a function for the `deprecated` entry to map the old numbers to the new string, but it seems worth it, since it will make the block comment a lot easier to understand.

Inwerpsel: Maybe a first step would be to replace all these numbers with constants that have the number as a value, so we can at least understand the code :sweat_smile:

Inwerpsel (dismissed):
It's really hard to keep track of which number means which type. Can you replace these with constants? :pray: We can eventually change those constant values so that also the block comments are not cryptic.

assets/src/blocks/Covers/CampaignCovers.js
mleray: These actually kind of make sense since they are the amount of rows: either 1, 2 or all. So the only possible value that doesn't really make sense is '3', which stands for 'all rows'... But I can do something similar as what I did for the `cover_type` attribute which was also a mysterious number and now has constants?

Inwerpsel: Right, I first thought it contained a type, since it's a string and not an int. Maybe better to make it an int and use it directly? We could call it `initialRowsLimit`, and wouldn't need a special value for the "no limit" case, we could instead leave it empty in that case.

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -0,0 +1,45 @@
+const { __ } = wp.i18n;
+
+export const CampaignCovers = ({ covers, covers_view, row, loadMoreCovers }) => {
+  const rowAmount = 3;
```
Inwerpsel, line 4: This could be read as amount per row, or amount of rows. If we call it`amountPerRow` it won't have that ambiguity.

assets/src/blocks/Covers/CampaignCovers.js
Inwerpsel: Then the equation would be much simpler:
```
const showLoadMore = !!initialRowsLimit && covers.length > (amountPerRow * initialRowsLimit)

File: assets/src/blocks/Covers/CampaignCovers.js [hunk 6]
```diff
@@ -0,0 +1,45 @@
...
+              <a
+                href={href}
+                data-ga-category='Campaign Covers'
+                data-ga-action='Image'
+                data-ga-label='n/a'
+                aria-label={__('Check our campaign about ' + name, 'planet4-blocks')}
+              >
+                <div className='thumbnail-large'>
+                  {image && image[0] &&
+                    <img loading='lazy' src={image[0]} alt={alt_text} />
```
Inwerpsel, line 25: Images always need to have an `srcset`

Inwerpsel: After you added an `srcset` you can get the `sizes` attribute with this bookmarklet https://ausi.github.io/respimagelint/ . You can right click and "copy link adress" on the big button, and add it as a bookmark in your browser. Then you can open a page with a covers block and click the bookmarklet. It will in the background resize the current page to all possible screen sizes and send you to a report page where it suggests the `sizes` attribute. Probably won't work in a private tab (it doesn't for me), but regular Chrome it should.

admin/js/editor.js
mleray: That returns `undefined`

assets/src/blocks/Covers/CampaignCovers.js
mleray: Updated! 🙂

assets/src/blocks/Covers/CampaignCovers.js
mleray: Added! 🙂

admin/js/editor.js
Inwerpsel: Right, it's `wp.data.select('core/editor').getCurrentPostType()` instead

mleray: Ah nice, that works! Updated 🙂

assets/src/blocks/Covers/deprecated/coversV1.js
```diff
@@ -0,0 +1,39 @@
...
+    cover_type: {
+      type: 'string',
+    },
+  },
+  isEligible(attributes) {
+    return attributes.covers_view;
+  },
+  migrate( { covers_view, ...attributes } ) {
+    attributes.initialRowsLimit = covers_view === '3' ? 0 : Number(covers_view);
+    delete attributes.covers_view;
```
Inwerpsel, line 34: No need to delete it, it's already the other attributes because of `{ covers_view, ...attributes }`

assets/src/blocks/Covers/deprecated/coversV1.js
```diff
@@ -0,0 +1,39 @@
...
+    covers_view: {
+      type: 'string',
+      default: '1'
+    },
+    cover_type: {
+      type: 'string',
+    },
+  },
+  isEligible(attributes) {
+    return attributes.covers_view;
```
Inwerpsel, line 30: Maybe we should start adding a version attribute :thinking: Could make it easier once we start having more versions. Then we can know the version from the block comment itself, without needing to check all the `isEligible` functions.

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -0,0 +1,50 @@
...
+  const showLoadMore = !!initialRowsLimit && covers.length > amountPerRow * initialRowsLimit;
+
+  return (
+    <div className='container'>
+      <div className='thumbnail-largeview-container'>
+        {covers.map((cover, index) => {
+          const { href, image, alt_text, name, src_set, sizes } = cover;
+          const hideCover = !!initialRowsLimit && index >= row * amountPerRow;
+          return (
+            <div key={name} className={`campaign-card-column ${hideCover ? 'hidden' : ''}`}>
```
Inwerpsel, line 14: You could use `classnames` here. We're not using it on the front end yet, but it's really small so no significant effect on our bundle size.
```
classNames('campaign-card-column', {hidden:hideCover})

File: assets/src/blocks/Covers/ContentCovers.js [hunk 10]
```diff
@@ -0,0 +1,87 @@
+import { useState, useLayoutEffect } from '@wordpress/element';
+
+const { __ } = wp.i18n;
+
+export const ContentCovers = ({ covers, initialRowsLimit, row, loadMoreCovers }) => {
+  const [amountPerRow, setAmountPerRow] = useState(4);
+
+  const updateRowAmount = () => setAmountPerRow(window.innerWidth >= 576 && window.innerWidth < 992 ? 3 : 4);
```
Inwerpsel, line 8: A bit more readable if you move `window.innerWidth >= 576 && window.innerWidth < 992` into a function. Something like
```js
const updateRowAmount = () => setAmountPerRow(isMediumWindow ? 3 : 4);

Comment (Inwerpsel) [reply assets/src/blocks/Covers/ContentCovers.js:8]: I guess you could then put `setAmountPerRow(isMediumWindow ? 3 : 4)` directly into `useLayoutEffect`

File: assets/src/blocks/Covers/CoversEditor.js [hunk 11]
```diff
@@ -0,0 +1,153 @@
...
+      <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
+        <SelectControl
+          label='Rows to display'
+          value={initialRowsLimit}
+          options={rowLimitOptions}
+          onChange={value => toAttribute('initialRowsLimit')(Number(value))}
+        />
+
+        {!posts.length &&
+          <div>
```
Inwerpsel, line 38: Do we still need these divs around the component now it's in the sidebar? :thinking:

assets/src/blocks/Covers/CoversEditor.js
```diff
@@ -0,0 +1,153 @@
...
+const { RichText } = wp.blockEditor;
+const { __ } = wp.i18n;
+
+const renderEdit = (attributes, toAttribute) => {
+  const { initialRowsLimit, posts, tags, cover_type, post_types } = attributes;
+
+  const rowLimitOptions = [
+    { label: __('1 Row', 'planet4-blocks-backend'), value: 1 },
+    { label: __('2 Rows', 'planet4-blocks-backend'), value: 2 },
+    { label: __('All rows', 'planet4-blocks-backend'), value: 0 },
```
Inwerpsel, line 24: Using 0 for this is a bit in accurate :sweat_smile: If we make the value `null`, it unsets the attribute, which makes sense, we're removing the limit.

Inwerpsel (commented):
Some more comments, continuing with the rest, think we're almost there :ok_hand: :ok_hand:

assets/src/blocks/Covers/CoversEditor.js
mleray: For this one we do actually, because of the help paragraph below... But I'll remove the one around `PostTypeSelector`!

assets/src/blocks/Covers/deprecated/coversV1.js
mleray: Added! 🙂

mleray:
> It's really hard to keep track of which number means which type. Can you replace these with constants? 🙏 We can eventually change those constant values so that also the block comments are not cryptic.

I've replaced the "string numbers" by actual strings for `cover_type`, please have a look 🙂

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -0,0 +1,51 @@
...
+                    />
+                  }
+                  <span className='yellow-cta'><span aria-label='hashtag'>#</span>{name}</span>
+                </div>
+              </a>
+            </div>
+          );
+        })}
+      </div>
+      {showLoadMore && (row * amountPerRow) < covers.length &&
```
Inwerpsel, line 40: Doesn't this kinda check the same thing twice? :thinking:

mleray: Right, since `row` is at first set to `initialRowsLimit` this is unnecessary 🙂 Fixed!

assets/src/blocks/Covers/CoversEditor.js
mleray: If I change this to use `null` instead, there is a warning in the console when selecting that value:
```
Received NaN for the `value` attribute. If this is expected, cast the value to a string.
```
...and it's no longer possible to select "All rows"!

assets/src/blocks/Covers/ContentCovers.js
mleray: I've moved the window width calculations (here and in TakeActionCovers) to separate function (i.e. `isMediumWindow`), but I still can't directly use it in `useLayoutEffect` as it needs to happen on window resize... I have however changed from `useLayoutEffect` to `useEffect` as it seems recommended to do so whenever possible!

assets/src/blocks/Covers/CoversEditor.js
Inwerpsel: That's because `onChange` is doing `Number(value)`, I'm not sure if we even need to do that, since the values are already numbers. But if we do need to cast them, we could make an exception for the "all rows" case.

Inwerpsel (commented):
Still had some comments pending, just some small things. I'll continue reviewing now. I expect the only time consuming thing left is to look into how the images load if we render markup front end. It will only start fetching them once React has rendered the component for the first time. We should check the impact of this and asses whether we should add a preload tag to the initial page html. A preload tag will make the images load as fast as though they were included in the initial html.

assets/src/blocks/Covers/CoversEditor.js
mleray: I've tested and you do need the cast since `SelectControl` converts the values to strings... And if we set the value to `null`, then it takes the label instead (`All rows`) for the value and leads to the warning in the console that I previously wrote. Now, I can remove the cast for this specific case and try to set `null` but in that case I get another console warning:
```
Warning: `value` prop on `select` should not be null. Consider using an empty string to clear the component or `undefined` for uncontrolled components.
```

assets/src/blocks/Covers/getCoversClassName.js
```diff
@@ -0,0 +1,19 @@
+import { COVER_TYPES } from './Covers';
+
+export const getCoversClassName = (cover_type, initialRowsLimit) => {
```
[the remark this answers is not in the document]
mleray: I doubt it, there are quite a lot of conditions involved here! 🤔

classes/blocks/class-covers.php
```diff
@@ -195,7 +211,7 @@ private function filter_posts_by_ids( $fields ) {
 			];
 
 			// If cover type is take action pages set post_type to page.
-			if ( isset( $fields['cover_type'] ) && '1' === $fields['cover_type'] ) {
+			if ( isset( $fields['cover_type'] ) && 'take-action' === $fields['cover_type'] ) {
```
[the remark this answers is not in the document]
mleray: Updated! 🙂

assets/src/blocks/Covers/getCoversClassName.js
```diff
@@ -0,0 +1,19 @@
...
+  const isContentType = cover_type === COVER_TYPES.content;
+  const isTakeActionType = cover_type === COVER_TYPES.takeAction;
+
+  let rowClassName = isContentType ? 'show-all-rows' : 'show-all-covers';
+  if (initialRowsLimit == 1) {
+    rowClassName = isContentType ? 'show-1-row' : 'show-3-covers';
```
Inwerpsel, line 9: Why aren't both things using the same `show-1-row` class :thinking:

Inwerpsel: I mean, I guess `show-3-covers` means showing a single row of 3 covers, but why would we include that detail in the classname?

assets/src/blocks/Covers/getCoversClassName.js
```diff
@@ -0,0 +1,19 @@
+import { COVER_TYPES } from './Covers';
+
+export const getCoversClassName = (cover_type, initialRowsLimit) => {
+  const isContentType = cover_type === COVER_TYPES.content;
+  const isTakeActionType = cover_type === COVER_TYPES.takeAction;
+
+  let rowClassName = isContentType ? 'show-all-rows' : 'show-all-covers';
```
Inwerpsel, line 7: Doesn't this PR also delete these classes? https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/486/files#diff-5d4325cb4bf033f21bfb7aa6b87ecb070d875d33c687d00d8cb70e71d7400968L55

assets/src/blocks/Covers/CoversEditor.js
Inwerpsel: Right we can use an empty string instead. But we don't need to set a value if we show all rows, so using 0 here is not really needed.

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -0,0 +1,52 @@
...
+  const showLoadMore = !!initialRowsLimit && covers.length > amountPerRow * row;
+
+  return (
+    <div className='container'>
+      <div className='thumbnail-largeview-container'>
+        {covers.map((cover, index) => {
+          const { href, image, alt_text, name, src_set } = cover;
+          const hideCover = !!initialRowsLimit && index >= row * amountPerRow;
+          return (
+            <div key={name} className={classnames('campaign-card-column', { hidden : hideCover })}>
```
Inwerpsel, line 16: It's actually better if we don't render what is not shown. The images for example will probably start loading for all rows.

classes/rest/class-rest-api.php
Inwerpsel: Nice. We should probably also use `useSelect` in the editor like we did there. Then we can remove this endpoint and can probably simplify the block's PHP functions too.
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/2e695d0ed1e4db9755c554666c6044c97ab56bb1/assets/src/blocks/Gallery/GalleryEditor.js#L160-L172

Inwerpsel: It would also be faster in most cases, since `useSelect` will cache things, while our endpoint will keep getting called.

assets/src/blocks/Covers/ContentCovers.js
```diff
@@ -0,0 +1,89 @@
...
+
+const isMediumWindow = () => window.innerWidth >= 576 && window.innerWidth < 992;
+
+export const ContentCovers = ({ covers, initialRowsLimit, row, loadMoreCovers }) => {
+  const [amountPerRow, setAmountPerRow] = useState(isMediumWindow() ? 3 : 4);
+
+  const updateRowAmount = () => setAmountPerRow(isMediumWindow() ? 3 : 4);
+
+  // The amount of covers per row depends on the window width
+  useEffect(() => {
```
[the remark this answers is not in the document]
Inwerpsel: This was more a braindump than an actual remark on this PR, don't think it requires changes :smile:

assets/src/blocks/Covers/getCoversClassName.js
mleray: Removed!

assets/src/blocks/Covers/CampaignCovers.js
mleray: Updated! 🙂

classes/rest/class-rest-api.php
mleray: This will be taken care of in a [follow-up ticket](https://jira.greenpeace.org/browse/PLANET-5975)!

assets/src/blocks/Covers/CoversEditor.js
Inwerpsel: This one is not a blocker if it turns out complicated to solve, just seemed defensive to not use 0 so that we can't inadvertently hide all rows. But maybe easier to consider separately afterwards :slightly_smiling_face:

Inwerpsel (commented):
It's been along way but we're almost there :+1: Left some comments for things we can address later. Just want to compare the impact on performance with the previous version before I approve, could be some small things come out of that.

comzeradd:
I added a `HOLD` label, so we can conclude/approve the review without merging till we have [PLANET-5982](https://jira.greenpeace.org/browse/PLANET-5982)

Inwerpsel (approved):
I'll approve this already since the changes required for cache invalidation will be done in a [separate PR](https://jira.greenpeace.org/browse/PLANET-5982). This one should not be merged before that is done, else we'll be serving outdated covers blocks for some time. Might require some small changes here, so we best do a final review pass after making those.

Inwerpsel:
We need to merge https://github.com/greenpeace/planet4-master-theme/pull/1366/files before acceptance tests can pass. That PR also contains the fixes to `tag.php` which are needed to ensure it doesn't start using the beta block.

PR #487, opened 2021-01-27
PLANET-5873 Fix Image block caption alignment update for WP 5.6

mleray:
## Description

See [PLANET-5873](https://jira.greenpeace.org/browse/PLANET-5873)

This PR solves the latest task listed in the ticket, a.k.a:
- Fix caption alignment option in the sidebar (from P4 options) which now doesn't update properly, probably due to this line taking the previous value of `captionAlignment`

### DIFF
--- a/assets/src/ImageBlockExtension.js
+++ b/assets/src/ImageBlockExtension.js
@@ -75,16 +75,13 @@ const addExtraControls = function() {
         );
       }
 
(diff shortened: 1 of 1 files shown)


sagarsdeshmukh (approved):
Tested on local dev env. LGTM 👍

PR #488, opened 2021-02-01
[No issue yet]: PoC - use `useGesture` for swipe detection

pablocubico:
Ref: ---

No issue yet, but we talked about this to avoid mixing DOM based libraries like Hammer with React. This is just a test of  useGesture with the Gallery Carousel, as a PoC.

---


pablocubico:
Closing this as disregarded.

PR #490, opened 2021-02-02
PLANET-5666: Add image sizes

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5666

Misnamed the branch, has nothing to do with lazy loading :facepalm: 

Add image sizes to some of our blocks. There's more to do but let's evaluate the impact of these already.

The attributes were generated using this bookmarklet https://github.com/ausi/respimagelint

Also adds missing `srcset` to Articles, that could help a lot too :grin:


mleray (approved):
Looks good to me 👍

PR #494, opened 2021-02-09
PLANET-5945: Add credits to gallery caption

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5945

> Image credits are not visible in the Gallery block.

The current code displays credits only if there's no caption.

## Fix

- Compile image title with credits if it exists

## Test

- Insert a Gallery block in a page
- Add images that have the `Credit` field filled, or fill it
- Credits should appear on the frontend, after the caption 

Screenshot from 2021-02-09 10-14-11

### DIFF
--- a/assets/src/blocks/Gallery/GalleryFrontend.js
+++ b/assets/src/blocks/Gallery/GalleryFrontend.js
@@ -5,12 +5,19 @@ import { getGalleryLayout, GALLERY_BLOCK_CLASSES } from './getGalleryLayout';
 import { Lightbox } from '../../components/Lightbox/Lightbox';
 import { useLightbox } from '../../components/Lightbox/useLightbox';
 
+const getTitle = (image) => {
+  const caption = image.caption || '';
+  const credits = image.credits && !caption.includes(image.credits)
+    ? (image.credits.includes('©') ? image.credits : `© ${image.credits}`) : '';
+  return `${caption}  ${credits}`.trim();
+}
+
 const imagesToItems = images => images.map(
   image => ({
     src: image.image_src,
     w: 0,
     h: 0,
-    title: image.caption || image.credits || ''
+    title: getTitle(image)
   })
 );
 


assets/src/blocks/Gallery/GalleryFrontend.js
```diff
@@ -5,12 +5,19 @@ import { getGalleryLayout, GALLERY_BLOCK_CLASSES } from './getGalleryLayout';
 import { Lightbox } from '../../components/Lightbox/Lightbox';
 import { useLightbox } from '../../components/Lightbox/useLightbox';
 
+const getTitle = (image) => {
+  const caption = image.caption || '';
+  const credits = image.credits && image.credits.length > 0 && ! caption.includes(image.credits)
```
mleray, line 10: I'm pretty sure that you can remove the `image.credits.length > 0` part, since an empty string will always return `false` 🤔 
https://stackoverflow.com/questions/8692982/in-javascript-is-an-empty-string-always-false-as-a-boolean

assets/src/blocks/Gallery/GalleryFrontend.js
```diff
@@ -5,12 +5,19 @@ import { getGalleryLayout, GALLERY_BLOCK_CLASSES } from './getGalleryLayout';
 import { Lightbox } from '../../components/Lightbox/Lightbox';
 import { useLightbox } from '../../components/Lightbox/useLightbox';
 
+const getTitle = (image) => {
+  const caption = image.caption || '';
+  const credits = image.credits && ! caption.includes(image.credits)
```
mleray, line 10: I find it a bit strange to have a space between `!` and the rest, but I think that's a PHP habit? 😁 Probably doesn't matter

lithrel: Ah right, not even a habit but a rule in our codebase at the moment (and one day, I'll write this ADR)

mleray (approved):
Looks good to me 👍

PR #495, opened 2021-02-09
PLANET-5978: Theme creator experiment

Inwerpsel:
REVIEW INSTRUCTIONS:
* The most important thing to look out for in this PR is that it should not affect the site unless the feature is turned on.
* So the code in `/theme` doesn't need to be extensively reviewed, we just need to be sure it has no effect whatsoever if the feature is off.

This is currently deployed on my dev instance: https://www-dev.greenpeace.org/pieterdev 
Alt + click on an element to open the editor for it (works on anything, as long as there is something that can be configured)

A while back I started trying to make a theme editor based entirely on information extracted from the selectors CSS variables are in (and a few assumptions about the naming of the variables). Then matching the clicked element against those to get all available settings for it.

I worked on it from time to time, now I feel it's at a point where it could be useful for designers on the project, since it allows for instant prototyping of any property. In fact it's not even a prototype, it's the real thing. Since we're currently in the process of setting up a design system, the ability to instantly prototype things can probably help a lot there.

~~I also included the styleguide with a bunch of added variables.~~ Removed again for merging. Unfortunately that also removes most of the themable things :sad_panda:

So after this change, what work is needed to set up a UI control for a given CSS value?
1. Add a variable around the value
2. If npm watch is on, you just need to refresh so that it can fetch the new config file that was built

~~TODO blocker for merging:~~
* ~~Integrate in a way that it can be used by logged in users if beta feature is turned on without changing anything for non logged in.~~ Done: master theme will have a toggle to enable the editor, and it's loaded with separate files.

TODO:
* Consistent handling of variables for the same property but different screen sizes
* Cleaner webpack implementation, or maybe even parse the CSS files in the browser?
* Custom controls for more properties
* Combining :active,:focus and others into the same control as the regular version of the a property
* Apply consistent variable naming.
* Consistent way to handle complex properties (e.g. either add a control for `margin` or always split up into single value properties `margin-left`,`margin-bottom`,...)


package.json
```diff
@@ -41,6 +41,10 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "jest-environment-node": "^26.0.1",
-    "photoswipe": "^4.1.3"
+    "photoswipe": "^4.1.3",
+    "font-picker-react": "^3.5.2",
```
Inwerpsel, line 45: These added deps are only used in the theme editor script, which is now loaded separately from `frontendIndex.js` only if the feature is on.

assets/src/styles/components/VarPicker.scss
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.theme-editor-highlight {
+  border: 2px solid yellow !important;
+  box-sizing: border-box !important;
+}
+
+.var-picker .var-group {
+  //max-height: 400px;
+  background: white !important;
```
comzeradd, line 50: We can remove the commented code, if not needed anymore.

assets/src/styles/components/VarPicker.scss
```diff
@@ -0,0 +1,119 @@
...
+    margin-bottom: 2px;
+  }
+}
+
+.var-picker ul > li > ul {
+  padding-left: 8px;
+}
+
+.var-picker .var-control {
+  //max-height: 400px;
```
comzeradd, line 66: Same here.

assets/src/styles/components/VarPicker.scss
```diff
@@ -0,0 +1,119 @@
...
+  color: black !important;
+  line-height: initial !important;
+}
+
+.var-picker > ul {
+  padding: 0;
+  max-height: 80vh;
+  width: 300px;
+  overflow-y: scroll;
+  resize: 'horizontal';
```
comzeradd, line 15: [Double quotes](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/.stylelintrc#L21) :grin:

classes/class-loader.php
```diff
@@ -422,6 +422,24 @@ public function enqueue_public_assets() {
 
 		// Sets translated strings for a JS script.
 		wp_set_script_translations( 'planet4-blocks-frontend', 'planet4-blocks', P4GBKS_PLUGIN_DIR . '/languages' );
+
+		if (
+			Features::is_active( Features::THEME_EDITOR_NON_LOGGED_IN )
+			|| ( is_user_logged_in() && Features::is_active( Features::THEME_EDITOR ) )
```
comzeradd, line 428: I think it's safer to enable the theme editor when both conditions are true, to catch the case where someone accidentally enables the feature on production. Or even to allow that, but just for logged in users.

comzeradd (changes requested):
Overall looks good. I left some minor comments about linting fixes. There are a few commented lines of code on the JS side, that we can remove. Also we should squash the commits (either down to one or to a few atomic commits).

On the general architecture, why is the theme editor on this repo and not on master theme? It's a "theme" editor, so semantically it would make more sense to be in the theme. I'm also thinking NROs that may want to use it, but they don't use the blocks plugin.

@pablocubico thoughts?

classes/class-loader.php
Inwerpsel: Yeah it's indeed risky. But it's also unfortunate that it has the WP admin bar if you're logged in, which spoils the design a bit. We could also add a check so that it never adds these things on production sites, and maybe not show that checkbox there?

assets/src/styles/components/VarPicker.scss
```diff
@@ -0,0 +1,119 @@
+.var-picker {
+}
```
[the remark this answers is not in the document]
Inwerpsel: Ah right, I simply forgot that I needed to clean up this file :+1:

assets/src/styles/components/VarPicker.scss
Inwerpsel: > On the general architecture, why is the theme editor on this repo and not on master theme?

Not really a deliberate choice. I think when I started it (over half a year ago) it seemed doing it in master theme would involve a lot of extra set up that was already in place in the blocks plugin. But indeed it makes more sense in master-theme.

It would also make sense as a standalone lib/plugin, since it's not really coupled to WordPress :thinking:

I'll check what moving it in master-theme would involve.

Inwerpsel:
> On the general architecture, why is the theme editor on this repo and not on master theme?

Not really a deliberate choice. I think when I started it (over half a year ago) it seemed doing it in master theme would involve a lot of extra set up that was already in place in the blocks plugin. But indeed it makes more sense in master-theme.

It would also make sense as a standalone lib/plugin, since it's not really coupled to WordPress :thinking:

I'll check what moving it in master-theme would involve.

comzeradd:
> It would also make sense as a standalone lib/plugin, since it's not really coupled to WordPress

If it's not too p4-specific that would be a great option. We could even submit it to the WP plugins registry for other orgs to use it. In that case, the Feature toggle should probably live in a different menu 🤔

Inwerpsel:
> > It would also make sense as a standalone lib/plugin, since it's not really coupled to WordPress
> 
> If it's not too p4-specific that would be a great option. We could even submit it to the WP plugins registry for other orgs to use it. In that case, the Feature toggle should probably live in a different menu

Think I remember the actual reason why it's here. The first attempt I wanted to integrate it with the block editor, and only realized after a while it would work a lot better on the front end.

Are you ok with moving it later so we have more time to think about it? For now it will not be of that much use yet outside of the team. In fact we should probably actively tell NRO devs to wait using the variables, since we need to adapt some naming here and there.

> Also we should squash the commits (either down to one or to a few atomic commits)

Down from 50+ to 6, I guess these now reflect on a high level the stages the code went through.

classes/class-loader.php
Inwerpsel: You meant checking it like this, right? https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/b18c582f52baa1049bb132516bcbe4bc4b6a4dcb/classes/class-loader.php#L426-L429

Inwerpsel:
I replaced the WP color picker with one from https://github.com/casesandberg/react-color. For now it uses the sketch picker, but there's other pickers to pick from. This fixes the issue of the color picker not updating when something else changes its value, like when undoing.

I also bumped the dashdash version, since it now supports pseudo selectors on the inside, which makes these much more concise.

classes/class-loader.php
comzeradd: Yes, but I’m still skeptical about allowing this on productions for non-loggedin users. We can check for environment, but if that’s too complex with our current setup we can wait till we [upgrade](https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/) and just allow it only for logged in users for now.

Inwerpsel: Right, would be too risky, I'll check if we can do an environment check :+1:

comzeradd (approved):
🍺

PR #496, opened 2021-02-10
PLANET-5812 Fix spreadsheet block colors issue

sagarsdeshmukh:
Ref: [JIRA 5812](https://jira.greenpeace.org/browse/PLANET-5812)

---

- Update cssvar names in spreadsheet block
- It works fine for newly added blocks. For existing spreadsheet blocks, I was thinking to add a migration script but as per the [P4 block usage report](https://docs.google.com/spreadsheets/d/1uAmZLIWYsxrBByqbhoF_vVtSM7WGebYWIc0xftPRPwE/edit?pli=1#gid=359432436), the spreadsheet block is used by only 4 NRO's (7 times), So in my opinion it would be much simple to just login and update cssvar names from block editor instead of adding a migration script.

### DIFF
--- a/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
+++ b/assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
@@ -20,21 +20,21 @@ const colors = [
 const colors_variables_map = {
   // Grey variables (default)
   '#ececec': {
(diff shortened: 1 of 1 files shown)


mleray (approved):

PR #504, opened 2021-02-16
PLANET-5960: Timeline block breaks the editing of media blocks

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-5960

TimelineJS seems to include lodash in the global `_` variable.
It replaces the Underscore `_` used by Wordpress for media interactions in the editor.
Any block that calls a Wordpress media function (here in `wp-includes/js/media-views.js`) might try to use the `_` object, and if the function has a different API in Underscore and Lodash, it will fail, and put the block (or even the editor in some situations) in a broken state (block/editor has encoutered an error).
```
Uncaught TypeError: this.activateMode is not a function
    _createModes media-views.js:2815
```

## Fix 

Running a [`_.noConflict()`](https://docs-lodash.com/v4/no-conflict/) function (also available in [Underscore](https://underscorejs.org/#noConflict)) after including TimelineJS reverts the `_` variable to its previous value.  

Although this works, I don't know if it's the right place, maybe there's a better time to run this function than the useEffect here.

## Test

With a Timeline and Carousel block:
- Add a page with 2 blocks
  - A Timeline (with url https://docs.google.com/spreadsheets/d/1wIVkeDhC5E0b_NsVVPyVEHKS2aozSaWdkdn5zl-dzFg/edit#gid=0 )
  - A carousel
- On `master` branch it will just crash the Carousel block when you add it.
- On this branch, editing the Carousel is possible
  

With only a Timeline block:

- Add a page with a Timeline (with url https://docs.google.com/spreadsheets/d/1wIVkeDhC5E0b_NsVVPyVEHKS2aozSaWdkdn5zl-dzFg/edit#gid=0 )
  - Try to edit the Featured Image
- On `master` branch the editor will crash
- On this branch, the Featured Image section opens properly

You can also try to edit blocks in an imported campaign (see attachments in https://jira.greenpeace.org/browse/PLANET-5960 )

### DIFF
--- a/assets/src/blocks/Timeline/Timeline.js
+++ b/assets/src/blocks/Timeline/Timeline.js
@@ -49,5 +49,17 @@ export const Timeline = (props) => {
 		],
 	);
 
+	// Revert TimelineJS global usage of lodash,
+	// as it conflicts with Wordpress underscore lib
+	// see https://jira.greenpeace.org/browse/PLANET-5960
+	useEffect(
+		() => {
+			if (scriptLoaded) {
+				_.noConflict();
+			}
+		},
+		[scriptLoaded],
+	);
+
 	return <div ref={ timelineNode }></div>
 }
--- a/assets/src/blocks/Timeline/Timeline.js
+++ b/assets/src/blocks/Timeline/Timeline.js
@@ -29,8 +29,16 @@ export const Timeline = (props) => {
 		});
 	}
 
+	// Revert TimelineJS global usage of lodash,
+	// as it conflicts with Wordpress underscore lib
+	// see https://jira.greenpeace.org/browse/PLANET-5960
+	const revertLodash = function () {
+		_.noConflict();
+	}
+
 	const [scriptLoaded, scriptError] = useScript(
-    `https://cdn.knightlab.com/libs/timeline3/${TIMELINE_JS_VERSION}/js/timeline-min.js`
+    `https://cdn.knightlab.com/libs/timeline3/${TIMELINE_JS_VERSION}/js/timeline-min.js`,
+    revertLodash
 	);
 
 	useEffect(
@@ -47,18 +55,6 @@ export const Timeline = (props) => {
 			timenav_position,
 			language,
 		],
-	);
-
-	// Revert TimelineJS global usage of lodash,
-	// as it conflicts with Wordpress underscore lib
-	// see https://jira.greenpeace.org/browse/PLANET-5960
-	useEffect(
-		() => {
-			if (scriptLoaded) {
-				_.noConflict();
-			}
-		},
-		[scriptLoaded],
 	);
 
 	return <div ref={ timelineNode }></div>
--- a/assets/src/components/useScript/useScript.js
+++ b/assets/src/components/useScript/useScript.js
@@ -2,7 +2,7 @@
 import { useEffect, useState } from 'react';
 import { addScriptTag } from './addScriptTag';
 
-export const useScript = (src) => {
+export const useScript = (src, onScriptLoaded) => {
   // Keeping track of script loaded and error state
   const [state, setState] = useState({
     loaded: false,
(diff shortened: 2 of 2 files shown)


assets/src/blocks/Timeline/Timeline.js
```diff
@@ -35,8 +35,12 @@ export const Timeline = (props) => {
 
 	useEffect(
 		() => {
-			if (stylesLoaded && scriptLoaded) {
-				setupTimeline();
+			if (scriptLoaded) {
+				_.noConflict();
```
Inwerpsel, line 39: We're only using lodash to generate a unique ID, maybe we can generate the ID another way and remove the dependency?

lithrel: From what I've tested it's not our usage (tried removing uniqueId and debounce), it's the timeline lib itself that is the problem https://cdn.knightlab.com/libs/timeline3/3.8.10/js/timeline.js

lithrel: Also tried compiling their lib with different options to restrict lodash import to some functions, but I'm not very competent :grimacing:

Inwerpsel: Right, missed that by not reading your description well enough :facepalm: Is there a way we can prevent the timeline JS from overriding `_`? If we run `_.noConflict` after it there will still be a window in which `wp` could crash if it's trying to use underscore. Though we could have a look at how WP uses underscore, maybe there's just a few usages and they're probably not going to use it more extensively I guess.

Inwerpsel: The window would be quite small, so I guess this fix is good. Maybe it's better to create a separate effect with just the `scriptLoaded` dependency that only does the noConflict?

lithrel: > Is there a way we can prevent the timeline JS from overriding `_`?

Couldn't find a local solution; so  I looked at compiling their lib with https://github.com/lodash/babel-plugin-lodash, it's supposed to pick only the used functions of lodash without the complete `_` object - but I think I'm failing to use it properly (my compiled file is exactly the same), so if anybody with some experience at this could try it that'd be helpful :pray:  
I don't see any other solution at the moment, but I'll take any idea :bulb:

lithrel:
@Inwerpsel  I've updated the PR with a separate effect doing only `_.noConflict` (+ some comment)

assets/src/blocks/Timeline/Timeline.js
```diff
@@ -16,10 +16,10 @@ export const Timeline = (props) => {
 	const timelineNode = useRef(null);
 
 	const [stylesLoaded, stylesError] = useStyleSheet(
-    `https://cdn.knightlab.com/libs/timeline3/${TIMELINE_JS_VERSION}/css/timeline.css`
+		`https://cdn.knightlab.com/libs/timeline3/${TIMELINE_JS_VERSION}/css/timeline.css`
```
Inwerpsel, line 19: Just checking if this spacing change is indeed intended?

Inwerpsel (approved):

assets/src/blocks/Timeline/Timeline.js
lithrel: Excellent question :grin: I realized that file was indented with tabs and my code had spaces in it so I converted the whole file to tabs for coherence. I know that the local `.editorconfig` file defines the indentation as 2 spaces, I chose the lesser diff for now; I can also do the opposite if you're ok with it.

Inwerpsel: Since we need to revisit our style config anyway, I guess it's ok to leave the file inconsistent if that means we don't need to modify more lines than are actually changing. Spacing changes can be an obstacle hiding more relevant commit messages.

Inwerpsel (approved):

lithrel:
@Inwerpsel I did a last test before merging and realized this function was called for each insertion of a timeline, thus reverting the noconflict and loading the wrong lib if you add it twice (or add/remove/add) :facepalm: 
I moved the noConflict call to useScript::onScriptLoad because it is ran only once per script there. Another solution would be to detect if `_` is Underscore or Lodash, but it looks a bit tricky and not stable (it would be based on some functions presence which can change with versions).

Inwerpsel (approved):

PR #508, opened 2021-02-18
PLANET-4535: Perf: Split blocks JS into chunks where possible

pablocubico:
This is a preliminary step for the [Carousel Header PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/483).

Ref: https://jira.greenpeace.org/browse/PLANET-4535

Notes:
* We had many inconsistencies around the BLOCK_NAME of each block. In some cases it was defined but not used. Some names had underscores (not allowed), some names were prefixed with `planet4-blocks` and others weren't.
* To fix those consistencies, I splitted the namespace into a different Base_Block property that can be overriden.
* Also made sure we are not hardcoding block names anywhere.
* Wordpress uses 4 standard names to load the assets for each block:
  - script
  - style
  - editor_script
  - editor_style
  I changed the key for the **bundled** scripts (editorIndex, etc) accordingly. So `planet4-blocks-frontend` is now `planet4-blocks-script` and so on.
* Added a helper to generate the entry points for a given block in `webpack.config.js`
* Added a helper to Base_Block to register the 4 assets for any given block.
* Did all the necessary changes for the Accordion block for two reasons: it's the first alphabetically, and it uses a custom namespace (planet4-blocks-beta), so it was a good PoC.

This implementation is compatible with the previous one, so we don't need to split all the blocks in one go. We can go block by block, I can describe the steps to do so in a new ticket and we can add one for each block.

---


pablocubico:
heads up @mleray @Inwerpsel :

Using the `script` and `style` attributes of the `register_block_type` function causes WP to enqueue all the blocks whether they are used on a page or not.

So we'll need to implement the same check GPNL implemented here: 

https://github.com/greenpeace/planet4-gpnl-plugin-gutenberg-blocks/blob/08328f26500bb8abc3faf638b687035e6c00be60/classes/blocks/class-heroimage.php#L64-L78

assets/src/blocks/Accordion/AccordionEditorScript.js
```diff
@@ -0,0 +1,59 @@
+import { AccordionEditor } from './AccordionEditor';
```
Inwerpsel, line 1: Not sure why the other ones are seen as a rename, while this one as a removal and addition :thinking:

classes/blocks/class-columns.php
```diff
@@ -79,7 +79,7 @@ public function __construct() {
 		add_shortcode( 'shortcake_columns', [ $this, 'add_block_shortcode' ] );
 
 		register_block_type(
-			'planet4-blocks/columns',
+			self::BLOCK_NAMESPACE_PREFIX . '/' . self::BLOCK_NAME,
```
Inwerpsel, line 82: We can put this into the base block class in a function since it's the same for all blocks.

classes/blocks/class-cookies.php
```diff
@@ -24,13 +24,12 @@ public function __construct() {
...
 		register_block_type(
-			'planet4-blocks/cookies',
+			self::BLOCK_NAMESPACE_PREFIX . '/' . self::BLOCK_NAME,
 			[
-				'editor_script'   => 'planet4-blocks/cookies',
 				// todo: Remove when all content is migrated.
 				'render_callback' => static function ( $attributes ) {
 					$json = wp_json_encode( [ 'attributes' => $attributes ] );
-					return '<div data-render="planet4-blocks/cookies" data-attributes="' . htmlspecialchars( $json ) . '"></div>';
+					return '<div data-render="' . self::BLOCK_NAMESPACE_PREFIX . '/' . self::BLOCK_NAME . '" data-attributes="' . htmlspecialchars( $json ) . '"></div>';
```
Inwerpsel, line 32: This is also the same for all front end rendered blocks. So instead of replacing the name with the constants everywhere, we could extract it to the base block class as well.

classes/blocks/class-columns.php
Inwerpsel: In fact we can probably move the entire `register_block_type` call into the base class.

Inwerpsel: Though that would increase this PR's size a lot, just replacing all concatenations with a function is maybe better for now.

Inwerpsel:
@pablocubico Maybe it's easier if we set up an implementation for just one block until we have something stable, and only then apply that to all blocks. Then we don't need to change all the files each time we change strategy.

classes/blocks/class-columns.php
pablocubico: Sounds good.

assets/src/blocks/Accordion/AccordionEditorScript.js
pablocubico: Maybe I forgot to use `git mv` for this one... 🤔

Inwerpsel: `git mv` shouldn't make a difference, since Git detects whether it's a name "after the fact". `git mv` is just a shortcut for deleting the old file and staging the new one.

Inwerpsel: Maybe spacing in the file changed? When deleting a file, Git will look for exact matches of lines and past a certain threshold it will be regarded as a rename.

pablocubico:
> @pablocubico Maybe it's easier if we set up an implementation for just one block until we have something stable, and only then apply that to all blocks. Then we don't need to change all the files each time we change strategy.

@Inwerpsel Well, actually I tried to keep this to a minimum, adding the entry points only for the Accordion block (which is the first one alphabetically and has a custom namespace). That and the Base Block are the only things with serious changes, the rest is just replacing one or two hardcoded strings per block and leaves things in a more consistent and predictable state, I think we should move forward with that.

assets/src/blocks/Accordion/AccordionEditorScript.js
pablocubico: Oh wait, it's because the `registerAccordionBlock` function (that was wrapping almost everything) is gone, as it's not needed in this approach, it runs `registerBlockType` directly.

classes/blocks/class-base-block.php
```diff
@@ -94,4 +96,74 @@ protected function is_rest_request() {
...
+		wp_enqueue_style(
+			static::BLOCK_NAMESPACE_PREFIX . '/' . static::BLOCK_NAME . '-style',
+			$file_url,
+			[],
+			\P4GBKS\Loader::file_ver($file_path),
+		);
+	}
+
+	public static function enqueue_frontend_assets() {
+		if ( has_block( static::BLOCK_NAMESPACE_PREFIX . '/' . static::BLOCK_NAME ) ) {
```
Inwerpsel, line 142: We should keep an eye on the performance of this, since it does parse the post content again for each type of block we have, so with more types of blocks and lengthy post content this might not scale well. Maybe we can set it up in a way so that we parse blocks once and then search the result for each block type?

Could be `parse_blocks` performance is good enough and this is a total non issue.

Inwerpsel: Ah nvm it internally just does a regex, should be fine :+1:

assets/src/blocks/Accordion/AccordionEditorScript.js
comzeradd: It's because you also made a change to that file. Ideally you should create an initial commit with just all the renames/moves and then add the changes to a separate commit.

pablocubico: Right, I did that in the Carousel PR.

classes/blocks/class-cookies.php
pablocubico: I added the function, I didn't extract the render callback to `Base_Block` in case we want to do some custom logic here, as happens with the attributes for some blocks.

Inwerpsel: I was more thinking of putting just what is now returned into a function, not the entire `render_callback`. But either way works for me :ok_hand:

Inwerpsel (commented):
We can simplify a few things, but overall looks great already :ok_hand:

planet4-gutenberg-blocks.php
```diff
@@ -107,7 +107,7 @@
 */
 
 const POST_BLOCK_TYPES = [
-	'planet4-blocks/accordion',
+	'planet4-blocks-beta/accordion',
```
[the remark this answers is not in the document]
pablocubico: Currently, it has `beta` in the name, so changing that would mean we'll need to migrate content after this merge. Maybe we should split this into a "promote Accordion" ticket. 🤔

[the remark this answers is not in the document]
Inwerpsel: Do you mean "currently" as in what is now live? Looks like this PR makes that name change, maybe I'm missing something? https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/508/files/9d9640fef2a918b0c3fc8d7ac35adec430ff9d5e#diff-b770fab1ec8843c94dcf94f02b81fcf335cc4e234df177bf9658002b121ccc66L6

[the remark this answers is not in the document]
Inwerpsel: Because it's a deleted file you need to open it (the first one). After opening it will still highlight the line I linked to.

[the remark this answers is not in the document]
pablocubico: Yes, I meant live in production but you're right, I was looking at the PHP class: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-accordion.php#L30

The block is indeed registered without `-beta` in the editor JS so the HTML comments in the content should have `planet4-blocks/accordion`. :+1:

I'll make it use the default namespace then, at least this was useful to add support for a different namespace!

classes/blocks/class-base-block.php
```diff
@@ -94,4 +96,93 @@ protected function is_rest_request() {
...
+	 */
+	public static function enqueue_editor_assets() {
+		$camelized_block_name = static::get_camelized_block_name();
+		$url_path             = trailingslashit( P4GBKS_PLUGIN_URL ) . 'assets/build/' . $camelized_block_name;
+		$dir_path             = trailingslashit( P4GBKS_PLUGIN_DIR ) . 'assets/build/' . $camelized_block_name;
+		$block_name           = static::get_full_block_name();
+
+		// Editor Script.
+		wp_enqueue_script(
+			$block_name . '-editor-script',
```
Inwerpsel, line 111: Do we need to enqueue this ourselves for the editor? I think it can also be passed with `register_block_type`.

Inwerpsel: It's documented here, but only in the comments :roll_eyes: https://developer.wordpress.org/reference/functions/register_block_type/#user-contributed-notes

classes/blocks/class-base-block.php
```diff
@@ -94,4 +96,93 @@ protected function is_rest_request() {
...
+			$url_path . 'EditorScript.js',
```
Inwerpsel, line 112: It seems simpler to give each block its own directory with all files. Then we can use generic names (`edit.js` like WP does) for the files and don't need to repeat the block name in all of them.

classes/blocks/class-base-block.php
pablocubico: Yes, I enqueued them in `register_block_type` initially, along with the frontend scripts, but as it always enqueues the frontend scripts no matter if the block is present or not, I moved the frontend scripts apart, and then I moved the editor scripts apart too because I found it weird to have some enqueued by `register_block_type` and some outside it.

Inwerpsel: It may be a bit weird but it's a more high level way of writing it since it avoids the enqueue logic, we just pass a name. Also if WP decides to lazy load the scripts in the editor (not sure if they do or plan to), then using their API will allow us to benefit from that.

pablocubico: Hmmm not sure about this, we will still need to keep the frontend styles apart from `register_block_type`. So it would mean we'll have to keep `enqueue_editor_assets` anyway... OR, maybe we can include the frontend styles in the editor styles build. What do you think?

classes/class-loader.php
```diff
@@ -330,7 +330,7 @@ public function enqueue_editor_scripts( $hook ) {
 
 		// Enqueue editor script for all Blocks in this Plugin.
 		self::enqueue_local_script(
-			'planet4-blocks-script',
+			'planet4-blocks-editor-script',
```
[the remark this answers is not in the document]
pablocubico: I think it's sane to change the handles if we are going to make it closer to WP's conventions, as the change to `edit.js` and so on.

classes/blocks/class-base-block.php
pablocubico: I just checked their convention and it goes as follows:

`index.js` -> the place where `registerBlockType` happens.
`edit.js` -> the `edit()` function
`save.js` -> the `save()` function

Of course there is no frontend script.
Is that your proposal?

classes/blocks/class-enform.php
```diff
@@ -37,13 +37,13 @@ class ENForm extends Base_Block {
 	 * ENForm constructor.
 	 */
 	public function __construct() {
-		if ( WP_Block_Type_Registry::get_instance()->is_registered( 'planet4-blocks/enform' ) ) {
+		if ( WP_Block_Type_Registry::get_instance()->is_registered( self::get_full_block_name() ) ) {
```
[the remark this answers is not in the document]
sagarsdeshmukh: yes indeed. we don’t need this check anymore 👍

classes/class-loader.php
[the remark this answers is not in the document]
Inwerpsel: It's definitely a good idea, not arguing with that. My suggestion is to separately rename in a follow up PR, with the goal of reducing the amount of risk for this PR, since we're already changing a lot.

classes/blocks/class-base-block.php
Inwerpsel: Makes sense :+1:

pablocubico: Ok, I'll go ahead with the change so we can move forward, personally, I'm not sure if I like WP's conventions, I'm not a big fan of having a bunch of files with the same name but in different folders (`index.js` was a long standing discussion in the NodeJS community), aaaand... we are doing things a bit differently than WP, for example, there is no frontend script except for the `save()` method used to generate the static (pre-hydration) markup. We'll need a different script to make the hydration work. We'll see about that in the Carousel Header PR.

pablocubico: After going back to this, there is an issue: the `edit.js`, `index.js` and `save.js` convention is for the source files, not the build files. Registering blocks actually changed a lot since we implemented our blocks, I'm not sure about this change, feels like an unnecessary step at this point, and the current version is consistent with the entry points we need to build to make things run in the browser, I actually kinda like it a bit more than the WP way. 

In fact, our implementation is a bit outdated if you compare it to the current WP implementation:
https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src

So maybe we should consider updating our implementation using JSON metadata, CSS modules, etc., but I think it's something for a subsequent stage. 

I don't know what problem are we solving here, I mean, we can exchange our opinionated views about file and function names endlessly but I don't think is truly relevant.

classes/class-loader.php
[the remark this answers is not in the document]
pablocubico: Sure, I get the risk, but I think this handle is not used broadly, right? We can do a quick search in the blocks repo and the master theme and that's it. Maybe in some customized child-themes in case someone added it as a dependency, but other than that, this change was meant to make it consistent with the rest of the naming conventions in this PR, not changing seems like a flaw to me.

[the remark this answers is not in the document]
comzeradd: From a search on Github it looks like this repo is the only one using this handle.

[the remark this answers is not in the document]
Inwerpsel: Yeah I guess if we're just renaming the handle and nothing else is using it, it should be fine :ok_hand:

classes/blocks/class-base-block.php
Inwerpsel: If you think it's better to address this later that's fine for me. It seemed to make sense at this moment since it would make the code to handle these easier to work with (less string concatenation). But seems easy to address later :ok_hand:

classes/blocks/class-base-block.php
Inwerpsel: > maybe we can include the frontend styles in the editor styles build

That may be a simple solution too, probably all individual scripts are always loaded in the editor anyway, so performance would be about the same. And it's not very performance senstive too.

pablocubico: Circling back to this: as mentioned in another comment, since the WP implementation of the handling of these files changed quite a bit, I would keep this enqueueing logic in our side, so we have more control over it and it's consistent in both sides (editor & frontend). As you said, we don't know what plans WP has for them. I think it would be better to catch up with the WP implementation later, using the JSON metadata files and such.

Inwerpsel (approved):
Looks good :+1: We could still keep the handle rename for later, since you also mentioned we'll wait a bit with following WP's way of loading things. But I'm fine with including it too.

pablocubico:
@Inwerpsel @comzeradd So after a few days trying to fix things that weren't broken... I realized the Acceptance tests are failing because the build is created with the Master Theme's `master` branch insted of the corresponding `planet-4535`. How do I make the Github checks run the correct master theme branch?

pablocubico:
> @Inwerpsel @comzeradd So after a few days trying to fix things that weren't broken... I realized the Acceptance tests are failing because the build is created with the Master Theme's `master` branch insted of the corresponding `planet-4535`. How do I make the Github checks run the correct master theme branch?

The build pipeline is failing in the test environment: 
https://app.circleci.com/pipelines/github/greenpeace/planet4-test-telesto/14/workflows/a7ff411e-124d-4199-b378-b802ac691c79/jobs/67

Inwerpsel:
> > @Inwerpsel @comzeradd So after a few days trying to fix things that weren't broken... I realized the Acceptance tests are failing because the build is created with the Master Theme's `master` branch insted of the corresponding `planet-4535`. How do I make the Github checks run the correct master theme branch?
> 
> The build pipeline is failing in the test environment:
> https://app.circleci.com/pipelines/github/greenpeace/planet4-test-telesto/14/workflows/a7ff411e-124d-4199-b378-b802ac691c79/jobs/67

It might be using the right branch, our docker build first installs https://github.com/greenpeace/planet4-base/blob/main/composer.json, then it does a second install with the local file that has your branch name. It fails on the first install, that's why you're seeing master. The fix is on [this branch of the builder](https://github.com/greenpeace/planet4-builder/compare/fix-composer-build?expand=1), still needs to be reviewed but it seems to be working well. You can deploy the test instance if you pin the builder like this https://github.com/greenpeace/planet4-test-nix/blob/main/.circleci/config.yml#L13-L15

PR #510, opened 2021-02-19
Fix stylelint recursive path

comzeradd:
stylelint needs a double-quoted path, otherwise it doesn't make a recursive lookup.

Fixed all pending issues stylelint did't report correctly up until now.


mleray (approved):
Looks good to me 👍

PR #511, opened 2021-02-19
PLANET-5875 Update Table  block edit file for WP5.6

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5875

I've updated the `edit` file with the new version for WP5.6 and modified it to fit our needs. This PR should also solve the Table block part of [PLANET-5871](https://jira.greenpeace.org/browse/PLANET-5871) (it adds back the top controls for moving/deleting/etc the block).

You can test it on the [test-titan instance](https://www-dev.greenpeace.org/test-titan/).

There is another way to solve these issues, explored in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/515.

### DIFF
--- a/assets/src/components/p4_table/edit.js
+++ b/assets/src/components/p4_table/edit.js
@@ -35,19 +35,6 @@ import {
 	ToolbarGroup,
 	ToolbarItem,
 } from '@wordpress/components';
-import {
-	alignLeft,
-	alignRight,
-	alignCenter,
-	blockTable as icon,
-	tableColumnAfter,
-	tableColumnBefore,
-	tableColumnDelete,
-	tableRowAfter,
-	tableRowBefore,
-	tableRowDelete,
-	table,
-} from '@wordpress/icons';
 import { createBlock } from '@wordpress/blocks';
 
 /**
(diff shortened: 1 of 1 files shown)


comzeradd:
@mleray I rebased `wp56` to master. Can you rebase you branch with `wp56`?

Inwerpsel (changes requested):
We should only keep on overriding this much code as an absolute last resort, it's a maintenance burden since we will have to match all changes to core and will probably cause bugs.

What changes were made compared to the core file? Currently it's hard to see because it all happens in one commit. Maybe you can add another commit in between with the changes to the core file alone?

assets/src/components/p4_table/edit.js
```diff
@@ -67,56 +54,51 @@ import {
...
 	},
 ];
 
 const withCustomBackgroundColors = createCustomColorsHOC( BACKGROUND_COLORS );
 
 const cellAriaLabel = {
-	head: __( 'Header cell text' ),
-	body: __( 'Body cell text' ),
-	foot: __( 'Footer cell text' ),
+	head: __( 'Header cell text', 'planet4-blocks-backend' ),
```
Inwerpsel, line 94: Why are we adding our text domain here? :thinking: All these strings are already translated by core, and for blocks where we don't override we also just use the core translations.

assets/src/components/p4_table/edit.js
```diff
@@ -35,19 +35,6 @@ import {
 	ToolbarGroup,
 	ToolbarItem,
 } from '@wordpress/components';
-import {
```
Inwerpsel: Didn't this work? We're modifying it to pass strings, but as far as I can tell it's doing the same thing.

Inwerpsel (commented):
I don't think we need to modify all these text domains. Doing so makes us unable to use the core translations, and so gives us a bunch of additional translations to maintain. Probably the same goes for the icons, we're overriding it to a different way of doing the same thing.

assets/src/components/p4_table/edit.js
```diff
@@ -67,41 +54,36 @@ import {
...
+    slug: 'blue'
+	}
 ];
 
 const ALIGNMENT_CONTROLS = [
 	{
-		icon: alignLeft,
-		title: __( 'Align column left' ),
+		icon: "editor-alignleft",
+		title: __( 'Align Column Left' ),
```
sagarsdeshmukh: The original translation string is case sensitive. so I think we should keep it as it is.

sagarsdeshmukh (approved):
The PR tested on local with WPv5.6.1 The table block is working as expected. 👍

assets/src/components/p4_table/edit.js
mleray: Good catch! Updated 👍

PR #512, opened 2021-02-23
PLANET-5815 Fix split two columns text size on large screen

sagarsdeshmukh:
Ref: [JIRA 5815](https://jira.greenpeace.org/browse/PLANET-5815)

---

- I don’t have a big monitor to test the exact screen sizes, but with browser debugger tool I tested the fix

### DIFF
--- a/assets/src/styles/blocks/SplitTwoColumns.scss
+++ b/assets/src/styles/blocks/SplitTwoColumns.scss
@@ -78,12 +78,7 @@
     margin-bottom: 12px;
   }
 
-  @include large-and-up {
-    font-size: 4.125rem;
-  }
-
   @include x-large-and-up {
-    font-size: $font-size-xxl;
     margin-top: -.5em;
     margin-bottom: 20px;
   }


assets/src/styles/blocks/SplitTwoColumns.scss
```diff
@@ -79,7 +79,7 @@
   }
 
   @include large-and-up {
-    font-size: 4.125rem;
+    font-size: $font-size-xxl;
```
mleray, line 82: Maybe you can just remove this entire `large-and-up` query? Since the font-size is the same as the one set in [just before](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/37bb64fefc55b4ac2958d62e0566f7e3dc63f1f0/assets/src/styles/blocks/SplitTwoColumns.scss#L75) 🙂 
And actually you can also probably remove the one set [just after](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/37bb64fefc55b4ac2958d62e0566f7e3dc63f1f0/assets/src/styles/blocks/SplitTwoColumns.scss#L86) since it's also the same value 🤔

sagarsdeshmukh: yes, indeed 👍  Thanks 😊

mleray (approved):
Looks good to me 👍

PR #513, opened 2021-02-23
PLANET-5871 Update button block files with custom colors, remove BorderPanel & WidthPanel

sagarsdeshmukh:
Ref: [JIRA 5871](https://jira.greenpeace.org/browse/PLANET-5871)

---

- Updated the native button block files with the new version for WP5.6.
- Make the default button style selected by default
- Move P4 button styles override from [editor.js](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/edf475f0c227657b91a32a31b7b43e4b880c9d9d/admin/js/editor.js#L4) to [BlockFilter.js](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/BlockFilters.js)

### DIFF
--- a/assets/src/components/p4_button/edit.js
+++ b/assets/src/components/p4_button/edit.js
@@ -17,11 +17,8 @@ import classnames from 'classnames';
 import { __ } from '@wordpress/i18n';
 import { useCallback, useState } from '@wordpress/element';
 import {
-	Button,
-	ButtonGroup,
 	KeyboardShortcuts,
 	PanelBody,
-	RangeControl,
 	TextControl,
 	ToggleControl,
 	ToolbarButton,
@@ -37,7 +34,6 @@ import {
 	__experimentalUseEditorFeature as useEditorFeature,
 } from '@wordpress/block-editor';
 import { rawShortcut, displayShortcut } from '@wordpress/keycodes';
-import { link, linkOff } from '@wordpress/icons';
 import { createBlock } from '@wordpress/blocks';
 
 /**
@@ -47,9 +43,6 @@ import ColorEdit from './color-edit';
 import getColorAndStyleProps from './color-props';
 
 const NEW_TAB_REL = 'noreferrer noopener';
-const MIN_BORDER_RADIUS_VALUE = 0;
-const MAX_BORDER_RADIUS_VALUE = 50;
-const INITIAL_BORDER_RADIUS_POSITION = 5;
 
 const EMPTY_ARRAY = [];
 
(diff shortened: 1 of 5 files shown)


comzeradd:
@sagarsdeshmukh can you rebase with `wp56`? I see it has some commits that are already there.

assets/src/components/p4_button/edit.js
```diff
@@ -15,107 +16,71 @@ import classnames from 'classnames';
  */
 import { __ } from '@wordpress/i18n';
 import { useCallback, useState } from '@wordpress/element';
-import { compose } from '@wordpress/compose';
 import {
+	Button,
```
mleray, line 20: You've added these but they seem unused...?

assets/src/components/p4_button/edit.js
```diff
@@ -15,107 +16,71 @@ import classnames from 'classnames';
...
-	__experimentalPanelColorGradientSettings as PanelColorGradientSettings,
 	RichText,
-	withColors,
+	useBlockProps,
 	__experimentalLinkControl as LinkControl,
+	__experimentalUseEditorFeature as useEditorFeature,
 } from '@wordpress/block-editor';
 import { rawShortcut, displayShortcut } from '@wordpress/keycodes';
-import { link } from '@wordpress/icons';
+import { link, linkOff } from '@wordpress/icons';
```
mleray, line 40: These icons are also unused (in fact I don't think we even include `wordpress/icons` 🤔)

mleray (commented):
~~Maybe it would be nice to take this opportunity to add the translation domain (`planet4-blocks-backend`) to the translatable strings? 🤔~~
Nvm, @Inwerpsel just informed me that these strings have core translations so we don't need to worry about that 😄

sagarsdeshmukh:
> Maybe it would be nice to take this opportunity to add the translation domain (`planet4-blocks-backend`) to the translatable strings? 🤔

If we use our `text-domain`, then that need to be translate by NRO teams, if we keep it empty then wordpress uses a default core wordpress translation strings and this code belong to wp core, I thing we could use the default translation 🤔

mleray:
> If we use our `text-domain`, then that need to be translate by NRO teams, if we keep it empty then wordpress uses a default core wordpress translation strings and this code belong to wp core, I thing we could use the default translation 🤔

Yes yes indeed, I reverted my previous comment because I was not aware of this!

assets/src/components/p4_button/edit.js
sagarsdeshmukh: yes indeed. removing unused stuff 👍

mleray (approved):

PR #515, opened 2021-02-24
PLANET-5875 Update Table block colors without overriding the core file

mleray:
### Description
See https://jira.greenpeace.org/browse/PLANET-5875

This is another possible way to solve the issues described in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/511. It has the advantage to not override the core file, and therefore potentially prevent some other issues in the future.


sagarsdeshmukh:
I was testing the PR on local and looks like the existing Table block colors are reset with new changes 🤔

mleray:
> I was testing the PR on local and looks like the existing Table block colors are reset with new changes 🤔
> The newly added table block color settings works fine.

Right, I guess that's because we also updated the [colors' slugs](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/components/p4_table/edit.js#L51) when previously overriding the core file... The right slugs are [here](https://github.com/WordPress/gutenberg/blob/f3c538578aa818e76dbe09445d0fb1678166086d/packages/block-library/src/table/edit.js#L61). So in existing blocks the class name will be for instance `has-green-background` instead of `has-subtle-pale-green-background` 🤔 I can think of 2 solutions: 
- Add back the old `has-green-background` etc selectors to our css, in addition to the new `has-subtle-pale-green-background` ones
- Run a script to replace old class names with the new ones in existing Table blocks

sagarsdeshmukh:
> * Add back the old `has-green-background` etc selectors to our css, in addition to the new `has-subtle-pale-green-background` ones

Hi @mleray , 
The first option (`Add back the old has-green-background etc selectors to our css, in addition to the new has-subtle-pale-green-background ones`) looks simple and less complicated.

mleray:
It turns out that with this solution, existing blocks lose their previous CSS classes (e.g. `has-green-background`) and therefore it's not really possible to update them and they lose their color setting... So I'm closing this PR in favor of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/511 which makes the block behave properly.

PR #516, opened 2021-02-25
PLANET-5970 Restore previous youtube embed width

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5970
---

The CSS packaged with the player sets a max-width on the element, whereas previously our iframe had a 100% width and no max-width. I added a new rule for this element specifically.


lithrel (approved):
:+1: tested locally

PR #518, opened 2021-02-27
A11y: Convert carousel anchors to buttons

aeisenberg:
Partial fix for:
Ref: https://github.com/greenpeace/planet4/issues/125

Accessibility fix: convert anchors into buttons on the carousels.

### DIFF
--- a/assets/src/blocks/Gallery/GalleryCarousel.js
+++ b/assets/src/blocks/Gallery/GalleryCarousel.js
@@ -80,10 +80,10 @@ export const GalleryCarousel = ({ images, onImageClick }) => {
       }
       <div className="carousel-inner" role="listbox">
         {images.length > 1 &&
-          <a className="carousel-control-prev" role="button" onClick={goToPrevSlide}>
+          <button className="carousel-control-prev" role="button" onClick={goToPrevSlide}>
             <span className="carousel-control-prev-icon" aria-hidden="true"><i></i></span>
             <span className="sr-only">{__('Previous', 'planet4-blocks')}</span>
-          </a>
+          </button>
         }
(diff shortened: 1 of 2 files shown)


mleray (changes requested):
I've tested your changes for the GalleryCarousel on local and it looks good to me 👍
However your changes for `CarouselHeader_full_width_classic.scss` and `CarouselHeader.scss` should go together with updating the corresponding [twig template](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/templates/blocks/carousel_header_full-width-classic.twig#L70)... But I don't think we actually need to update that component right now since it's being [converted to WYSIWYG](https://jira.greenpeace.org/browse/PLANET-4809) 🤔 So I think you can revert these changes and just update the GalleryCarousel for now!

aeisenberg:
Thanks for the review. I addressed your comments and I haven't (yet) looked at the failing tests. Hmmmm....

Also, I have some questions on https://github.com/greenpeace/planet4/issues/125, would you be able to answer them?

mleray (approved):
Looks good to me 👍

aeisenberg:
It _looks_ like the `main` check is failing because I don't have access. 

> request-instance - Unauthorized

Is this OK to merge as-is, or should this be fixed before merging?

mleray:
> It _looks_ like the `main` check is failing because I don't have access.
> 
> > request-instance - Unauthorized
> 
> Is this OK to merge as-is, or should this be fixed before merging?

This would be to deploy your changes to one of our test instances, indeed you are probably missing some permissions for that... But I've tested it on local so it's okay to merge as is 🙂

PR #521, opened 2021-03-03
PLANET-5981 Updates for Bootstrap 5

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5981

Changes made to comply with the [migration doc](https://getbootstrap.com/docs/5.0/migration/):
- Rename all Bootstrap attributes' prefixes from `data-` to `data-bs-`
- Remove `form-inline` and `custom-select` classes (removed from Bootstrap 5)
- Add `form-select` to select components instead of `form-control` (which now has the `appearance: none;` hiding the select's arrow)
- Remove `mr-auto` class (it should have been renamed `me-auto`, but wasn't doing anything except causing issues in RTL sites so better remove it 🤷‍♀️)
- Update ContentCovers width overrides
- Add `position: relative;` to `col` class for GalleryThreeColumns since it was removed from Bootstrap
- Rename `carousel-item-left` and `carousel-item-right` to `carousel-item-start` and `carousel-item-end` respectively
- Remove max-height and max-width from video embeds and update vertical margin

### Testing

You can check the changes on [this test instance](https://www-dev.greenpeace.org/test-janus/), basically everything should look/behave the same as it does in the current sites 🙂 

### Related PRs

- [styleguide](https://github.com/greenpeace/planet4-styleguide/pull/104)
- [master-theme](https://github.com/greenpeace/planet4-master-theme/pull/1323)

### DIFF
--- a/assets/src/blocks/Articles/ArticlesFrontend.js
+++ b/assets/src/blocks/Articles/ArticlesFrontend.js
@@ -39,7 +39,7 @@ export const ArticlesFrontend = (props) => {
         { hasMorePages &&
         <div className="row">
           { read_more_link ?
-            <div className="col-md-12 col-lg-5 col-xl-5 mr-auto">
+            <div className="col-md-12 col-lg-5 col-xl-5">
               <a
                 className="btn btn-secondary btn-block article-load-more"
                 href={ read_more_link }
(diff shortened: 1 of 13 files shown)


comzeradd:
Let's fix/review that with the current code (Bootstrap5), but let's hold merge till we decide we are realising Bootstrap5.

lithrel (approved):
:+1:

PR #523, opened 2021-03-04
PLANET-5983: Add variables to blocks CSS, part I

Inwerpsel:

---
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/523/files?diff=split&w=1

There are some more to do, but we can already assess this first batch of blocks. Shouldn't be an issue to merge this in multiple chunks.

~~Master theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1325~~ merged
New master theme PR with variables I needed to add because of blocks CSS: https://github.com/greenpeace/planet4-master-theme/pull/1360

Initially my plan was to replicate the existing themes in code. But while doing that I found it's quite clear cut which variables need to be added in most cases, while trying to make a theme work property by property without having the variables yet is a slow process. With the simple way of saving themes on the server, it's a lot easier to create these on the fly. So I'll do that and add them to Git in the end.

Some things that we may want to reconsider before adding variables:
- Articles: Uses the same CSS for the tags as the one used for the tags at the top of the page in master theme. I'm adding them now anyway since the top tags need them. But maybe it's better to separate these. Mainly the font size and spacing requirements for both places could be different.

TO TEST/REVIEW:
All these changes should leave existing content unchanged. I had to make minor changes to the CSS structure, but each of them seemed safe to do. Left a comment on the PR where I did so.

### DIFF
--- a/assets/src/styles/blocks/Columns.scss
+++ b/assets/src/styles/blocks/Columns.scss
@@ -4,10 +4,6 @@
       font-weight: inherit;
     }
 
-    &.btn-primary {
-      color: $white;
-    }
-
     &.btn-secondary {
       width: 100%;
     }
--- a/assets/src/styles/blocks/Columns.scss
+++ b/assets/src/styles/blocks/Columns.scss
@@ -15,20 +15,22 @@
 
   h3, h3 > a {
     line-height: 1.6rem;
-    font-size: $font-size-lg;
-    font-family: $roboto;
-    font-weight: 700;
 
     --block-columns--column-heading-- {
+      font-size: $font-size-lg;
+      font-family: $roboto;
+      font-weight: 700;
       color: $grey-80;
-    }
+      margin-bottom: .5rem;
+      text-transform: none;
 
-    @include medium-and-up {
-      font-size: 1.5rem;
-    }
+      @include medium-and-up {
+        font-size: 1.5rem;
+      }
 
-    @include large-and-up {
-      font-size: 1.5rem;
+      @include large-and-up {
+        font-size: 1.5rem;
+      }
     }
   }
 
--- a/assets/src/styles/blocks/Covers/TakeActionCovers.scss
+++ b/assets/src/styles/blocks/Covers/TakeActionCovers.scss
@@ -154,12 +154,14 @@
       display: block;
 
       --block-covers--card-btn-- {
+        color: white;
         background-color: $orange;
         border-color: $orange;
       }
 
-      &:hover, &:focus, &:focus-within {
+      &:hover, &:focus {
         --block-covers--card-btn-hover-- {
+          color: white;
           background-color: $orange-hover;
           border-color: $orange-hover;
         }
@@ -268,13 +270,15 @@
 }
 
 .cover-card-btn {
+  --block-covers--card-button-- {
+    display: none;
+  }
   position: absolute;
   bottom: 0;
   left: 0;
   right: 0;
   margin: $n15 auto;
   width: 92%;
-  display: none;
   z-index: 1;
   pointer-events: all;
 }
--- a/assets/src/styles/blocks/Covers/TakeActionCovers.scss
+++ b/assets/src/styles/blocks/Covers/TakeActionCovers.scss
@@ -157,10 +157,8 @@
         color: white;
         background-color: $orange;
         border-color: $orange;
-      }
 
-      &:hover, &:focus {
-        --block-covers--card-btn-hover-- {
+        &:hover, &:focus {
           color: white;
           background-color: $orange-hover;
           border-color: $orange-hover;
@@ -242,6 +240,11 @@
 .cover-card-tag {
   --block-covers--card-tag-- {
     color: $yellow;
+
+    &:hover {
+      text-decoration: underline;
+      color: $yellow;
+    }
   }
 
   display: inline-block;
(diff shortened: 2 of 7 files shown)


assets/src/styles/blocks/Columns.scss
```diff
@@ -4,32 +4,28 @@
       font-weight: inherit;
     }
 
-    &.btn-primary {
```
Inwerpsel, line 7: In the default theme this is the same value as other primary buttons, so it seems this shouldn't be needed? If we don't remove this rule then all primary buttons variables are applied except the text always stays white.

assets/src/styles/blocks/Columns.scss
```diff
@@ -4,32 +4,28 @@
...
-    font-family: $roboto;
-    font-weight: 700;
 
     --block-columns--column-heading-- {
+      font-size: $font-size-lg;
+      font-family: $roboto;
+      font-weight: 700;
       color: $grey-80;
-    }
+      margin-bottom: .5rem;
```
Inwerpsel, line 20: Campaigns customize this bottom margin, but we didn't have a property for it. For now I added it with the value it effectively has.

assets/src/styles/blocks/Columns.scss
Inwerpsel: Another option is to make this one variable too, but in that case it might be better to give it its own button class instead of making adjustments to the primary button?

assets/src/styles/blocks/Covers/TakeActionCovers.scss
```diff
@@ -154,12 +154,12 @@
       display: block;
 
       --block-covers--card-btn-- {
```
Inwerpsel, line 156: All styles for the covers button are inside of the `:focus, ...` selector of the card. Probably we only need to put the `display` property in here so that other styles can be regardless of the hover.

assets/src/theme/useServerThemes.js
```diff
@@ -24,13 +24,17 @@ const deleteTheme = async (name) => {
...
+
   useEffect(() => {
     const doApiCall = async () => {
       const themes = await wp.apiFetch({
         path: 'planet4/v1/themes/',
         method: 'GET',
       });
-      setServerThemes(themes);
+      setServerThemes({
+        'default': {},
```
Inwerpsel, line 35: I added this as otherwise there's no easy way to start a new theme from the default theme.

assets/src/styles/blocks/Covers/TakeActionCovers.scss
```diff
@@ -268,13 +265,15 @@
 }
 
 .cover-card-btn {
+  --block-covers--card-button-- {
+    display: none;
```
Inwerpsel, line 269: I added this to test out toggling the hover behavior with a variable. It works, but because we put other styles only on the `:hover` selector, it looks very different. We could get around this like described in this comment: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/523/files?diff=split&w=1#r612283529

assets/src/styles/blocks/Covers/TakeActionCovers.scss
lithrel: I didn't get what you meant by 
> Probably we only need to put the display property in the pseudo selector so that other styles can be regardless of the hover.

Can you explain with images or code ? `.btn` has a different background and border color on hover, I don't get if/how you want to regroup styles there :thinking:

assets/src/styles/blocks/Covers/TakeActionCovers.scss
```diff
@@ -154,12 +154,12 @@
...
       --block-covers--card-btn-- {
+        color: white;
         background-color: $orange;
         border-color: $orange;
-      }
 
-      &:hover, &:focus, &:focus-within {
-        --block-covers--card-btn-hover-- {
+        &:hover, &:focus {
+          color: white;
```
lithrel, line 162: that's something I didn't realize before, but you have to explicitly duplicate a property so that it can be extended as a variable right ? like `color` is already defined in the parent, but it's here too for this reason ?

lithrel (approved):
:+1:

assets/src/styles/blocks/Covers/TakeActionCovers.scss
Inwerpsel: Yeah in this case both code and images are more explanatory :ok_hand: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/572 Also noticed that I used a different name in the one on the bottom. We could include these changes, however it adds some risk by changing CSS structure.

Currently if you remove the `display: none` rule, that hides the button when not hovered, it looks like this:
Screenshot from 2021-04-28 09-56-36

assets/src/styles/blocks/Covers/TakeActionCovers.scss
Inwerpsel: Yep, it's here because our base theme does give a separate color to background and border. For us white works in both cases, but that's not necessarily the case if you choose another background.

assets/src/styles/blocks/Covers/TakeActionCovers.scss
Inwerpsel: Ended up moving these rules already, after verifying that doesn't affect child themes.

assets/src/styles/blocks/Covers/TakeActionCovers.scss
Inwerpsel: See https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/523/files#r622006056

planet-4:

:new_moon: [nix](https://www-dev.greenpeace.org/test-nix)

:watch: 2021.04.28 16:11:07

PR #524, opened 2021-03-04
PLANET-5990 Enable Social Media block on Posts

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5990

---

Since native embed block doesn't support latest Facebook API changes anymore,
we need this in order to enable people to be able to embed FB/IG content
in Posts.

> _[Test post](https://www-dev.greenpeace.org/test-proteus/story/1007/social-media-test/)_

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -110,8 +110,9 @@ const POST_BLOCK_TYPES = [
 	'planet4-blocks/accordion',
 	'planet4-blocks/counter',
 	'planet4-blocks/gallery',
-	'planet4-blocks/take-action-boxout',
+	'planet4-blocks/social-media',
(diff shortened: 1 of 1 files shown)


mleray (approved):
Looks good to me 👍 But while checking the test post I noticed some small issues with the block:
- Hard to select for editing in the backend (starts the video on click)
- When changing the alignment the loading seems to be happening in two times, first it shows some text then loads again and finally shows the embed
- Speaking of the alignment, it's pretty useless since it's full width anyway (not sure if it's the case for all Social media block possibilities though)
Of course these are not blockers for this particular ticket, but maybe worth creating a ticket for? 🙂

comzeradd:
Thanks! All are valid points. I'll open a follow up ticket, since these affect the block on Pages too.

PR #528, opened 2021-03-10
Prevent unneeded prefixes, we autoprefix

Inwerpsel:
Ref: no ticket yet

---

Since we use autoprefixer, we can use a stylelintrule to detect any prefixes in our source that would be added during build anyway. This adds the rule and fixes violations, except `-webkit-box-orient` which is buggy.

### DIFF
--- a/assets/src/styles/blocks/Covers/TakeActionBoxoutBottom.scss
+++ b/assets/src/styles/blocks/Covers/TakeActionBoxoutBottom.scss
@@ -74,7 +74,11 @@
     margin-bottom: 0;
     display: -webkit-box;
     -webkit-line-clamp: 2;
+    /* stylelint-disable property-no-vendor-prefix */
+    /*! autoprefixer: off */
     -webkit-box-orient: vertical;
+    /*! autoprefixer: on */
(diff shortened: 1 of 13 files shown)


assets/src/styles/blocks/Covers/TakeActionBoxoutBottom.scss
```diff
@@ -74,7 +74,11 @@
     margin-bottom: 0;
     display: -webkit-box;
     -webkit-line-clamp: 2;
+    /* stylelint-disable property-no-vendor-prefix */
+    /*! autoprefixer: off */
```
Inwerpsel, line 78: I had to add this because of an issue with autoprefixer. If we add the rule to prevent webkit prefixes, it complains about this. But if I then replace it with the unprefixed version stylelint complains about the property being unknown. See https://github.com/postcss/autoprefixer/issues/776#issuecomment-297734206

I used this fix so that we can apply the rule without risking this behaving differently. We probably need to check this anyway because it's a legacy property.

mleray (approved):
Looks good to me 👍

PR #529, opened 2021-03-13
A11y: Convert anchor to button

aeisenberg:
Ref: 
Fixes https://github.com/greenpeace/planet4/issues/125

Converts one more anchor to a button. According to the linked issue,
this is the last one that should change.
---


mleray (dismissed):
Hi @aeisenberg, I've tested this on local, looks good! 👍 However, sorry if it still wasn't clear in the issue but there are actually more anchors that need to be converted... So can you please remove the `Fixes` in the PR description?

I'll update the issue with more examples 🙂

PR #530, opened 2021-03-15
PLANET-5983: Enable high level overview of loading code

Inwerpsel:

---
I need to make some changes in this file, but first it really needs a clean up. The code currently is hard to make sense of. The naming is not accurate, it's deeply nested and does a lot of unnecessary things.

I moved the error output to the respective functions that check them, so they don't need to return a boolean and can just die inside if requirements are not satisfied.

The goal is to make it possible to have a high level overview of what's happening in this file.

This relates to the ticket in the following way: we're moving parts of the styleguide into master theme to more easily change them. First we need to move parts of `editorStyle` into master theme to remove the dependency for the plugin. For that I need to replicate the loading of that file in master theme.

### DIFF
--- a/classes/class-loader.php
+++ b/classes/class-loader.php
@@ -11,10 +11,10 @@ namespace P4GBKS;
 use P4\MasterTheme\Features;
 use P4\MasterTheme\MigrationLog;
 use P4\MasterTheme\Migrations\M001EnableEnFormFeature;
+use P4GBKS\Controllers;
+use P4GBKS\Views\View;
 use WP_CLI;
 use P4GBKS\Command\Controller;
-use P4GBKS\Controllers\Ensapi_Controller;
-use P4GBKS\Controllers\Menu\Enform_Post_Controller;
 
 /**
  * Class Loader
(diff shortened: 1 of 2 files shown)


classes/class-loader.php
```diff
@@ -33,50 +33,18 @@ final class Loader {
 	 * @var Loader $instance
 	 */
 	private static $instance;
-	/**
```
Inwerpsel, line 36: All of these were really not needed. 2 just had a constant assigned to them and constant could be used instead. The others were just used locally in one place so there's no need to put them on an instance.

classes/class-loader.php
```diff
@@ -87,39 +52,34 @@ public static function get_instance( $services, $view_class ) : Loader {
...
-			new Blocks\SplitTwoColumns(),
-			new Blocks\Spreadsheet(),
-			new Blocks\SubMenu(),
-			new Blocks\SubPages(),
-			new Blocks\TakeActionBoxout(),
-			new Blocks\Timeline(),
-			new Blocks\SocialMediaCards(),
-			new Blocks\ENForm(),
-		];
+		new Blocks\Accordion();
```
Inwerpsel, line 64: No point keeping these instances in memory if they're never used, which was the case for `$this->blocks`

classes/class-loader.php
```diff
@@ -153,19 +113,24 @@ function ( $class_name ) {
...
-	 * @param array  $services The Controller services to inject.
-	 * @param string $view_class The View class name.
+	 * Add some admin pages.
 	 */
-	public function load_services( $services, $view_class ) {
-		$this->services = $services;
-		$this->view     = new $view_class();
+	public function load_services(): void {
+		$services = [
+			Controllers\Menu\Settings_Controller::class,
```
Inwerpsel, line 120: These were being passed around from `get_instance` to `__construct` to `load_services` to the private property `$services`, to then only be used here.

Inwerpsel: Same for `$view`

lithrel (approved):
:+1:

Inwerpsel:
I started following Sonarcloud's hint to make this static. I turned `__construct` into a static register function. But then I needed to change all other functions to be static as well. Everything could be made static without changes because nothing actually used the instance. But it's [a lot of changes](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/compare/unspaghettify...staticify?expand=1) and I initially just wanted to refactor 1 function :smile: If you want to review it, that's great. Otherwise I'm fine with just bypassing Sonarcloud for now, since I didn't introduce the violation.

lithrel:
I just pulled it and stumbled on a typo crashing the site there https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/532/files#diff-b0e752e847c025f266c46aadf983955b5796c943773645980d725d628c73120fR54 , so let's merge this one first, I understand why Sonar is protesting but we can take more time to write and test the static change later :grin: 
And I see the usefulness of static for this register method, if you want an intermediate solution with less changes you can keep the instance first, using `(new Foo_Block())->register_block();` another solution would be to use an anonymous function for the render callback and use an instance of the object `function () { return (new Foo_Block())->render(); }`, as it seems to make sense and would only generate an object instance on use. _But_ that's for a bit later, so we have time to test :pray:

PR #534, opened 2021-03-16
[No issue]: Check for a truish value in err before logging it

pablocubico:
This is just to remove a `null` string in the Webpack build logs.

```
[./assets/src/theme/initializeThemeEditor.js] 6.42 KiB {themeEditor} [built]
    + 504 hidden modules
null    <---- this thingy
```


lithrel (approved):
:partying_face: 
<sub>edit: same issue in master-theme</sub>

PR #536, opened 2021-03-19
PLANET-5986 Disable Take Action Boxout from Pages

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-5986


lithrel (approved):
:+1:

PR #539, opened 2021-03-24
PLANET-4802 Columns block WYSIWYG

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-4802

### Testing

Changes can be tested for instance on [this page](https://www-dev.greenpeace.org/test-deimos/all-the-columns/)


comzeradd:
You can also fix the failing test, since tests are now part of this repository. Markup has changed a bit (`container` now only wraps the `<header>`, thus breaking the Image style test.

assets/src/blocks/Columns/ColumnsEditor.js
```diff
@@ -0,0 +1,185 @@
...
+            <RichText
+              tagName='h2'
+              className='page-section-header'
+              placeholder={__('Enter title', 'planet4-blocks-backend')}
+              value={columns_title}
+              onChange={toAttribute('columns_title')}
+              keepPlaceholderOnFocus={true}
+              withoutInteractiveFormatting
+              allowedFormats={[]}
+              characterLimit={40}
```
lithrel, line 149: I realized this week (well React insulted me) that `characterLimit` is not a RichText property (and [doesn't look like](https://github.com/WordPress/gutenberg/issues/22972) it's gonna be) but [our own thing](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/components/withCharacterCounter/withCharacterCounter.js). Not sure what to do with this information.

assets/src/blocks/Columns/ColumnsEditor.js
```diff
@@ -0,0 +1,185 @@
...
+  const removeColumn = () => setAttributes({ columns: columns.slice(0, -1) });
+
+  if (!isSelected) {
+    return null;
+  }
+
+  return (
+    <>
+      <InspectorControls>
+        <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
```
lithrel, line 45: `Settings` ?

lithrel (commented):
Editing an existing (created before this PR) Columns block always shows the "No image" style selected, even if it has a different style. That's probably because `className` is not set in old blocks, we could set it if it does not exist.

There's also a minor issue that's present only on dev database, because of some `attachment` attributes being empty strings. It's minor because I couldn't find any occurrence of this in any NRO, so I don't think it's problematic or needs modification; for dev db it leads to those blocks not being displayed on frontend, see pages like "Press center" (we would need to replace the php declaration `'type'    => 'integer'` with `'anyOf' => ['type' => 'integer', 'type' => 'string']`)

<sub>Continuing tomorrow, looking good :+1: </sub>

assets/src/blocks/Columns/ColumnsEditor.js
mleray: It's been bugging me as well, but we write `Setting` for other blocks so I kept it 😅

assets/src/blocks/Columns/ColumnsEditor.js
mleray: Yes I've noticed that too 😁 not sure what to do about it either... It's in so many of our blocks at the moment! From a design perspective it's tempting to add character limits, but we support so many languages that I'm not sure it makes sense in our case, maybe we should just remove this all together.

mleray:
> Editing an existing (created before this PR) Columns block always shows the "No image" style selected, even if it has a different style. That's probably because `className` is not set in old blocks, we could set it if it does not exist.

Thanks for checking @lithrel 💚 I've just pushed a fix for this issue!

lithrel (approved):
:+1: cool cool cool

I was checking the `.img-wrap` usage elsewhere, it is used by social media cards in the editor but that block is actually broken so :shrug: 
Also worth noting that the option "Open in a new tab" is valid for local links, but external links will always be opened in a new tab (because of external_link script) and we use this in other blocks too, maybe we should centralize this component and add a note, or make it disabling itself for external links, someday.

mleray:
> Also worth noting that the option "Open in a new tab" is valid for local links, but external links will always be opened in a new tab (because of external_link script) and we use this in other blocks too, maybe we should centralize this component and add a note, or make it disabling itself for external links, someday.

That's a good point, especially because at the moment it only affects certain Columns styles (some use buttons) but once we update the block according to the design system all will use links so it will be really necessary to do something about this! I'll ping Mag about it to get her recommendations 🙂

planet-4:

:new_moon: [deimos](https://www-dev.greenpeace.org/test-deimos)

:watch: 2021.04.28 13:52:00

PR #541, opened 2021-03-25
PLANET-5983 Theme as submodule

Inwerpsel:

---

Same story as the other styleguide move PRs here, but more files to adjust.

We should probably check for steps so that everyone can easily ignore these folders in their IDE. For me (PHPStorm) it involved marking the entire `master-theme` directory as excluded, then marking the styleguide folder separately as non excluded. You could exclude the whole thing if you only want to see the styleguide files in the master theme itself. Probably works the same way in VSCode.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "assets/src/styles/styleguide"]
-	path = assets/src/styles/styleguide
-	url = https://github.com/greenpeace/planet4-styleguide
+	path = assets/src/styles/master-theme
+	url = https://github.com/greenpeace/planet4-master-theme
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "assets/src/styles/styleguide"]
```
Inwerpsel, line 1: Preserving this name will probably make it easier to pull in the changes, since it just needs to update the remote and path.

Inwerpsel: Github seems to conflate the folder name and the submodule name :confused: 
Screenshot from 2021-03-26 11-13-58

Inwerpsel: So probably `[submodule "assets/src/styles/styleguide"]` is just a unique identifier within `.gitmodules`.

comzeradd (approved):
Looks good :+1:

PR #548, opened 2021-04-06
PLANET-5983: Properly extract vars from multiple files

Inwerpsel:

---

The previous version of this plugin didn't account for multiple entrypoints. To resolve this the new version now writes to a separate key per entrypoint.

The plugin to combine these and write the file is now also included. Before this was in both repos using this package.

This also disables extracting the variables from the source SCSS files. Nothing was done with the result yet, and probably the same can be achieved directly in the postcss plugin by ensuring the rules source is properly set.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1292,6 +1292,35 @@
         }
       }
     },
+    "@greenpeace/planet4-postcss-css-variables": {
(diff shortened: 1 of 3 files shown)


lithrel (approved):
:+1:

PR #551, opened 2021-04-13
PLANET-5523 Add WYSIWYG Covers block to beta blocks

mleray:
### Description

- Remove `load_more.js` (jQuery): [PLANET-5764](https://jira.greenpeace.org/browse/PLANET-5764)
- Convert Covers block to WYSIWYG: [PLANET-5523](https://jira.greenpeace.org/browse/PLANET-5523)

There is another PR (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/486) open for these tickets, that would directly make the change and remove the old block, but it needs to be rebased and updated quite often until it's 100% ready... This would allow us to release the WYSIWYG version of the block in the beta blocks, while keeping the older version of the block until we are fully satisfied with the new one.


mleray:
Another approach was chosen and merged into https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/486

PR #552, opened 2021-04-13
PLANET-5983 Make CSS variables file work with WPML

Inwerpsel:
Ref: 

---

Just a quick fix to make it usable when WPML is installed and `document.body.dataset.nro` contains the language suffix. This removes what looks like a language suffix from the end, except for NL (see comment).


lithrel (approved):
:+1:

PR #554, opened 2021-04-14
No ticket - Fix Table block placeholder

mleray:
### Description

I'm not sure where this change came from and when (it's not from us, seems to be from WP's core files) but there is now `    flex-direction: column;` added to this placeholder and it looks weird. So this PR just makes a quick fix for it!

- Broken:
Screenshot 2021-04-14 at 14 08 33

- Fixed:
Screenshot 2021-04-14 at 14 14 49

### DIFF
--- a/assets/src/styles/blocks/core-overrides/TableEditor.scss
+++ b/assets/src/styles/blocks/core-overrides/TableEditor.scss
@@ -1,0 +1,3 @@
+.blocks-table__placeholder-form {
+  flex-direction: row !important;
+}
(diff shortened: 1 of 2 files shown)


sagarsdeshmukh:
Hi @mleray , I am trying to reproduce the issue on below test instances but won’t reproduce. do you have example of issue on test instance?
https://www-dev.greenpeace.org/international/wp-admin/post-new.php?post_type=page
https://www-dev.greenpeace.org/sagardev/wp-admin/post.php?post=46471&action=edit

mleray:
That's apparently just on my side, so nevermind 😁

PR #555, opened 2021-04-15
PLANET-5523 Move old Covers block

mleray:
This is a preparatory PR to simplify the addition of the new WYSIWYG version of this block in the beta blocks. Then we can just rebase the WYSIWYG PR (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/486) and register the new version of the block with another name (i.e. `Covers WYSIWYG`) in the beta blocks, but already using the right folder/file names.


classes/class-loader.php
```diff
@@ -95,7 +95,10 @@ function ( $class_name ) {
 					if ( false !== strpos( $class_name, 'P4GBKS' ) ) {
 						$class_name_parts = explode( '\\', $class_name );
 						$real_class_name  = array_pop( $class_name_parts );
-						$file_name        = 'class-' . str_ireplace( '_', '-', strtolower( $real_class_name ) );
+						if ( 'covers' === $real_class_name ) {
```
Inwerpsel, line 98: Probably not needed, since currently `$real_class_name` doesn't match this and it still works :smile:

Inwerpsel (commented):
We might need to do something for files that are both in the old and new version. Otherwise Git will want to merge the rename and the new version's code. Maybe we should already split the files you're going to re-use in this PR :thinking:

Inwerpsel:
> We might need to do something for files that are both in the old and new version. Otherwise Git will want to merge the rename and the new version's code. Maybe we should already duplicate the files you're going to re-use in this PR 
> 
> Duplicating a file can be done this way: https://stackoverflow.com/a/44036771/4961158

Could be keeping the history only for the new version is simpler.

mleray:
Another approach was chosen and merged into https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/486

PR #556, opened 2021-04-15
PLANET-6080 Move fonts out of CSS file

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6080

---
* Moved to [HTML head in the theme](https://github.com/greenpeace/planet4-master-theme/pull/1364/files#diff-a40ff5ae673d9e734e86aef984e9588608dee8d8e31ee3da3a1a5b845101ea3cR3-R5), so that Cloudflare can optimize them.
* Preserve in a separate file so the editor can still load them.

### DIFF
--- a/assets/src/styles/editorStyle.scss
+++ b/assets/src/styles/editorStyle.scss
@@ -26,6 +26,7 @@
 @import "components/Preview";
 @import "components/CharacterCounter";
 @import "campaigns/base/fonts";
+@import "campaigns/base/google_fonts";
 @import "components/FormSectionTitle";
 @import "components/FormHelp";
 @import "components/InlineFormFeedback";
--- a/assets/src/styles/campaigns/base/_fonts.scss
+++ b/assets/src/styles/campaigns/base/_fonts.scss
@@ -35,9 +35,3 @@
   font-weight: bold;
   font-style: italic;
 }
-
-@import url("https://fonts.googleapis.com/css?family=Montserrat:300,400,500,600,700,800");
-@import url("https://fonts.googleapis.com/css?family=Kanit:400,500,600,800");
-@import url("https://fonts.googleapis.com/css?family=Open+Sans:400,500,600,800");
-@import url("https://fonts.googleapis.com/css?family=Anton:400,500,600,800");
-@import url("https://fonts.googleapis.com/css?family=Amiko:400,500,600,800");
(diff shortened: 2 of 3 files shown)


assets/src/styles/editorStyle.scss
```diff
@@ -26,6 +26,7 @@
 @import "components/Preview";
 @import "components/CharacterCounter";
 @import "campaigns/base/fonts";
+@import "campaigns/base/google_fonts";
```
lithrel, line 29: Should we add the same to [`assets/src/styles/campaigns.scss`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/f918ee7627cfe4329b202d5ec33d1c122ccfe887/assets/src/styles/campaigns.scss#L11) ?

lithrel (approved):
:+1: 
While checking this I realized the title of the Antarctic campaign in the editor doesn't have the Sanctuary font applied, and I clearly remember seeing it before, I wonder when we lost this or if it was on purpose, or if I imagined things ?

Inwerpsel:
> While checking this I realized the title of the Antarctic campaign in the editor doesn't have the Sanctuary font applied (doesn't come from this PR), and I clearly remember seeing it before, I wonder when we lost this or if it was on purpose, or if I imagined things ?

Probably [this part that relies on `DOMContentLoaded`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/5df0e0378749f1d5fd0b03f22ca545152ba953dc/assets/src/connectCssVariables.js#L46-L52) got broken.

Inwerpsel:
> > While checking this I realized the title of the Antarctic campaign in the editor doesn't have the Sanctuary font applied (doesn't come from this PR), and I clearly remember seeing it before, I wonder when we lost this or if it was on purpose, or if I imagined things ?
> 
> Probably [this part that relies on `DOMContentLoaded`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/5df0e0378749f1d5fd0b03f22ca545152ba953dc/assets/src/connectCssVariables.js#L46-L52) got broken with a WP version update that caused the event to fire earlier. There's definitely a better way to write it.

I made it work again locally, but now it works too well :grin: 
Screenshot from 2021-04-21 17-12-50

PR #557, opened 2021-04-16
PLANET-5983 Follow variable name change

Inwerpsel:

---

Needs to be updated to track the new format of generated variable names https://github.com/greenpeace/planet4-master-theme/blob/d1189ab0148b2ab6c35ffc0b67ae3788224fdb3d/assets/src/scss/base/_typography.scss#L134-L138


mleray (approved):
Looks good to me 👍

PR #561, opened 2021-04-17
PLANET-6087 Move SubPages, Spreadsheet and Accordion to stable category

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6087

---

All 3 blocks are already used in production so we'll move them to the normal "Planet 4 Blocks" category so they stay usable.


comzeradd:
Let's do the same for the `sub-page` block. It may not be that popular, but it's part of the Campaign Themes feature.

mleray (approved):
👍

PR #563, opened 2021-04-19
PLANET-4535: Split Timeline block scripts

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4535
Following https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/508/

We now split blocks each in their own script, instead of merging everything  in `(frontend|editor)Index.js`.

This PR mostly renames files, and adds a render function. The render function could be generalized and made as a dependency of other scripts (as it's gonna happen for hydration I guess ?).
- `TimelineScript.js` instantiate a Timeline on the frontend side
- `TimelineStyle.scss` is used for frontend instance
- `TimelineEditorScript.js` instantiate a Timeline editor, in the gutenberg editor
- `TimelineEditorStyle.scss` is used for editor specific styles

## Test

Timeline block should work as usual.
Here's a googlesheet to test https://docs.google.com/spreadsheets/d/1XruuRzjiu4t2ia7NaBXMqItr4NAbc7ctcolg1uhlIL4/edit#gid=0
- Create a page, insert a Timeline block with the test spreadsheet
- Check in your browser network console that Timeline scripts are included individually:
  Screenshot from 2021-04-19 09-13-14
- On preview/published page, check that Timeline scripts are included individually
Screenshot from 2021-04-19 09-16-55
- Timeline scripts shouldn't be included in any page without a Timeline in it (default Homepage for example)


classes/blocks/class-timeline.php
```diff
@@ -59,6 +63,9 @@ public function __construct() {
 				],
 			]
 		);
+
+		add_action( 'enqueue_block_editor_assets', [ self::class, 'enqueue_editor_assets' ] );
```
Inwerpsel, line 67: This function currently will load a fixed set of files, even if there's nothing to put in them. We should probably split it up so that we don't need to serve an empty file for this block https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/563/files#diff-7703839d1af73e6cde77fab17eeb77ed2f90c57a4559c3f45bf73795178e1fae

lithrel: I guess I need to adapt `entryPoint` function in the webpack config too

Inwerpsel: Indeed. Would be nice if we could somehow write both in the same place :thinking:

assets/src/blocks/Timeline/TimelineFrontend.js
```diff
@@ -1,25 +1,24 @@
-import { Fragment } from '@wordpress/element';
 import { Timeline } from './Timeline';
```
Inwerpsel, line 1: Not crucial for this block, since the resulting size is only 5.6KB (2.3KB compressed), but we could check whether there are things that are imported in these chunks that are actually already loaded by the main index file. Some things like `useEffect` are resolved to globals anyway, but others might be included in each entrypoint if they are imported?

classes/blocks/class-timeline.php
Inwerpsel: Though maybe not worth it as these things will not change that often.

Inwerpsel: We could have a function per enqueue, so `enqueue_editor_scripts`, `enqueue_editor_styles`, `enqueue_frontend_scripts`, `enqueue_frontend_styles`.

Inwerpsel:
Timeline currently loads an external CSS and JS file from within the code. If we set up code splitting, we could enqueue both as a dependency of our CSS and JS file. They were loaded through code to avoid always loading them even if the block is not on the page.
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/4cad017c7d8d95614a1f4f166e8c6513dab86cb5/assets/src/blocks/Timeline/Timeline.js#L18-L20
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/4cad017c7d8d95614a1f4f166e8c6513dab86cb5/assets/src/blocks/Timeline/Timeline.js#L39-L42

lithrel:
> we could enqueue both as a dependency of our CSS and JS file

Yeah but there's a small bit of code there to revert the lodash issue, I'd prefer to keep everything together

Inwerpsel:
> > we could enqueue both as a dependency of our CSS and JS file
> 
> Yeah but there's a small bit of code there to revert the lodash issue, I'd prefer to keep everything together

It's not just a matter of where the code lives. If we enqueue them through WP they will be in the DOM when the browser receives the initial HTML. Whereas currently the block "manually" does requests for both resources inside the script, which will get initiated only after the first time React has rendered the block (because it uses `useEffect`).

lithrel:
@Inwerpsel  Ok, moved frontend script/style enqueue to php class, and dynamic script loading to TimelineEditor. 
I had to move all `use*()` calls to the main TimelineEditor component because if they are distributed in `renderEdit()` and `renderView()`, React complains about hook rules (hook calls are not the same/in the same order depending on block state) and breaks :panda_face:

planet-4:

:new_moon: [telesto](https://www-dev.greenpeace.org/test-telesto)

:watch: 2021.05.12 06:54:58

assets/src/blocks/Timeline/TimelineFrontend.js
lithrel: I took a look at the compiled code, I think wp.element is registered as external and so not included in this script (there are a bunch of dependencies noted as external) - not sure how it works exactly, but we can take a look at the other scripts when we split them

Inwerpsel:
> @Inwerpsel Ok, moved frontend script/style enqueue to php class, and dynamic script loading to TimelineEditor.
> I had to move all `use*()` calls to the main TimelineEditor component because if they are distributed in `renderEdit()` and `renderView()`, React complains about hook rules (hook calls are not the same/in the same order depending on block state) and breaks

Yep, React complains rightly so :smile: Hooks are implemented as a linked list for performance, so it fundamentally cannot work if they're not called in the same order. The source code of hooks is actually the best explanation, which is really only a handful of lines in most cases.

https://github.com/facebook/react/blob/48907797294340b6d5d8fecfbcf97edf0691888d/packages/react-debug-tools/src/ReactDebugHooks.js#L127-L136

https://github.com/facebook/react/blob/48907797294340b6d5d8fecfbcf97edf0691888d/packages/react-debug-tools/src/ReactDebugHooks.js#L95-L102

I think for us this means it's probably better to not use a regular function for `renderEdit` and `renderView`, but instead either put everything together into `render`, or use a component instead, which would allow us to put the hooks there ([like here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/47e573dd5d41c168da606d0c6e82df6642d19671/assets/src/blocks/Splittwocolumns/SplittwocolumnsEditor.js#L48-L53)).

assets/src/blocks/Timeline/TimelineFrontend.js
Inwerpsel: Yeah should be fine for this one :+1:

Inwerpsel (approved):
Looks good, left a suggestion for readability of the file name templates. If you're up for it we should probably also try to remove the `lodash` dependency :fire:

PR #564, opened 2021-04-19
PLANET-6086 Code splitting Covers block

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6086

---

[Easier with whitespace ignored](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/564/files?diff=split&w=1).

Code splitting, set up like for Accordion block. The way it currently works I [had to hard code an exception](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/564/files#diff-b0e752e847c025f266c46aadf983955b5796c943773645980d725d628c73120fR140) for locating the block. We'll reconsider the logic to specify these files anyway, probably in the PR for code splitting of the Timeline block ([see comment](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/563/files#r615690514)), so at that point we'll be able to remove the exception.

Note the CSS files are currently still empty. This is because both the old and new version use the same CSS, so it's already loaded in the main `style.css`. When we replace the old with the new block we can move these to the respective entrypoints.

To test: both the old and new block should work properly in the editor and front end (styles + scripts are loaded, no 404s).
https://www-dev.greenpeace.org/test-saturn/explore/


assets/src/blocks/Covers/CoversEditorScript.js
```diff
@@ -97,3 +97,4 @@ export const registerCoversBlock = () => {
   });
 }
```
Inwerpsel, line 99: I first removed the function call around it, but it's easier if we preserve it and call it so the file is similar enough during diff/merge.

assets/src/styles/blocks/Covers/CoversEditorStyle.scss
```diff
@@ -0,0 +1 @@
+// Empty for now.
```
Inwerpsel, line 1: These 2 can be empty because both the old and new version use the same CSS, so it's included in our bundle anyway. Once we removed the old version we can easily move the CSS into here. I added the file because the code splitting is currently set up to require each type of file be present, and I didn't want to get blocked by rewriting it.

classes/blocks/class-covers.php
```diff
@@ -58,7 +58,7 @@ class Covers extends Base_Block {
 	 */
 	public function __construct() {
 		register_block_type(
-			self::get_full_block_name(),
+			'planet4-blocks/covers-beta',
```
Inwerpsel, line 61: If the `BLOCK_NAME` constant is changed, that affects all file names of scripts being loaded with the current logic. We only want to temporarily register the block to WP with a different name, so I changed only this usage. Because of that I had to [add a little exception in the base block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/564/files#diff-b0e752e847c025f266c46aadf983955b5796c943773645980d725d628c73120fR140). It's not pretty, but we can revisit this afterwards and remove the workaround at that point, so that we have something that's easier to apply for all blocks.

assets/src/styles/blocks/Covers/CoversEditorStyle.scss
mleray: Shouldn't this change also be applied [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/67d611e28d82545caa21864a14a8cdb61725bb28/classes/blocks/class-covers.php#L83)? 🤔

Inwerpsel: Right, it puts the wrong thing there :+1:

Inwerpsel: We can already make this use the final name, the previous version didn't use `data-render`, so it won't collide. So I updated this line instead https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/564/files#diff-f4f3a78077b528d85adf10a5c03acfa82930b2de5f93086b8d375181b083c9adR3

assets/src/blocks/Covers/CoversScript.js
```diff
@@ -0,0 +1,8 @@
+import { CoversFrontend } from './CoversFrontend';
+
+document.querySelectorAll( `[data-render='planet4-blocks/covers']` ).forEach(
```
Inwerpsel, line 3: Also has the advantage that we can migrate `planet4-blocks/covers-beta` to `planet4-blocks/covers` without changing the block's inner content.

mleray (approved):
Looks good to me 👍

PR #565, opened 2021-04-20
Improve ordering and exclude bootstrap variables

Inwerpsel:
Ref: 

---

Bootstrap variables mostly have no effect for us, and the selectors match a lot of elements. For now it's better to exclude them.

Sort by property name first instead of last, which makes it easier to find things in the list.

Previous ordering:

New ordering:


lithrel (approved):
:+1:

PR #566, opened 2021-04-21
WIP make custom fonts work again in editor

Inwerpsel:
* For now this causes the titles in the editor sidebar to follow the
style too, which looks interesting but probably not what we want.
* It probably broke with a WP version update that changed the
DOMContentLoaded event the function relied on to add the event
listeners. It now gets called when the post type is not there yet.
* Instead now it checks this in the listener. Since it just needs to
read an in memory string, it can get called frequently.
* It was broken a second time (but not noticed since it already was
broken) because the variable names changed. Fixed these too.

Ref: 

---

@lithrel Spotted this while reviewing https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/556

### DIFF
--- a/assets/src/connectCssVariables.js
+++ b/assets/src/connectCssVariables.js
@@ -13,11 +13,11 @@ const getMeta = () => {
 const metaToVariableMapping = [
   {
     metaKey: 'campaign_body_font',
-    cssVariable: '--body-font',
+    cssVariable: '--body--font-family',
   },
   {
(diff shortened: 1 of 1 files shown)


Inwerpsel:
This will be fixed for the converted themes after merging https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/589 . That PR would also make all variables work in the editor, not just the fonts.

PR #567, opened 2021-04-23
Responsive theme editor

Inwerpsel:
Ref: https://docs.google.com/document/d/1-UX-5TNLow9766yaTGUa0I2KxpfeUtgl6sf0iCizWoM/edit

---

Added a new mode that finally allows to theme for smaller screen sizes (or bigger). If the mode is enabled, the current page is shown in an iframe. You can change the iframe's width and height, choose one of the presets, and scale it down in case it doesn't fit on your screen.

Temporarily added a commit so it can be used without logging in on my instance.

Also:
* Fix broken "minus" icon button. I used another technique of adding a button with a minus sign: a button with a minus sign in it.

TODO:
* Tune the frame's width and height offsets (to account for the scrollbar size) so that the inner frame gets exactly the chosen resolution. I just eyeballed this for now.
* Give more space to the sidebar. It's still the same dragable overlay as I created for the first "in-page" version.
* Make the theme editor more "aware" of media queries if a variable is used in one. So that we can show only the ones applying to the current screen size.

### DIFF
--- a/assets/src/styles/components/VarPicker.scss
+++ b/assets/src/styles/components/VarPicker.scss
@@ -1,7 +1,50 @@
+html,
+body {
+  scroll-behavior: auto !important;
+}
+
 .var-picker * {
   font-family: Roboto, sans-serif !important;
   color: black !important;
   line-height: initial !important;
+}
+
+.simulating-touch-device body {
+  &::-webkit-scrollbar {
+    width: 9px;
+  }
+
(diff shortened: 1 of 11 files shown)


Inwerpsel:
Moved these changes to the new separate repository https://github.com/Inwerpsel/use-theme-editor. See https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/578

PR #568, opened 2021-04-23
PLANET-4809 Copy CarouselHeader to OldCarouselHeader

mleray:
This PR is to prepare the CarouselHeader block for adding the WYSIWYG version (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/483) to beta blocks.

### DIFF
--- a/assets/src/styles/blocks.scss
+++ b/assets/src/styles/blocks.scss
@@ -1,8 +1,8 @@
 // P4 Blocks
 @import "blocks/Articles";
 @import "blocks/Columns";
-@import "blocks/CarouselHeader";
-@import "blocks/CarouselHeader_full_width_classic";
+@import "blocks/OldCarouselHeader";
+@import "blocks/OldCarouselHeader_full_width_classic";
 @import "blocks/Covers/TakeActionCovers";
 @import "blocks/Covers/CampaignCovers";
 @import "blocks/Covers/ContentCovers";
--- a/assets/src/styles/editorStyle.scss
+++ b/assets/src/styles/editorStyle.scss
@@ -13,8 +13,8 @@
 @import "blocks";
 
 @import "blocks/ArticlesEditor";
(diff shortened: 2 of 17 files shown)


assets/src/styles/blocks.scss
```diff
@@ -3,6 +3,8 @@
 @import "blocks/Columns";
 @import "blocks/CarouselHeader";
 @import "blocks/CarouselHeader_full_width_classic";
```
Inwerpsel, line 5: You can remove these

assets/src/styles/editorStyle.scss
```diff
@@ -15,6 +15,8 @@
 @import "blocks/ArticlesEditor";
 @import "blocks/CarouselHeaderEditor";
```
Inwerpsel, line 16: We can remove these as well, and re-add them when we merge wysiwyg beta (or not if we do code splitting)

planet4-gutenberg-blocks.php
```diff
@@ -138,6 +138,7 @@
 
 const BETA_PAGE_BLOCK_TYPES = [
 	'planet4-blocks/covers-beta',
+	'planet4-blocks/carousel-header-beta',
```
Inwerpsel, line 141: Let's only add them when merging the wysi PR? Otherwise this will be an identical version to the previous :smile:

Inwerpsel (commented):
Can you remove the previous imports? :pray: Otherwise they increase the size of what we send to visitors.

planet4-gutenberg-blocks.php
Inwerpsel: Although just adding them here does not allow them in the editor. So we can actually already add it I guess.

mleray: Removed 🙂

Inwerpsel (approved):
:100:

PR #569, opened 2021-04-26
Articles paginate with after

Inwerpsel:
Ref: 

---


Inwerpsel:
I'll close this as this way of paginating only makes sense if there's no numeric pagination, like the load more button had.

It may be possible still to get numeric pagination this way, which could have caching advantages. But those advantages are not that big, especially because our cache is frequently invalidated by code changes.

PR #572, opened 2021-04-28
Add styles outside hover selector

Inwerpsel:
* Currently the button is hidden if the card is not hovered. So the
styles can be added to the regular selector since they're not visible
anyway.
* This allows to have the button be visible (and styled) by just
changing the `display: none` property. Before this change this would
result in a partially unstyled button.

Ref: 

---


Inwerpsel:
Included these changes in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/523/files

PR #575, opened 2021-04-29
PLANET-6125 Add 50 Years Guestbook block

mleray:
### Description

See [PLANET-6125](https://jira.greenpeace.org/browse/PLANET-6125)

### Testing

The "50 Years GuestBook" block should now be available in Planet 4 blocks, and it should show the iframe properly in both the backend and frontend 🙂
If you want to make the iframe full-width, you can add the `block-wide` class in the sidebar in the additional classes!

**Question**: I'm not sure whether showing the actual iframe in the editor is the best, it makes it a bit difficult to select the block (you need to click right above it), maybe we should just show an image of the map there? 🤔 On the other hand it's nice to see how it will look like in the frontend and test the interactive aspect...

### DIFF
--- a/assets/src/blocks/GuestBook/GuestBookBlock.js
+++ b/assets/src/blocks/GuestBook/GuestBookBlock.js
@@ -1,0 +1,22 @@
+import { frontendRendered } from '../frontendRendered';
+const { registerBlockType } = wp.blocks;
+const { __ } = wp.i18n;
+
+const BLOCK_NAME = 'planet4-blocks/guestbook';
+
+export const registerGuestBookBlock = () =>
+  registerBlockType(BLOCK_NAME, {
+    title: __('50 Years GuestBook', 'planet4-blocks-backend'),
+    icon: 'admin-site-alt2',
+    category: 'planet4-blocks',
+    supports: {
+      html: false, // Disable "Edit as HTMl" block option.
+      multiple: false, // Use the block just once per post.
+    },
+    edit: () => (
+      <p className='EmptyMessage'>
+        {__('This block only renders in the frontend', 'planet4-blocks-backend')}
+      </p>
+    ),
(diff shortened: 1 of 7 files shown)


planet-4:

:new_moon: [pandora](https://www-dev.greenpeace.org/test-pandora)

:watch: 2021.04.29 14:36:42

assets/src/blocks/GuestBook/GuestBook.js
```diff
@@ -0,0 +1,15 @@
+export const GuestBook = () => {
+  const buildURL = () => {
+    const hostname = window.location.hostname;
+    const pathname = window.location.pathname.slice(1);
```
comzeradd, line 4: One thing we need to think here, is that this is generated on the editor, so the `pathname` is containing `wp-admin`.

For instance, if you check a story on [this page|https://www-dev.greenpeace.org/test-pandora/story-map/], pick one of the stories and click on the copy icon just to get the share url. It's `https://www.greenpeace.org/test-pandora/wp-admin/post-new.php?id=1`

comzeradd:
> I'm not sure whether showing the actual iframe in the editor is the best, it makes it a bit difficult to select the block (you need to click right above it), maybe we should just show an image of the map there?

I remember the native Facebook embed (removed in 5.6) was doing something similar. It was just printing an error saying this block only render in the frontend.

assets/src/blocks/GuestBook/GuestBook.js
mleray: I think this issue is solved now! 🙂

sagarsdeshmukh (approved):
Tested on local and [test](https://www-dev.greenpeace.org/test-pandora/test-guestbook-block/) instance. 
PR looks good to me 💯

PR #576, opened 2021-04-30
PLANET-6129 Extract vars at runtime

Inwerpsel:
* Previously this required a build time script. The script had multiple
issues and complicated our setup. The same info can easily (and quickly)
be extracted at runtime by the theme editor script.

Ref: 

---


assets/src/theme/balancedVar.js
```diff
@@ -0,0 +1,41 @@
+const balanced = require( 'balanced-match' );
```
Inwerpsel, line 1: This was entirely copied from https://github.com/greenpeace/planet4-postcss-css-variables/blob/master/lib/balanced-var.js, still need to make it use `import` syntax.

assets/src/theme/extractPageVariables.js
```diff
@@ -0,0 +1,67 @@
+const balancedVar = require('./balancedVar');
+
+const isSameDomain = ({ href }) => !href || href.indexOf(window.location.origin) === 0;
+const isNotCoreFile = ({ href }) => !href || !href.includes('wp-includes')
```
Inwerpsel, line 4: For now I excluded the core files because they have their own variables which won't work for us. Not sure why those stylesheets are loaded :thinking:

planet-4:

:watch: 2021.04.30 11:50:20

assets/src/theme/extractPageVariables.js
```diff
@@ -0,0 +1,67 @@
...
+        }
+        collected[variableName].usages.push(usage);
+        // Replace variable name (first occurrence only) from result, to avoid circular loop
+        remainingValue = (match.pre || '') + match.body.replace(variableName, '') + (match.post || '');
+      }
+    })
+    return collected;
+  }
+  if (rule.type === 4) {
+    // No nested media queries for now.
```
Inwerpsel, line 39: We don't have media queries inside of media queries atm, so I kept it simple for now. Same for `@supports`

assets/src/theme/extractPageVariables.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const collectSheetVars = (vars, sheet) => {
+  return [...sheet.cssRules].reduce((sheetVars, rule) => collectRuleVars(sheetVars, rule, sheet), vars);
+};
+
+export const extractPageVariables = () => {
+  const asObject = [...document.styleSheets].filter(isSameDomain).filter(isNotCoreFile).reduce(collectSheetVars, {});
+
+  return Object.keys(asObject).sort().map((k) => {
```
Inwerpsel, line 60: This currently maps it to an array, because that's how the `css-variables.json` file was structured. It's probably easier to make the code work with the object itself, since that's also how the `theme` object is handled. But didn't want to add all those changes to this PR.

lithrel (approved):
:+1:

PR #577, opened 2021-05-03
PLANET-6119 Improve CarouselHeader WYSIWYG after UAT feedback

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6119

**Frontend** (mobile/tablet)
- On scroll, it blocks the page
- I am not really able to change to the next slide
- The description and title are not visible
- The selectors indicating on which slide we are are very low on the page, there is a big space between the image and those selector

**Backend**
- Not able to see the menu b/c of the block above (only happens if previous block is the "old" carousel header, which should never happen anyway)
- Maybe the arrow from description can be moved so it does not overlap with the field
- The space where the editor needs to click to write the CTA is very small

### Testing

You can check the fixes on [this page](https://www-dev.greenpeace.org/test-rhea/carousel-header-new-wysiwyg/) for example

### DIFF
--- a/assets/src/blocks/CarouselHeader/CarouselHeaderEditor.js
+++ b/assets/src/blocks/CarouselHeader/CarouselHeaderEditor.js
@@ -71,7 +71,6 @@ export const CarouselHeaderEditor = ({ setAttributes, attributes }) => {
         {slidesWithImages?.map((slide, index) => (
           <Slide
             key={index}
-            index={index}
             ref={element => slidesRef.current[index] = element}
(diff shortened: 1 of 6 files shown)


planet-4:

:new_moon: [rhea](https://www-dev.greenpeace.org/test-rhea)

:watch: 2021.05.03 13:19:31

comzeradd (approved):
Works great! :+1:

comzeradd (approved):
Nice work!

PR #578, opened 2021-05-04
PLANET-6131 Theme editor from package

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6131

---

We discussed earlier separating the theme editor to a separate repository, since it's decoupled from Planet 4 and WordPress. It was only coupled by requiring the CSS variables to be collected into a file at webpack build time. We had to rewrite that to extract these at runtime instead, since the webpack plugin we used was causing a few issues. After this it was fairly straightforward to perform the move, so it's nice to have it out of the way. It also allows to develop the plugin faster by not requiring a code review for every small change. Instead we can review bumping the repository.

Probably it's better to include this package in master theme instead, but it seems more efficient to perform that move once we confirmed and merged these changes.

Before I did the move I merged a few pending branches into this branch. These additions are removed again by the later commits, but it would be nice to still merge these commits, so that I can link to them from the repo.

The repo is currently on my own GitHub account, but I can move it to a Greenpeace repo if we decide to work on it with the team.

TESTING:
Except for a few added features, everything should still work like before (alt + click gives you clicked element variables, saving a theme on the server, switching themes). And the resulting files (theme editor JS and CSS entrypoints) should still be the same contents and size.

The pending branches I included in the initial version add 1) a "responsive view" to theme mobile, and 2) filtering out overridden properties by checking which variable's selector is the most specific.


planet-4:

:new_moon: [proteus](https://www-dev.greenpeace.org/test-proteus)

:watch: 2021.05.05 08:10:42

lithrel (approved):
:+1:

PR #579, opened 2021-05-05
PLANET-6010 Send dataLayer event when render is finished

Inwerpsel:
* To check if this resolves/reduces GTM blocking issues.
* In order to "know" when all blocks are done, we need to render them in
a single React element using portals, instead of doing separate renders.
* Also emit event at start and finish of frontendIndex.js.

Ref: 

---


planet-4:

:new_moon: [janus](https://www-dev.greenpeace.org/test-janus)

:watch: 2021.05.05 14:31:07

PR #588, opened 2021-05-12
PLANET-6136 Make legacy sidebar code manageable

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6136

---

I first wanted to rewrite to hooks in one go, since this is the most confusing class components we have. However it's too complex to do at once.

This PR does a few low hanging fruit improvements that are easy to review without needing to go too much into the more confusing parts.

The most confusing being the 2 tightly coupled HOCs WithPostMeta and FromThemeOptions. I'm not sure yet at this point whether we want to try "save" the logic in them that handles the "theme field dependencies" (see [this added code comment](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/588/files#diff-0a557c7b06bf5a384f50fa0e07bbb3e59ba232e788d9125e74ddad2f20ce013aR7-R11)).

For now the purpose of the Jira ticket is just to have some way to select the newly created themes in the sidebar on the test instance, so that design can review the refactored version, compare with the original, and easily create small variations and switch campaigns to them.

* Move some of the logic to separate components so that it's not all inside of CampaignSidebar.
* Return early a few times.
* Some naming improvements.
* Add documentation for some obscure parts.

### DIFF
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -7,7 +7,7 @@ import { savePreviewMeta } from '../../saveMetaToPreview';
 import { PostParentLink } from './PostParentLink';
 import { LegacyThemeSettings } from './LegacyThemeSettings';
 
-const loadTheme = async( value ) => {
+const loadTheme = async (value) => {
   if ( value === '' || !value ) {
     value = 'default';
   }
@@ -44,6 +44,8 @@ export class CampaignSidebar extends Component {
     const prevTheme = this.state.theme;
     this.setState({ theme });
 
+    // Loop through the new theme's fields, and check whether any of the already chosen options has a value that is not
+    // available anymore.
     const invalidatedFields = prevTheme.fields.filter( field => {
 
       const resolvedField = resolveField(theme, field.id, meta);
@@ -58,7 +60,13 @@ export class CampaignSidebar extends Component {
 
     } ).map( field => resolveField( theme, field.id, meta ) )
 
+    // Set each of the invalidated fields to their default value, or unset them.
     return invalidatedFields.reduce( ( result, field ) => {
+      // Adding this check to prevent a crash. Probably the previous code can be rewritten to not produce null, but
+      // that would probably cascade into many changes and this is code we'll probably remove soon.
+      if (!field) {
+        return result;
+      }
       return {
         ...result,
         [ field.id ]: field.default || null,
@@ -69,7 +77,7 @@ export class CampaignSidebar extends Component {
   }
 
   componentDidMount() {
-    wp.data.subscribe( async() => {
+    wp.data.subscribe(async () => {
       const meta = wp.data.select( 'core/editor' ).getEditedPostAttribute( 'meta' );
 
       if (!meta) {
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -1,75 +1,23 @@
 import { PluginSidebar, PluginSidebarMoreMenuItem } from "@wordpress/edit-post";
 import { Component } from '@wordpress/element';
-import { PanelBody, RadioControl, SelectControl } from '@wordpress/components';
-import { compose } from '@wordpress/compose';
-import ColorPaletteControl from '../ColorPaletteControl/ColorPaletteControl';
-import { withPostMeta } from '../PostMeta/withPostMeta';
-import { withDefaultLabel } from '../withDefaultLabel/withDefaultLabel';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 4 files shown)


planet-4:

:new_moon: [pluto](https://www-dev.greenpeace.org/test-pluto)

:watch: 2021.05.12 11:06:19

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -1,75 +1,23 @@
...
-  withPostMeta,
-)( ColorPaletteControl );
-
+const loadTheme = async( value ) => {
+  if ( value === '' || !value ) {
+    value = 'default';
+  }
+  const baseUrl = window.location.href.split( '/wp-admin' )[ 0 ];
+  const themeJsonUrl = `${ baseUrl }/wp-content/themes/planet4-master-theme/campaign_themes/${ value }.json`;
+  console.log( `fetching theme ${ value }` );
```
mleray, line 16: Is this console log meant to stay here? 🙂

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -1,75 +1,23 @@
...
 
+  const json = await fetch(themeJsonUrl);
+  return await json.json();
```
mleray, line 19: Should we maybe add a check here in case `json` is undefined? 🤔

assets/src/components/Sidebar/CampaignSidebar.js
Inwerpsel: It was already there, not sure why I preserved it last time :thinking:

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -118,23 +69,26 @@ export class CampaignSidebar extends Component {
...
+        return;
+      }
+      let themeName = meta['theme'];
+      if (themeName === '') {
+        themeName = 'default';
+      }
+      if (isShallowEqual(this.state.meta, meta)) {
+        return;
+      }
+      this.setState({ meta: meta });
```
mleray, line 85: This can be shortened to `this.setState({ meta });` 🙂

assets/src/components/Sidebar/CampaignSidebar.js
Inwerpsel: It's actually does the same as before, only written using `await` instead of a promise. I don't think `json` can be undefined if `await fetch(themeJsonUrl)` fails, since then it will throw an error.

assets/src/components/Sidebar/CampaignSidebar.js
Inwerpsel: I think I kept it because I didn't trust the code would call this only when needed. Not sure if I trust it now :sweat_smile:

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -118,23 +69,26 @@ export class CampaignSidebar extends Component {
   }
 
   componentDidMount() {
-    wp.data.subscribe( () => {
+    wp.data.subscribe( async() => {
```
mleray, line 72: There's a space missing between `async` and the parenthesis 🙂

mleray (changes requested):
Tested on local; switching themes in the dropdown is not working anymore and I get an error in the console (see attached screenshot). The rest seems to be fine 👍 
Screenshot 2021-05-17 at 10 24 25

Inwerpsel:
> Tested on local; switching themes in the dropdown is not working anymore and I get an error in the console (see attached screenshot). The rest seems to be fine 
> Screenshot 2021-05-17 at 10 24 25

What are the exact steps you performed to get this error?

Inwerpsel:
> > Tested on local; switching themes in the dropdown is not working anymore and I get an error in the console (see attached screenshot). The rest seems to be fine
> > Screenshot 2021-05-17 at 10 24 25
> 
> What are the exact steps you performed to get this error? (new campaign? Which theme was it before switching?)

Turns out this was already occurring on master, but can be prevented with a `null` check. https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d1b86dbb7742174f463c19d4c2d49d93029d4c75/assets/src/components/Sidebar/CampaignSidebar.js#L65-L69

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -1,75 +1,23 @@
...
-  fromThemeOptions,
-  withPostMeta,
-)( RadioControl );
-
-const ColorPalette = compose(
-  fromThemeOptions,
-  withPostMeta,
-)( ColorPaletteControl );
-
+const loadTheme = async( value ) => {
```
mleray, line 10: Another space missing here after `async`

mleray (approved):
Looks good to me 👍

PR #589, opened 2021-05-17
PLANET-6136 Add select for new themes

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6136

---

This PR makes it possible to use a theme that was created with the experimental theme editor on a campaign. The theme name is still stored in the same meta field.

To make it possible to use the same options as one of the legacy themes, you can create a theme with the same name but `-new` appended, in which case those sidebar options are loaded. This is not intended to be used by end users, just an easy way for us to ensure the same options still work on the newer version.

As a bonus, it also includes the code needed to apply the theme's variables in the post editor. I didn't intend to do that, but it didn't require a lot of changes. It doesn't work 100% of the time, since not all front end CSS is used in the editor. But most things do and we can add the variables to the editor specific CSS later.

This PR should leave functionality unchanged on production, since there it won't have any created themes.

Review/test tips:
* We probably also want to test this on an instance that doesn't have any themes so we can verify it doesn't break that case. To do that you can change the label on this PR to another instance, and re-run the workflow to make it deploy.

### DIFF
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -23,8 +23,10 @@ const loadTheme = async (value) => {
   if ( value === '' || !value ) {
     value = 'default';
   }
+  const withoutNew = value.replace(/-new$/, '');
+  const name = isLegacy(withoutNew) ? withoutNew : 'default';
   const baseUrl = window.location.href.split( '/wp-admin' )[ 0 ];
-  const themeJsonUrl = `${ baseUrl }/wp-content/themes/planet4-master-theme/campaign_themes/${ value }.json`;
+  const themeJsonUrl = `${ baseUrl }/wp-content/themes/planet4-master-theme/campaign_themes/${ name }.json`;
 
   const json = await fetch(themeJsonUrl);
   return await json.json();
@@ -141,15 +143,13 @@ export class CampaignSidebar extends Component {
         >
           { !!parent && <PostParentLink parent={ parent }/> }
           { !parent && meta && <NewThemeSettings currentTheme={meta.theme} onChange={ async value => {
-            if (isLegacy(value)) {
-              const theme = await loadTheme();
-              this.setState({ theme });
-            }
+            const theme = await loadTheme(value);
+            this.setState({ theme });
           } }/> }
-          { !parent && theme && isLegacyTheme && <LegacyThemeSettings
-            disableStyles={!!theme}
+          { !parent && <LegacyThemeSettings
             theme={ theme }
             handleThemeSwitch={ this.handleThemeSwitch }
+            isLegacyTheme={isLegacyTheme}
           /> }
         </PluginSidebar>
       </>
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -143,8 +143,7 @@ export class CampaignSidebar extends Component {
         >
           { !!parent && <PostParentLink parent={ parent }/> }
           { !parent && meta && <NewThemeSettings currentTheme={meta.theme} onChange={ async value => {
-            const theme = await loadTheme(value);
-            this.setState({ theme });
+            this.handleThemeSwitch('theme', value, meta);
           } }/> }
           { !parent && <LegacyThemeSettings
             theme={ theme }
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -6,6 +6,18 @@ import isShallowEqual from '@wordpress/is-shallow-equal';
 import { savePreviewMeta } from '../../saveMetaToPreview';
 import { PostParentLink } from './PostParentLink';
 import { LegacyThemeSettings } from './LegacyThemeSettings';
+import { NewThemeSettings } from './NewThemeSettings';
+
+const isLegacy= theme => [
+  'default',
+  'antarctic',
+  'arctic',
+  'climate',
+  'oceans',
+  'oil',
+  'plastic',
+  'forest',
+].includes(theme) || !theme;
 
 const loadTheme = async (value) => {
   if ( value === '' || !value ) {
@@ -13,11 +25,11 @@ const loadTheme = async (value) => {
   }
   const baseUrl = window.location.href.split( '/wp-admin' )[ 0 ];
   const themeJsonUrl = `${ baseUrl }/wp-content/themes/planet4-master-theme/campaign_themes/${ value }.json`;
-  console.log( `fetching theme ${ value }` );
 
   const json = await fetch(themeJsonUrl);
   return await json.json();
 }
+
 export class CampaignSidebar extends Component {
   static getId() {
     return 'planet4-campaign-sidebar';
@@ -40,15 +52,15 @@ export class CampaignSidebar extends Component {
   // When theme switches, we need to check if any options were previously chosen that are not allowed in the new theme.
   // For each of these, we either set them to the default value
   async handleThemeSwitch( metaKey, value, meta ) {
-    const theme = await loadTheme( value )
+    const newTheme = await loadTheme( value )
     const prevTheme = this.state.theme;
-    this.setState({ theme });
+    this.setState({ theme: newTheme });
 
     // Loop through the new theme's fields, and check whether any of the already chosen options has a value that is not
     // available anymore.
     const invalidatedFields = prevTheme.fields.filter( field => {
 
-      const resolvedField = resolveField(theme, field.id, meta);
+      const resolvedField = resolveField(newTheme, field.id, meta);
 
       const currentValue = meta[ field.id ];
 
@@ -58,7 +70,7 @@ export class CampaignSidebar extends Component {
 
       return !(resolvedField.options.some( option => option.value === currentValue) );
 
-    } ).map( field => resolveField( theme, field.id, meta ) )
+    } ).map( field => resolveField( newTheme, field.id, meta ) )
 
     // Set each of the invalidated fields to their default value, or unset them.
     return invalidatedFields.reduce( ( result, field ) => {
@@ -96,7 +108,7 @@ export class CampaignSidebar extends Component {
         this.state.theme === null
       ) {
         const theme = await loadTheme(themeName);
-        this.setState({ theme });
+        this.setState({ theme: theme });
       }
     } );
     wp.data.subscribe( () => {
@@ -112,7 +124,9 @@ export class CampaignSidebar extends Component {
   }
 
   render() {
-    const { parent, theme } = this.state;
+    const { parent, theme, meta } = this.state;
+
+    const isLegacyTheme = !theme || isLegacy(theme.id);
 
     return (
       <>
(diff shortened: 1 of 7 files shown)


assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -40,15 +43,15 @@ export class CampaignSidebar extends Component {
   // When theme switches, we need to check if any options were previously chosen that are not allowed in the new theme.
   // For each of these, we either set them to the default value
   async handleThemeSwitch( metaKey, value, meta ) {
-    const theme = await loadTheme( value )
-    const prevTheme = this.state.theme;
-    this.setState({ theme });
+    const newTheme = await loadTheme( value )
```
Inwerpsel, line 46: I just renamed a few things to make it a bit more understandable. If you look at what this function does in detail, it's definitely not ideal. For example having it loop the meta instead of the theme would make more sense. But I'd rather not modify the logic unless it's really needed to avoid refactor hell :sweat_smile:

planet-4:

:new_moon: [pluto](https://www-dev.greenpeace.org/test-pluto)

:watch: 2021.05.17 15:35:45

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -6,18 +6,32 @@ import isShallowEqual from '@wordpress/is-shallow-equal';
...
+  'oil',
+  'plastic',
+  'forest',
+].includes(theme) || !theme;
 
 const loadTheme = async (value) => {
   if ( value === '' || !value ) {
     value = 'default';
   }
+  const withoutNew = value.replace(/-new$/, '');
```
Inwerpsel, line 26: Similar logic for the back end was already merged into master theme https://github.com/greenpeace/planet4-master-theme/blob/5c49b8a9bdc5d58bff429ed4c1446db4eacac964/src/PostCampaign.php#L440-L452

classes/class-loader.php
```diff
@@ -463,7 +464,9 @@ public function enqueue_campaigns_assets() {
 
 		$campaign_theme = $post->theme ?? $post->custom['_campaign_page_template'] ?? null;
 
-		if ( ! is_string( $campaign_theme ) || empty( $campaign_theme ) ) {
+		if ( ! is_string( $campaign_theme ) || empty( $campaign_theme )
+			|| ! in_array( $campaign_theme, PostCampaign::LEGACY_THEMES, true )
```
Inwerpsel, line 468: This is the part where it enqueues the files with CSS overrides. They only exist for the legacy themes.

classes/class-loader.php
```diff
@@ -463,7 +464,9 @@ public function enqueue_campaigns_assets() {
...
+		) {
 			return;
 		}
```
Inwerpsel, line 472: The `theme_antarctic` handle below is not correct, this should be generic. I'll change it if I'm sure the change can't cause issues.

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -0,0 +1,87 @@
...
+      console.log(meta);
+      onChange(selectedTheme);
+      editPost({meta});
+    }
+  };
+  useEffect(emitOnChange, [selectedTheme]);
+  const serverThemes = useServerThemes();
+  useAppliedCssVariables(serverThemes, currentTheme);
+
+  if (Object.keys(serverThemes).length === 0) {
```
Inwerpsel, line 70: This part should make it not show up on production. We didn't enable the theme editor on any site, so they won't have any.

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -0,0 +1,87 @@
...
+      const themes = await p4ServerThemes.fetchThemes();
+      setServerThemes(themes);
+    };
+    fetchThemes();
+  }, []);
+
+  return serverThemes;
+};
+
+const getAllDefinedProps = () => Object.entries(document.documentElement.style).filter(([, k]) => {
```
Inwerpsel, line 26: I'll change this to `Object.values`

Inwerpsel: `document.documentElement.style` is somewhat weird. For custom properties, it returns integer keys, and the value is the property name. It doesn't give you the custom property value. For regular properties of the object it acts normally.

I'll add an explanatory comment to the code.

package.json
```diff
@@ -42,6 +42,6 @@
     "classnames": "^2.2.6",
     "jest-environment-node": "^26.0.1",
     "photoswipe": "^4.1.3",
-    "use-theme-editor": "^1.0.3"
+    "use-theme-editor": "^1.0.9"
```
Inwerpsel, line 45: Relevant changes: https://github.com/Inwerpsel/use-theme-editor/compare/752813033934ae377a08b406598efad95ea61936..68a71cddf40af558ac3b371cdccd80a7dc7e878e

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -123,14 +139,17 @@ export class CampaignSidebar extends Component {
...
+          title={ __('Campaign Options', 'planet4-blocks-backend') }
         >
-          { parent ? <PostParentLink parent={ parent }/> :
-            <LegacyThemeSettings
-              theme={theme}
-              handleThemeSwitch={ this.handleThemeSwitch }
-            />
-          }
+          { !!parent && <PostParentLink parent={ parent }/> }
+          { !parent && meta && <NewThemeSettings currentTheme={meta.theme} onChange={ async value => {
```
Inwerpsel, line 145: Currently if you first pick a legacy theme, and then pick a new theme, it's still showing the legacy theme dropdown with the same value afterwards, even though it's not using it.

I'll see if I can improve that. For now you need to keep in mind that if the upper select is not "legacy" but one of the created themes, that one will take precedence.

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -0,0 +1,87 @@
...
+const keysAsLabel = obj => Object.keys(obj).map(k => ({ label: k, value: k }));
+
+// Just a lightweight way to have a separate UI element in the sidebar so that it's clear which themes are legacy and
+// which are new. This is likely just for internal usage to facilitate reviewing the refactored version, eventually
+// editors will work with a single select.
+const useServerThemes = () => {
+  const [serverThemes, setServerThemes] = useState({});
+
+  useEffect(() => {
+    const fetchThemes = async () => {
```
GP-Dan-Tovbein, line 16: You can implement an IIFE function, i think it's a better solutions instead of creating a new function variable
```
(async () => {
    const themes = await p4ServerThemes.fetchThemes();
     setServerThemes(themes);
})();

```
Not necessary to return a value. `useEffect` returns `void`. You will be able to take `serverThemes` value from its state

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -112,7 +126,9 @@ export class CampaignSidebar extends Component {
   }
 
   render() {
-    const { parent, theme } = this.state;
+    const { parent, theme, meta } = this.state;
+
+    const isLegacyTheme = !theme || isLegacy(theme.id);
```
GP-Dan-Tovbein, line 131: Correct me if I'm wrong but this validation is also added inside the `isLegacy` function [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/589/files#diff-030a0990cba4c830359767f7c640b7d3e58da55f7926014d6c1d4bf97c1d6d28R20)

assets/src/components/Sidebar/NewThemeSettings.js
Inwerpsel: Nice didn't think of IIFE :ok_hand: 

> Not necessary to return a value

Do you mean this line? https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/589/files#diff-a281525a512c25313074cd59d3954135907ee63aa7045f1ce29c686a9a291b2cR23

assets/src/components/Sidebar/CampaignSidebar.js
GP-Dan-Tovbein, line 145: As a best practice and thinking about the performance, it's better always to move out all`lambda` from the `render` render. Take in mind that with every render you are creating a new instance of this function.

```
const onChangeHandler = useCallback((value) => {
  this.handleThemeSwitch('theme', value, meta);
}, [
meta,
]);
```

assets/src/components/Sidebar/CampaignSidebar.js
Inwerpsel: I first also thought this after seeing your comment, but it's actually slightly different :grin: In this case `theme` is an object that is loaded from a JSON file, and not the name.

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -0,0 +1,87 @@
...
+    const customProps = getAllDefinedProps();
+
+    customProps.forEach(k => {
+      if (!Object.keys(theme).includes(k) && !initialVars.includes(k)) {
+        console.log('removing', k);
+        document.documentElement.style.removeProperty(k);
+      }
+    });
+  };
+  useEffect(applyChangesToDom, [serverThemes, currentTheme]);
```
GP-Dan-Tovbein, line 52: Not allowed this.

https://reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions

GP-Dan-Tovbein, line 52: I deleted my previous comment. I thought that it was JS function but I figured out that was a hook

assets/src/components/Sidebar/NewThemeSettings.js
GP-Dan-Tovbein: Perfect!

Yes, that line!

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -0,0 +1,87 @@
...
+  const emitOnChange = () => {
+    if (selectedTheme !== null) {
+      const meta = { theme: selectedTheme };
+      console.log(meta);
+      onChange(selectedTheme);
+      editPost({meta});
+    }
+  };
+  useEffect(emitOnChange, [selectedTheme]);
+  const serverThemes = useServerThemes();
```
GP-Dan-Tovbein, line 67: Move `const` at the top level

assets/src/components/Sidebar/CampaignSidebar.js
Inwerpsel: `useCallback` is only better for performance if the component is rendered extremely frequently (maybe in some edge cases too). If the component is not rendered frequently, its performance is actually (ever so slightly) worse. See https://kentcdodds.com/blog/usememo-and-usecallback

assets/src/components/Sidebar/NewThemeSettings.js
Inwerpsel: That is the return statement of the custom hook `useServerThemes`

assets/src/components/Sidebar/CampaignSidebar.js
GP-Dan-Tovbein: I agree with you. Feel free to avoid using `useCallback` if this component isn't extremely frequently!

GP-Dan-Tovbein:
Oh yeah! You're completely right

assets/src/components/Sidebar/NewThemeSettings.js
GP-Dan-Tovbein: Also you can do sth like this

```
if(Object.keys(serverThemes).length) {
  return (
    <div className="components-panel__body is-opened">
      <span>
        This dropdown contains themes that were created on this instance. It will be empty on production, as these won't
        have any themes created. Once we confirmed the created themes we add them to below dropdown.
      </span>
      <SelectControl
        label={ __('New Theme', 'planet4-blocks-backend') }
        title={ __('Only for reviewing themes, not intended to be 2 drop downs in the final version.', 'planet4-blocks-backend') }
        options={ [{ label: 'Legacy', value: '' }, ...keysAsLabel(serverThemes)] }
        onChange={ setSelectedTheme }
        value={ selectedTheme || '' }
        />
    </div>
  );
}

return null;
```

GP-Dan-Tovbein (approved):
Code looks good to me. 👍 
Tested on Local environment and works fine.

assets/src/components/Sidebar/NewThemeSettings.js
Inwerpsel: Both are equivalent, but putting `null` inside of an if-statement seems preferable over putting a complex component inside an if-statement.

classes/class-loader.php
Inwerpsel: On second thoughts, it's not worth the risk changing as this will be removed soon.

PR #590, opened 2021-05-20
Prevent console flooded by warnings

Inwerpsel:
Ref: no dedicated ticket

---

These 2 things are causing the same warnings each time the block is rendered, which gets in the way of debugging other console output. The `srcset` should probably be further investigated, since we always want to use it, but for some reason it's `false` sometimes.

### DIFF
--- a/assets/src/blocks/Articles/ArticlesList.js
+++ b/assets/src/blocks/Articles/ArticlesList.js
@@ -7,7 +7,7 @@ export const ArticlesList = (props) => {
     <div className="article-list-section clearfix">
       {posts && posts.length > 0 && posts.map(post =>
         <ArticlePreview
-          key={post.post_title}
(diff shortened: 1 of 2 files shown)


assets/src/blocks/Articles/ArticlesList.js
```diff
@@ -7,7 +7,7 @@ export const ArticlesList = (props) => {
     <div className="article-list-section clearfix">
       {posts && posts.length > 0 && posts.map(post =>
         <ArticlePreview
-          key={post.post_title}
+          key={post.ID}
```
Inwerpsel, line 10: Post titles are not guaranteed to be unique. While developing you often end up with an Articles block on each page with at least 2 times "Test" as the title.

mleray (approved):
Looks good to me 👍

PR #453, opened 2020-12-09
PLANET-4814: Convert EN Form block to WYSIWYG

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4814
Ref: https://jira.greenpeace.org/browse/PLANET-5767

This a beta version of the ENForm wysiwyg block. 

### Sidebar settings

Screenshot from 2020-12-09 12-16-54 Screenshot from 2020-12-09 12-17-27

### Switch between Signup form & Thank you message edition

Screenshot from 2020-12-09 12-12-20

## Test

- Enable the beta blocks in Planet4 > Features
- Add a new page
- Add a block _EN Form (beta)_

This block is enabled on test instance Venus https://www-dev.greenpeace.org/test-venus/en-form-beta-test/

## Checklist

- [x] Styles
  - [x] Full width
  - [x] Full width + background
  - [x] Side column style
- [ ] Form generator
  - [x] Text / Email
  - [x] Checkbox
  - [ ] Radio
  - [x] Country selector
  - [x] Position selector
- [x] Content
  - [x] Title & text
  - [x] Background
  - [x] Campaign logo
- [ ] Thank you
  - [x] Integrated edition
  - [ ] Share buttons
    - [x] Svg icons
    - [ ] Social accounts
- [ ] [jQuery dep](https://jira.greenpeace.org/browse/PLANET-5767)
  - [ ] Fields validation (email/url)
  - [x] Form fields interdependency
  - [x] Form submit
  - [ ] Form responsive layout (?)

PR #597, opened 2021-05-28
PLANET-4814: Move ENForm to "old"

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-4814

In preparation for https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/453 to move to beta, moving the current block to "old" name.

## Test 

ENForm block should work the same
- Check existing blocks in Block usage
- Add a block ENForm to a page, select Goal and Form, check that form displays as expected

### DIFF
--- a/classes/blocks/class-oldenform.php
+++ b/classes/blocks/class-oldenform.php
@@ -12,11 +12,11 @@ use P4GBKS\Controllers\Ensapi_Controller as Ensapi;
 use WP_Block_Type_Registry;
 
 /**
- * Class ENForm
+ * Class OldENForm
  *
  * @package P4GBKS\Blocks
  */
-class ENForm extends Base_Block {
+class OldENForm extends Base_Block {
 
 	/** @const string BLOCK_NAME */
 	protected const BLOCK_NAME = 'enform';
@@ -34,7 +34,7 @@ class ENForm extends Base_Block {
 	private const FIELDS_META = 'p4enform_fields';
 
 	/**
-	 * ENForm constructor.
+	 * OldENForm constructor.
 	 */
 	public function __construct() {
 		if ( WP_Block_Type_Registry::get_instance()->is_registered( self::get_full_block_name() ) ) {
--- a/assets/src/blocks/OldENForm/ENForm.js
+++ b/assets/src/blocks/OldENForm/ENForm.js
@@ -1,0 +1,369 @@
+import {Component} from '@wordpress/element';
+import {Preview} from '../../components/enform/Preview';
+import {LayoutSelector} from '../../components/enform/LayoutSelector/LayoutSelector';
+import {FormSectionTitle} from '../../components/enform/FormSectionTitle/FormSectionTitle';
+import {FormHelp} from '../../components/enform/FormHelp/FormHelp';
+import {InlineFormFeedback} from '../../components/enform/InlineFormFeedback/InlineFormFeedback';
+import {
+  TextControl,
+  TextareaControl,
+  ToggleControl,
+  SelectControl,
+  ServerSideRender
+} from '@wordpress/components';
+import {MediaPlaceholder} from "@wordpress/editor";
+import {ValidationMessage} from "../../components/enform/ValidationMessage/ValidationMessage";
+
+const {__} = wp.i18n;
+
+export class ENForm extends Component {
+  constructor(props) {
+    super(props);
+  }
+
+  isURLValid( url ) {
+    return !!url ? url.substr(0, 8).toLowerCase() === 'https://' : true;
+  }
+
+  renderEdit() {
+    const {getCurrentPostType} = wp.data.select("core/editor");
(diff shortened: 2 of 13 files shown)


planet-4:

:new_moon: [deimos](https://www-dev.greenpeace.org/test-deimos)

:watch: 2021.05.28 13:59:54

sagarsdeshmukh:
The PR tested on local, the functionality wise looks all good to me 👍

sagarsdeshmukh:
> We need to ensure the new code still preserves the history of the old version, and mainly that merging won't try to apply the new code changes to the `Old*` files. In order to do so, I think we have no option but to split the file in 2 in this PR.
> 
> For the Covers block we also did that, but it's harder to see since it was included in the main PR (not sure why, I think practically it was easier that way).
> 
> For CarouselHeader we did do that separately: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/568/files
> 
> We have at least 2 options:
> 
> 1. the easy way: amend the commit of the renames you did(remove previous name + add new name), so that it doesn't remove the previous name. Then Git will stop seeing it as a rename. The downside is that the `Old*` files won't "have"* history.
> 2. the somewhat harder way: Split the file while preserving history for both. This is a multi step process that results in 2 files with the right attributions for each line (pointing to the same history). We didn't try it for the other WYSIWYG blocks PRs, so I'm not sure how much more effort it is. Probably it's not worth it because we don't intend to make changes to the old code and it will be removed once WYSIWYG is ready. (in case you're interested, [this article describes the steps](https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892))

I agree with @Inwerpsel on keeping the git history with new EN form functionality. 
In my opinion and also said by Pieter, we just need to revert the git rename and copy the old en form related files(for git they are new files, when we merge the final version of wysiwyg EN form we just need to remove copied files, and git history will stay in wysiwyg EN form block version)

lithrel:
@Inwerpsel @sagarsdeshmukh  Changed commit to a copy :ok_hand:

sagarsdeshmukh (approved):
PR looks good to me 👍

We need fix the duplicate code issue from sonarcube, I tried a quick exp. [here](https://github.com/greenpeace/planet4-master-theme/pull/1397) but it wasn’t work.

Inwerpsel:
> PR looks good to me 
> 
> We need to fix the duplicate code issue from sonarcube, I tried a quick exp. [here](https://github.com/greenpeace/planet4-master-theme/pull/1397) but it wasn’t work.

Nice one @sagarsdeshmukh , I wasn't aware of this config file. For the previous WYSIWYG PR's I [configured it through the SonarCloud UI](https://sonarcloud.io/project/settings?category=exclusions&id=greenpeace_planet4-plugin-gutenberg-blocks), which results in the same config as the file you added (it also lists the same key).

Do you have permissions to see that config page on SonarCloud? Maybe the config file is not allowed/used for excluding things that are introduced in the same PR?

sagarsdeshmukh:
> Do you have permissions to see that config page on SonarCloud? Maybe the config file is not allowed/used for excluding things that are introduced in the same PR?

Nope, I didn’t found the config page. I signup with github login.

classes/blocks/class-oldenform.php
```diff
@@ -0,0 +1,378 @@
...
+		$social = [
+			'title'       => esc_attr( $og_title ),
+			'description' => esc_attr( wp_strip_all_tags( $og_description ) ),
+			'link'        => esc_url( $link ),
+		];
+
+		$data = [];
+
+		if ( isset( $attributes['thankyou_url'] ) && $attributes['thankyou_url'] && 0 !== strpos( $attributes['thankyou_url'], 'http' ) ) {
+			$attributes['thankyou_url'] = 'http://' . $attributes['thankyou_url'];
```
Inwerpsel, line 236: SonarCloud [signals this as a security issue](https://sonarcloud.io/project/security_hotspots?id=greenpeace_planet4-plugin-gutenberg-blocks&pullRequest=597&resolved=false&types=SECURITY_HOTSPOT). Now this isn't really changed by this PR, so we could ignore it. But unless there's a specific good reason for using `http`, we probably should redirect to `https` here.

Inwerpsel: Perhaps @comzeradd knows whether `http` is intentional here?

Inwerpsel: Since SonarCloud allows to, I marked the security issue as "safe", and created a [dedicated ticket](https://jira.greenpeace.org/browse/PLANET-6193) to look into it.

Inwerpsel (approved):

classes/blocks/class-oldenform.php
comzeradd: I don't think there was specific reason for that. We could just change that to `https`, since we only add that when people forget to add a protocol to the url.

PR #599, opened 2021-06-01
PLANET-5247 Debug composer issues in "A11y: Address accessibility of spreadsheet block"

Inwerpsel:
Testing if rebase fixes both install issues.

To make it use the same test instance as the PR we're debugging, I used the same Jira ticket prefix in the title (our CI will look for the test instance name in the Jira ticket data).


Inwerpsel:
Composer issues were fixed in the meanwhile :partying_face:

PR #600, opened 2021-06-02
PLANET-6155 Convert cookies block to hooks part I

Inwerpsel:

---

As a preparation step for adding HubSpot support to the cookies block, this converts the block from class components to function components and hooks.

The block currently uses a less known feature of classes (adding a callback function to `this.setState`) to set certain cookies as an "effect" when the state changes. This effect will need to be expanded to also trigger HubSpot cookie consent, which is much easier to achieve with `useEffect`.

I probably will also split the hooks conversion in 2 PRs. Currently on this branch I have a few commits that simplify the class which could already be merged, after which the rest should be easier to review.

### DIFF
--- a/assets/src/blocks/Cookies/CookiesFrontend.js
+++ b/assets/src/blocks/Cookies/CookiesFrontend.js
@@ -60,23 +60,24 @@ export class CookiesFrontend extends Component {
   }
 
   onNecessaryCookiesClick() {
-    const isChecked = !this.state.necessaryCookiesChecked;
-    let { allCookiesChecked } = this.state;
+    const {allCookiesChecked, necessaryCookiesChecked} = this.state;
 
-    if (isChecked) {
+    // Flip previous value.
+    const allowNecessary = !necessaryCookiesChecked;
+
+    if (allowNecessary) {
       createCookie('greenpeace', '1', 365);
       hideCookieNotice();
     } else {
-      allCookiesChecked = false;
       createCookie('greenpeace', '0', -1);
       showCookieNotice();
     }
 
     this.setState({
-      necessaryCookiesChecked: isChecked,
+      necessaryCookiesChecked: allowNecessary,
       // if Necessary Cookies is not checked,
       // All Cookies should be unchecked too
-      allCookiesChecked,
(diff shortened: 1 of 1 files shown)


assets/src/blocks/Cookies/CookiesFrontend.js
```diff
@@ -3,10 +3,51 @@ import { FrontendRichText } from '../../components/FrontendRichText/FrontendRich
 
 const { __ } = wp.i18n;
 
+const readCookie = (name) => {
+  const nameEQ = name + '=';
+  const ca = document.cookie.split(';');
+  let c;
+  for (let i = 0; i < ca.length; i++) {
+    c = ca[i];
+    while (c.charAt(0) === ' ') {
```
mleray, line 12: Couldn't we use `c.trim()` here? 🤔

mleray, line 12: We could use `trimStart` here 🙂

Inwerpsel: Not sure if it's needed, but we can safely use `trim` here. I also simplified the rest of the function.

mleray (changes requested):
Just tested this on local, the behaviour is now different:
- When checking `all cookies`, `necessary cookies` no longer gets checked at the same time
- When checking `necessary cookies` and then `all cookies`, `necessary cookies` gets unchecked
- The cookies banner is no longer always shown when `necessary cookies` is unchecked, which I think is wrong?

Inwerpsel:
> Just tested this on local, the behaviour is now different:
> 
> * When checking `all cookies`, `necessary cookies` no longer gets checked at the same time
> * When checking `necessary cookies` and then `all cookies`, `necessary cookies` gets unchecked
> * The cookies banner is no longer always shown when `necessary cookies` is unchecked, which I think is wrong?

Did you pull again before testing? The last commit I force pushed should have fixed these

mleray (approved):
Re-tested on local, looks good now 👍

planet-4:

:watch: 2021.06.02 12:29:22

PR #601, opened 2021-06-02
PLANET-6155 Convert cookies block to hooks part II

Inwerpsel:

---

**Easier to review [with hidden whitespace changes](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/601/files?diff=split&w=1)**

The actual conversion. I added a `useCookie` hook to encapsulate reading and writing the cookie. `no_track` cookie doesn't use it because we only ever need to write it.

Also renamed `setCookie` to `writeCookie`, because of a naming collision. Though after doing so the colliding other function was moved, so it could theoretically stay `setCookie`. But we can keep `writeCookie` to avoid confusion with setter methods returned by `useState`.

### DIFF
--- a/assets/src/blocks/Cookies/useCookie.js
+++ b/assets/src/blocks/Cookies/useCookie.js
@@ -1,0 +1,47 @@
+import { useState, useEffect } from 'react';
+
+export const readCookie = (name) => {
+  const declarations = document.cookie.split(';');
+  let match = null;
+  declarations.forEach(part => {
+    const [key, value] = part.split('=');
+    if (key.trim() === name) {
+      match = value;
+    }
+  });
+  return match;
+};
+
+export const writeCookie = (name, value, days = 365) => {
+  let date = new Date();
+  date.setTime(date.getTime() + (days * 24 * 60 * 60 * 1000));
+  let secureMode = document.location.protocol === 'http:'
+    ? ';SameSite=Lax'
+    : ';SameSite=None;Secure';
+  document.cookie = encodeURI(name) + '=' + encodeURI(value) + ';domain=.' + document.domain + ';path=/;' + '; expires=' + date.toGMTString() + secureMode;
+};
+
+// Value should not matter as cookie is expired.
+export const removeCookie = name => writeCookie(name, '0', -1);
+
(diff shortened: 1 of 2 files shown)


assets/src/blocks/Cookies/useCookie.js
```diff
@@ -0,0 +1,47 @@
...
+  declarations.forEach(part => {
+    const [key, value] = part.split('=');
+    if (key.trim() === name) {
+      match = value;
+    }
+  });
+  return match;
+};
+
+export const writeCookie = (name, value, days = 365) => {
```
Inwerpsel, line 15: I made days optional because currently all calls to `writeCookie` have the same days, except for when the cookie is removed, which has a wrapper function around it.

assets/src/blocks/Cookies/useCookie.js
```diff
@@ -0,0 +1,47 @@
...
+  const saveCookie = () => {
+    if (value === null) {
+      removeCookie(name);
+      return;
+    }
+    writeCookie(name, value);
+  };
+  useEffect(saveCookie, [value]);
+
+  return [
```
mleray, line 39: I think it would be clearer to return an object here instead of an array? 🤔 Then we can label what's returned, and also we don't need to worry about which order we write them in when using the hook.

Inwerpsel: There's a good reason for using arrays, which is that most of the time the name of object properties will be too generic. Most usages of the hook will need to give it a more specific name. Imagine having to do 
```
const { value: myFooValue, setValue: setMyFooValue) = useState('defaultFoo')
```
every time you want to use state. Many custom hooks will suffer from this issue too, and in this case it also needs to be called specifically.

That said I do agree that for more than 2 things being returned it maybe becomes too error prone.

mleray: Fair point, I guess it's also fine here to use an array since this hook is not meant to be reused 👍

mleray (approved):

planet-4:

:watch: 2021.06.02 15:31:09

PR #602, opened 2021-06-02
PLANET-6155 Wire up HubSpot consent with our cookie consent mechanism

Inwerpsel:

---

We want to ensure HubSpot consent matches our cookie settings. How exactly that will work I still need to figure out. It's not something that's directly supported by the plugin.

My current guess it we'll have to add a filter to block loading the JS unless the cookie is already present in the back end. Then on the front end we need to load the script when consent is given.

Once the HubSpot code is loaded, we can trigger a consent revoke by pushing a `revokeConsent` event. Still need to find out if this is sufficient or if the current page could continue tracking. The next page load should then not load HubSpot code again.

### DIFF
--- a/assets/src/blocks/Cookies/CookiesFrontend.js
+++ b/assets/src/blocks/Cookies/CookiesFrontend.js
@@ -38,7 +38,8 @@ export const CookiesFrontend = props => {
   } = props;
 
   // Whether consent was revoked by the user since current page load.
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2021.06.03 07:58:33

mleray (approved):
Looks good 👍

PR #603, opened 2021-06-03
PLANET-5662 Use Preact instead of core WP React

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-5662

### Testing
- Run `npm install`
- Run `npm run build:preact`

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,6 +4,7 @@
     ],
     "plugins": [
         "@babel/plugin-proposal-nullish-coalescing-operator",
-        "@babel/plugin-proposal-optional-chaining"
+        "@babel/plugin-proposal-optional-chaining",
+		["@babel/plugin-proposal-decorators", { "legacy": true }]
     ]
 }
--- a/assets/src/blocks/CarouselHeader/SlidesContainer.js
+++ b/assets/src/blocks/CarouselHeader/SlidesContainer.js
@@ -1,10 +1,7 @@
 import { ArrowsAndIndicators } from './ArrowsAndIndicators';
 
-import {
-  Children,
-  cloneElement,
-  forwardRef,
-} from '@wordpress/element';
+import { cloneElement } from '@render';
+import { forwardRef } from '@compat';
 
 const StaticSlidesContainer = ({
   children,
@@ -14,9 +11,11 @@ const StaticSlidesContainer = ({
   goToNextSlide = null,
   slides = [],
 }, ref) => {
-  const renderChildSlides = passedProps => Children.map(children, child => {
+  const renderChildSlides = passedProps => {
+    return children.map(child => {
       return child && cloneElement(child, passedProps)
-    });
+    })
+  };
 
   return (
     <section
--- a/package.json
+++ b/package.json
@@ -9,12 +9,14 @@
   "scripts": {
     "start": "wp-scripts start",
     "build": "wp-scripts build --mode production",
+    "build:preact": "wp-scripts build --mode production --config webpack.preact.config.js --webpack-no-externals",
     "lint:css": "wp-scripts lint-style",
     "setupgithooks": "bin/copy-git-hooks.sh",
     "test": "env JEST_TIMEOUT=15000 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e",
     "test:watch": "env JEST_TIMEOUT=30000 PUPPETEER_HEADLESS=false PUPPETEER_SLOWMO=100 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-decorators": "^7.14.2",
     "@babel/plugin-proposal-nullish-coalescing-operator": "*",
     "@babel/plugin-proposal-optional-chaining": "*",
     "@commitlint/cli": "^12.1.4",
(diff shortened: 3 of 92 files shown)


assets/src/blocks/CarouselHeader/SlidesContainer.js
```diff
@@ -1,10 +1,7 @@
 import { ArrowsAndIndicators } from './ArrowsAndIndicators';
 
-import {
-  Children,
```
Inwerpsel, line 4: Not sure why this was used and not just a regular map. Perhaps it's somehow needed in conjunction with [`cloneElement` that's used below](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/603/files#diff-24823ce53c591e97c607cbf7001b17b9977148cfac77f9b53be687ac20c6035bL18)?

assets/src/blocks/CarouselHeader/SlidesContainer.js
```diff
@@ -14,9 +11,11 @@ const StaticSlidesContainer = ({
   goToNextSlide = null,
   slides = [],
 }, ref) => {
-  const renderChildSlides = passedProps => Children.map(children, child => {
+  const renderChildSlides = passedProps => {
+    return children.map(child => {
```
Inwerpsel, line 15: We can use the short form like it was before to save us a few braces.

package.json
```diff
@@ -9,12 +9,14 @@
...
     "start": "wp-scripts start",
     "build": "wp-scripts build --mode production",
+    "build:preact": "wp-scripts build --mode production --config webpack.preact.config.js --webpack-no-externals",
     "lint:css": "wp-scripts lint-style",
     "setupgithooks": "bin/copy-git-hooks.sh",
     "test": "env JEST_TIMEOUT=15000 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e",
     "test:watch": "env JEST_TIMEOUT=30000 PUPPETEER_HEADLESS=false PUPPETEER_SLOWMO=100 WP_USERNAME=admin WP_PASSWORD=admin WP_BASE_URL=https://www.planet4.test wp-scripts test-e2e"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-decorators": "^7.14.2",
```
Inwerpsel, line 19: Could this be a leftover from the first PR? I didn't see anything added that uses this.

public/js/enform_submit.js
```diff
@@ -53,7 +53,7 @@ const p4_enform_frontend = (function ($) {
 
   enform.validateEmail = function(email) {
     // Reference: https://stackoverflow.com/questions/46155/how-to-validate-an-email-address-in-javascript
-    const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+    const re = /^(([^<Fragment>()[\]\\.,;:\s@"]+(\.[^<Fragment>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
Inwerpsel, line 56: Probably also because of a general find and replace?

webpack.preact.config.js
```diff
@@ -0,0 +1,154 @@
+const defaultConfig = require("./node_modules/@wordpress/scripts/config/webpack.config");    // Require default Webpack config
```
Inwerpsel, line 1: Duplicating the entire webpack config might not be the best way to handle this :thinking: I guess each of the things webpack is building should use either WordPress's React or Preact, depending on whether the script will be served on the front end or the back end.

.babelrc
```diff
@@ -4,6 +4,7 @@
     ],
     "plugins": [
         "@babel/plugin-proposal-nullish-coalescing-operator",
-        "@babel/plugin-proposal-optional-chaining"
+        "@babel/plugin-proposal-optional-chaining",
+		["@babel/plugin-proposal-decorators", { "legacy": true }]
```
Inwerpsel, line 8: I think this is not used, maybe a leftover from the earlier PR?

Inwerpsel (changes requested):
Mostly looks good except for the webpack config. If I read it correctly, currently it would not use the Preact build, since it's a separate `webpack.preact.config` file that's not called. But I don't think we need to duplicate the config, since each entrypoint needs to be built with only one of the two. So I guess we can handle this in our existing config file.

Also some occurrences of `<>` that are not fragments were replaced too which would break a few things.

webpack.preact.config.js
Inwerpsel: Definitely possible, but hard to find relevant documentation or examples. Closest I could find is [this plugin that allows to map depending on file name](https://github.com/heiseshandian/dynamic-alias-resolve-plugin#readme). Though it seems to expect the function to return a absolute path, while we want to use another alias.

public/js/enform_submit.js
mleray: Indeed! 🤦 Solved

assets/src/blocks/CarouselHeader/SlidesContainer.js
mleray: Yeah fair point, I've reverted it for now just to be sure 🙂

.babelrc
mleray: This was actually not in the initial PR, but now of course I can't remember why I added it 😅 I'll double-check!

planet-4:

:new_moon: [saturn](https://www-dev.greenpeace.org/test-saturn)

:watch: 2021.07.08 13:50:55

PR #604, opened 2021-06-03
Inject version key in composer file on zip creation

comzeradd:
This is just the counterpart of greenpeace/planet4-master-theme#1392

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -337,6 +337,12 @@ jobs:
           name: Remove unnecessary files
           command: rm -rf .circleci .git .githooks assets/src bin tests
       - run:
+          name: Inject version to composer
+          command: |
+            apk add jq
+            cat composer.json | jq --arg version ${CIRCLE_TAG} '. + {version: $version}' > tmp.json
+            mv -f tmp.json composer.json
+      - run:
           name: Create zip file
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -325,19 +325,28 @@ jobs:
...
+          name: Remove unnecessary files
           command: rm -rf .circleci .git .githooks assets/src bin tests
-      # Exclude node_modules instead of removing, which takes a long time (lots of small files).
       - run:
-          name: "Create zip file"
+          name: Inject version to composer
+          command: |
+            apk add jq
+            cat composer.json | jq --arg version ${CIRCLE_TAG} '. + {version: $version}' > tmp.json
+            cp -f tmp.json composer.json
```
lithrel, line 344: I just realized tmp.json ends up in the final zip

lithrel (approved):
:+1:

.circleci/config.yml
Inwerpsel: If it gets immediately overwritten in the next statement, we could also do 
```bash
cat composer.json | jq --arg version ${CIRCLE_TAG} '. + {version: $version}' > composer.json

Inwerpsel: Using `mv tmp.json composer.json` would fix that

PR #606, opened 2021-06-04
PLANET-5868: Make table caption styles same as image caption

kritisingh1:
Ref: 

Fixes https://github.com/greenpeace/planet4/issues/136

The change makes the table caption style same as the image caption style. Please review

### DIFF
--- a/assets/src/styles/blocks/core-overrides/Table.scss
+++ b/assets/src/styles/blocks/core-overrides/Table.scss
@@ -46,6 +46,15 @@
     }
   }
 
+  // Table caption
+  figcaption {
+    text-align: center;
+    font-size: $font-size-xxs;
+    font-family: $roboto;
+    color: $grey-60;
+    line-height: 1.4;
+  }
+
   &.is-style-stripes table {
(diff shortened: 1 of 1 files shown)


kritisingh1:
@mleray I was not sure about the attributes 'line-height' and 'margin-bottom' hence haven't included them. Let me know if I should add them as well

mleray (changes requested):
Hi @kritisingh1, thanks for looking into this 💚 I've tested it on local and I have a couple of remarks:
- the selector should be `figcaption` instead of `caption` 
- it should be nested under `.wp-block-table` directly and not within the `table` tag
- the `display: table-caption;` should be removed because it breaks the layout

And for you questions, I think that the line-height is necessary (in case the caption is several lines long), but indeed the margin-bottom seems to already be 0 by default so it can be omitted 🙂 Let me know if you have any more questions!

kritisingh1:
@mleray have added a commit incorporating the requested changes

mleray (approved):
Tested on local, looks good 👍 Thank you!

PR #608, opened 2021-06-07
PLANET-5757 Add image title attribute

sagarsdeshmukh:
Previously, the global.js adds the image title attribute with alt text. To remove jQuery from global.js, drop the gobal.js and add img title at source.

Ref: [JIRA 5757](https://jira.greenpeace.org/browse/PLANET-5757)

---

Related PR: https://github.com/greenpeace/planet4-master-theme/pull/1395

### DIFF
--- a/assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
+++ b/assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
@@ -35,7 +35,8 @@ export const SplittwocolumnsFrontend = ({
           <div className="split-two-column-item-image">
             <img src={issue_image_src}
                  srcSet={issue_image_srcset}
-                 alt={issue_image_title || ''}
+                 alt={issue_image_title}
+                 title={issue_image_title}
                  style={{objectPosition: focus_issue_image}}
                  sizes={IMAGE_SIZES.columnLeft}
             />
@@ -71,7 +72,8 @@ export const SplittwocolumnsFrontend = ({
           <div className="split-two-column-item-image">
             <img src={tag_image_src}
                  srcSet={tag_image_srcset}
-                 alt={tag_image_title || ''}
+                 alt={tag_image_title}
+                 title={tag_image_title}
                  style={{objectPosition: focus_tag_image}}
                  sizes={IMAGE_SIZES.columnRight}
             />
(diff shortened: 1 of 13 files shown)


assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
```diff
@@ -36,6 +36,7 @@ export const SplittwocolumnsFrontend = ({
             <img src={issue_image_src}
                  srcSet={issue_image_srcset}
                  alt={issue_image_title || ''}
+                 title={issue_image_title || ''}
```
mleray, line 39: I don't think that we need `|| ''` here, considering that the backend already has [this check](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-splittwocolumns.php#L131)? 🤔 However in some other blocks it's not there (for instance Gallery), so maybe we need to add this either in PHP or JS to avoid any undefined values

mleray (commented):
~~Shouldn't this PR actually also remove `global.js`? 😄~~

mleray (commented):
Also, I see that you didn't make this change in certain twig templates (for instance the teasers or the author ones), is that on purpose?

sagarsdeshmukh:
> global.js
Yes it is, but that belongs to [master theme](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/js/global.js) 😃

mleray:
Aaaah yes my bad, I searched in this repo but of course forgot that master-theme is a submodule there now 🤦 Sorry, ignore my global comments here then 😬

assets/src/blocks/Splittwocolumns/SplittwocolumnsFrontend.js
sagarsdeshmukh: Yes make sense, I will remove it from JS part, as its already checked in backend. 👍

sagarsdeshmukh:
> Aaaah yes my bad, I searched in this repo but of course forgot that master-theme is a submodule there now 🤦 Sorry, ignore my global comments here then 😬

No worries, I also got confuse for few seconds while making this change 😄

mleray (approved):
Looks good to me 👍

planet-4:

:new_moon: [venus](https://www-dev.greenpeace.org/test-venus)

:watch: 2021.06.14 12:03:27

comzeradd (approved):

PR #615, opened 2021-06-16
PLANET-6151 Simplify CarouselHeader WYSIWYG code + add RTL

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6151

- Simplify the CarouselHeader WYSIWYG code by removing some components, which avoids having to pass down the same props multiple times
- Add RTL implementation both in the frontend (also swipe on mobile) and backend
- Make some small CSS fixes

### Testing

You can add a "CarouselHeader (beta)" block to the MENA [dev site](https://www-dev.greenpeace.org/mena/ar/) in Arabic and compare it to the "normal" block in the corresponding [production site](https://www.greenpeace.org/mena/ar/) to test the changes. The behaviour should be the same in all screen sizes 🙂


assets/src/blocks/CarouselHeader/CarouselHeaderEditor.js
```diff
@@ -16,7 +16,7 @@ export const CarouselHeaderEditor = ({ setAttributes, attributes }) => {
   const slidesRef = useRef([]);
   const slidesWithImages = useCarouselHeaderImages(slides);
 
-  const { currentSlide, goToSlide, goToNextSlide, goToPrevSlide } = useSlides(slidesRef, slides.length);
+  const { currentSlide, goToSlide, goToNextSlide, goToPrevSlide } = useSlides(slidesRef, slides.length - 1);
```
mleray, line 19: This was actually broken, meant that in the backend we couldn't go from last slide back to the first one using the arrows 😬

planet-4:

:new_moon: [phobos](https://www-dev.greenpeace.org/test-phobos)

:watch: 2021.06.16 14:57:45

Inwerpsel (commented):
Still need to check the CSS, but so far it's looking good. Posting these already

assets/src/blocks/CarouselHeader/ArrowsAndIndicators.js
```diff
@@ -1,24 +1,41 @@
...
-        slides={slides}
-        currentSlide={currentSlide}
-        goToSlide={goToSlide}
-      />
-    </>
-  );
-};
+}) => slides.length > 1 && (
+  <>
+    {/* Arrows */}
```
Inwerpsel, line 11: Not sure if these comments are even needed, as the markup is self-explanatory.

assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss
```diff
@@ -549,12 +559,16 @@ $medium-image-height: 600px;
         display: block;
         margin: auto;
       }
+
+      html[dir="rtl"] & {
+        transform: rotate(180deg);
```
Inwerpsel, line 564: Should work, but we should be mindful of possible adverse affects of doing a CSS transform. Doing a rotate would not only swap left and right, but also up and down (not sure if that results in any visual or UX difference, but it might not play well with paddings). Maybe mirroring it using `transform: scaleX(-1)` would be a bit safer?

Inwerpsel (approved):
Looks good :+1: We should be a bit prudent with the CSS transform, but for now I don't see a problem with it. If you have an alternative that would be nice, otherwise we can keep an eye on it and perhaps do a follow up.

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
```diff
@@ -1,8 +1,11 @@
 import { useRef, useEffect } from '@wordpress/element';
 import { useSlides } from './useSlides';
-import { CarouselHeaderStaticContent } from './CarouselHeaderStaticContent';
+import { Slide } from './Slide';
+import { CarouselControls } from './CarouselControls';
+import { SlideBackground } from './SlideBackground';
+import { StaticCaption } from './StaticCaption';
 
-const { __ } = wp.i18n;
```
mleray, line 5: Wasn't used

assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss
mleray: Updated! Also added a small fix for the transitions on hover 👍

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
```diff
@@ -35,9 +38,9 @@ export const CarouselHeaderFrontend = ({ slides, carousel_autoplay }) => {
       goToPrevSlide
     ];
 
-    // if (isRTL) {
-    //   swipeListeners.reverse();
-    // }
+    if (isRTL) {
+      swipeListeners.reverse();
```
Inwerpsel, line 42: For readability it's significantly more straightforward if we don't create the array and instead check with a ternary in place.
```js
hammer.on('swipeleft', isRTL ? goToPrevSlide : goToNextSlide);
hammer.on('swiperight', isRTL ? goToNextSlide : goToPrevSlide);

File: assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss [hunk 6]
```diff
@@ -535,6 +540,15 @@ $medium-image-height: 600px;
         background-color: rgba(0, 0, 0, .15);
         transform: translateX(-5px);
       }
+
+      html[dir="rtl"] & {
+        right: 0;
+        left: auto;
+
+        &:hover {
+          transform: translateX(5px);
```
Inwerpsel, line 549: Does it have to be translate? Perhaps adjusting margin/padding would work too?

Inwerpsel: And it actually also makes sense here :ok_hand:

mleray: Updated to use padding instead 🙂

Inwerpsel (approved):

assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss
Inwerpsel: Just to document it here as well: we noticed that using `translateX` results in the button moving from under the cursor, if the cursor is within 5px of the edge. Because there is also a transition speed this results in the button moving back and forth a few times (exact behavior probably varies across browsers/screen sizes). Though quickly and not in all cases, so a very minor issue.

Using padding both fixes that and avoids needing to change the value for RTL. With `translateX` we would need to override this with the opposite value for RTL, for padding it works in both cases.

PR #616, opened 2021-06-17
PLANET-6147: Fix missing and breaking classes on blocks

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6147

The style name extraction method generates issues with the "additional classes" field.
Additional classes are also not included in the resulting blocks, except when the current bug extracts a style and if those additional classes are placed after the style class name.

Currently, for a Columns block with style `tasks`, an additional class `is-style-tasks` is included. If we edit the additional class to `foo is-style-tasks bar`, the column will end up with classes `block-style-tasks bar`. This is because the code splits the classname string on `is-style-`, keeps everything that is after it (so `tasks bar`) and concatenate to `block-style-`. 
Other blocks simply don't have these additional CSS classes included on their container section.

## Example

You should look at the `class` attribute of the `section` tag (usually).
**Config**:
`foobar is-style-no_image foobaz`
Screenshot from 2021-06-17 18-22-17
**Before fix**:
`block columns-block block-style-no_image foobaz`
Screenshot from 2021-06-17 18-24-25
**After fix**:
`block columns-block block-style-no_image foobar is-style-no_image foobaz`
Screenshot from 2021-06-17 18-28-30

## Fix 

- Adding a function `getStyleFromClassName` to use in relevant blocks
- Adding className string to generated blocks containers that are missing it

This PR does not fix
- ServerSideRendered blocks : OldENForm, OldCarouselHeader, SocialMedia, TakeActionBoxout
- SplitTwoColumns (not sure I understand where `className` is) 

## Test

Here is the boring part:
- Add one of every Planet4 block in a page
- Add additional classes in the "Additional CSS class(es)" field everywhere you can
- Check that it appears in the source inspector on backend and frontend 

The less boring way:
I've edited a page on Sinope (https://www-dev.greenpeace.org/test-sinope/?page_id=47006&preview=true), adding classes `foobar` to blocks, and `foobaz` when a class `is-style-*` is involved (to check that classes included before and after this string are properly added). So you can also:
- Go there and open the inspector
- Search (ctrl+f) for classes `.foobar`, `.foobaz`
- Check that they are in there, and that classes `is-style-*` are also included

### DIFF
--- a/assets/src/blocks/Accordion/AccordionEditor.js
+++ b/assets/src/blocks/Accordion/AccordionEditor.js
@@ -36,7 +36,7 @@ const renderView = ({ title, description, tabs, className }, setAttributes, isSe
   };
 
   return (
-    <div className={`block accordion-block ${className}`}>
+    <div className={`block accordion-block ${className ?? ''}`}>
       <header>
         <RichText
           tagName="h2"
--- a/assets/src/blocks/Accordion/AccordionFrontend.js
+++ b/assets/src/blocks/Accordion/AccordionFrontend.js
@@ -1,7 +1,7 @@
 window.dataLayer = window.dataLayer || [];
 
 export const AccordionFrontend = ({ title, description, tabs, className }) => (
-  <section className={`block accordion-block ${className}`}>
+  <section className={`block accordion-block ${className ?? ''}`}>
     {title &&
       <header>
         <h2 className="page-section-header">{title}</h2>
--- a/assets/src/blocks/Articles/ArticlesFrontend.js
+++ b/assets/src/blocks/Articles/ArticlesFrontend.js
@@ -10,6 +10,7 @@ export const ArticlesFrontend = (props) => {
     read_more_text,
     read_more_link,
     button_link_new_tab,
+    className,
   } = props;
 
   const postType = document.body.getAttribute('data-post-type');
@@ -27,7 +28,7 @@ export const ArticlesFrontend = (props) => {
   }
 
   return (
-    <section className="block articles-block">
+    <section className={`block articles-block ${className ?? ''}`}>
       <div className="container">
         <header>
           <h2 className="page-section-header">{ article_heading }</h2>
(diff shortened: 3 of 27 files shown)


planet-4:

:watch: 2021.06.17 13:37:41

mleray (commented):
For the CarouselHeader, you could probably also solve this by adding the className to the WYSIWYG [version](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js#L76)?

mleray (commented):
For the SplitTwoColumns block, I'm pretty sure that the issue is due to the filtering that happens [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-splittwocolumns.php#L162), since `className` isn't set in the [ATTRIBUTES](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-splittwocolumns.php#L36) it's removed and therefore not passed to the frontend 🤔

lithrel:
> For the CarouselHeader, you could probably also solve this by adding the className to the WYSIWYG [version](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js#L76)?

@mleray Good point, missed it because of missing rebase, I just added it :+1: 

> For the SplitTwoColumns block, I'm pretty sure that the issue is due to the filtering that happens [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-splittwocolumns.php#L162), since `className` isn't set in the [ATTRIBUTES](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/blocks/class-splittwocolumns.php#L36) it's removed and therefore not passed to the frontend thinking

:eyes: great find :+1:  added a short fix for it

Also added a [short fix](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/616/files#diff-7b24535513d85372ec7dcdd75454546d4ca1a95d7ae7120f6fbd62a19ff82237R33) for the very rare/temporary case of adding an empty carousel header, the absence of `slides` was breaking the editor, if you have a better version I'll take it

mleray (approved):

PR #623, opened 2021-06-30
PLANET-6198 Convert SocialMedia block to WYSIWYG

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6198

In addition to the WYSIWYG conversion, I've done a couple of things:
- Removed old shortcake conversion code
- Fixed the folders/files/components case from `Socialmedia` to `SocialMedia`
- Split the code for better performance
- Removed `facebook_page_url` attribute since we can directly use `social_media_url` instead

### Testing

Either on local or on the [test instance](https://www-dev.greenpeace.org/test-sinope/) add a Social Media block and make sure all embeds work as expected. You can also directly check it on [this page](https://www-dev.greenpeace.org/test-sinope/social-media-wysiwyg-tests/) that I set up for UAT 🙂


planet-4:

:watch: 2021.06.30 12:59:28

assets/src/blocks/SocialMedia/SocialMediaFrontend.js
```diff
@@ -0,0 +1,30 @@
...
+  description,
+  embed_code,
+  social_media_url,
+  facebook_page_tab,
+  alignment_class,
+  className,
+  embed_type,
+}) => (
+	<section className={`block social-media-block ${className || ''}`}>
+		{title &&
```
GP-Dan-Tovbein, line 14: I'd prefer to use ternaries rather than && and you'll avoid to render `null` or `undefined` values. The logic is the same, It's just a prefrence.
For example these two cases:

`[].length && [].map...` if `[]` is `falsy`, then a `0` will be rendered on this component 
`[].length ? [].map... : null` if `[]` nothing will be rendered on this component. You're being explicit

GP-Dan-Tovbein (approved):

assets/src/blocks/SocialMedia/SocialMediaBlock.js
```diff
@@ -0,0 +1,50 @@
...
+      type: 'string',
+      default: ''
+    },
+    embed_code: {
+      type: 'string',
+      default: '',
+    },
+  },
+  edit: SocialMediaEditor,
+  save: frontendRendered(BLOCK_NAME),
```
Inwerpsel, line 46: Does this need to be front end renderer? :thinking:

mleray: Hmmm yeah good point, I guess we could directly render `SocialMediaFrontend` 🤔

Inwerpsel: I had a quick look at what these embeds contain, and I think we can directly save the html inside of the block as its content, instead of saving the same HTML in the attribute. Only thing to look out for is maybe the scripts some of these would add, which would then run earlier. For example the Twitter embed for me contained this after the `blockquote` element:
```
<script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>

**Review by Inwerpsel — CHANGES_REQUESTED** [review]
We should be able to save the output of `frontendRendered` in the block as it's HTML. Then we don't need to render another block on the front end and further affect performance.

File: assets/src/blocks/SocialMedia/SocialMediaFrontend.js (hunk 1 above)
Comment (Inwerpsel) [reply assets/src/blocks/SocialMedia/SocialMediaFrontend.js:14]: You can also do `!!title &&`, that's a lot easier on the eyes than ternary and no need to add `null` everywhere

Comment (mleray) [reply assets/src/blocks/SocialMedia/SocialMediaFrontend.js]: Updated! 🙂

File: assets/src/blocks/SocialMedia/SocialMediaEditorScript.js [hunk 3]
```diff
@@ -0,0 +1,242 @@
...
+    body.appendChild(tag);
+  });
+};
+
+const initializeTwitterEmbeds = () => {
+  setTimeout(() => {
+    if ('undefined' !== window.twttr) {
+      window.twttr.widgets.load();
+    }
+  }, 2000);
```
Inwerpsel, line 44: Why does this need a timeout?

GP-Dan-Tovbein: I asked me the same, what if the end-user has a slow connection?

But until I don't feel completely familiar with the code base, I won't suggest significant changes.

mleray: I'm not sure why this timeout is needed tbh, it was there before already 🤔 I think @sagarsdeshmukh initially worked on this?

Inwerpsel: It's not a problem to leave the timeout in, also since it's only the editor and not the public JS, so not quite as performance sensitive context. But perhaps the conversion to hooks makes this easier to avoid? If it doesn't then we can maybe look at it later.

sagarsdeshmukh: > I'm not sure why this timeout is needed tbh, it was there before already 🤔 I think @sagarsdeshmukh initially worked on this?

This is the [initial commit](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/commit/43434bfacf5bfa5bab3ee4478ea265f830c29698), but not sure about it.

classes/blocks/class-socialmedia.php
```diff
@@ -107,40 +108,30 @@ public function __construct() {
...
-	public function prepare_data( $fields ): array {
+	private static function get_data( $fields ): array {
 		$title             = $fields['title'] ?? '';
 		$description       = $fields['description'] ?? '';
 		$url               = $fields['social_media_url'] ?? '';
 		$embed_type        = $fields['embed_type'];
 		$alignment_class   = $fields['alignment_class'];
 		$facebook_page_tab = $fields['facebook_page_tab'];
 
-		$data = [
```
mleray, line 118: This was no longer needed since we now use `array_merge` in the [render_callback](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/623/files#diff-749ecefa2f039c9f804d688723f97b5bef853b219013077e09311f6b70395cc4R55)

planet4-gutenberg-blocks.php
```diff
@@ -346,6 +346,19 @@ function ( $posts, $query ) {
...
 
+/**
+ * WordPress automatically adds rel="noopener" to links that have _blank target.
+ * This function removes that functionality since it breaks SocialMedia WYSIWYG blocks with Instagram embeds.
+ *
+ * @param string $rel The link rel.
+ * @param string $link The link.
+ * @return boolean
+ */
+function remove_automatic_link_rel( $rel, $link ) {
```
mleray, line 357: This is the [solution](https://github.com/WordPress/gutenberg/issues/9731#issuecomment-475382493) I've found to solve the block validation failing with Instagram embed, but I'm not sure that it's the right call so any suggestion is welcome 😬

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
```diff
@@ -0,0 +1,244 @@
...
+  useEffect(() => {
+    ALLOWED_OEMBED_PROVIDERS.forEach(provider => {
+      if (social_media_url.includes(provider)) {
+        checkProviderScript(provider);
+        // For Facebook we don't need the embed HTML code since we use an iframe
+        if (provider !== 'facebook') {
+          updateEmbed(social_media_url, provider);
+        }
+      } else {
+        setAttributes({ embed_code: '' });
```
Inwerpsel, line 146: There is no return from the `forEach` in the other cases. Doesn't this mean that it will always set `embed_code` to an empty string in the last iteration (`instragram`), except if it is `instagram`?

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
```diff
@@ -0,0 +1,244 @@
...
+  FACEBOOK_EMBED_TYPE,
+  FACEBOOK_PAGE_TAB_TIMELINE,
+  FACEBOOK_PAGE_TAB_EVENTS,
+  FACEBOOK_PAGE_TAB_MESSAGES,
+  ALLOWED_OEMBED_PROVIDERS,
+} from './SocialMediaConstants.js';
+
+const FB_API_BASE_URL  = 'https://graph.facebook.com/v9.0';
+const INSTAGRAM_OEMBED = `${FB_API_BASE_URL}/instagram_oembed`;
+const FB_ACCESS_TOKEN = window.p4ge_vars.planet4_options.fb_app_access_token;
```
Inwerpsel, line 23: Is this token intended to be distributed publicly? It's entered in the settings here:
https://github.com/greenpeace/planet4-master-theme/blob/a872eb5a98dd58e45da522be48704805ebc845ac/src/Settings.php#L281-L284

Inwerpsel: That line was added for this ticket: https://jira.greenpeace.org/browse/PLANET-5840

Inwerpsel: Their is a link to the documentation about the value that is entered there, that page contains following text:

> Note that because this request uses your app secret, it must never be made in client-side code or in an app binary that could be decompiled. It is important that your app secret is never shared with anyone. Therefore, this API call should only be made using server-side code.

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
```diff
@@ -0,0 +1,244 @@
...
+        embedCode = instagramEmbedData ? instagramEmbedData.html : '';
+      }
+    } catch (error) {
+      embedCode = '';
+    }
+    setAttributes({ embed_code: embedCode });
+  };
+
+  useEffect(() => {
+    ALLOWED_OEMBED_PROVIDERS.forEach(provider => {
```
Inwerpsel, line 138: This would be easier to follow if not everything happened within the `forEach`. You could split it like this:
```js
const provider = ALLOWED_OEMBED_PROVIDERS.find(provider => social_media_url.includes(provider);
if (!provider) {
  setAttributes({ embed_code: '' });
  return;
}
// ... other stuff

File: assets/src/blocks/SocialMedia/SocialMediaScript.js [hunk 9]
```diff
@@ -0,0 +1,10 @@
+import { SocialMediaFrontend } from './SocialMediaFrontend';
```
Inwerpsel, line 1: If I understand correctly, we would only add this script for older blocks, as the newer ones are already rendered and saved. Perhaps it's easier (and better for performance) if we, instead of adding a new front end script, re purpose the existing back end code and Twig template?

assets/src/components/MediaEmbedPreview/MediaEmbedPreview.js
```diff
@@ -0,0 +1 @@
+export const MediaEmbedPreview = ({ html }) => <div dangerouslySetInnerHTML={{ __html: html }}></div>;
```
Inwerpsel, line 1: This component is so simple that it's probably easier to include its contents directly. It would also show more clearly that a `**dangerously**SetInnerHTML` is happening.

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
mleray: Hmmm yes I see the problem 🤔 Should we use an API call here instead then?

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
mleray: Updated!

assets/src/components/MediaEmbedPreview/MediaEmbedPreview.js
mleray: Good point! Updated 🙂

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
Inwerpsel: I'm afraid so, I don't think the tokens can have only read access, since their documentation states (a bit further down)

> Again, for security, app access token should never be hard-coded into client-side code, doing so would give everyone who loaded your webpage or decompiled your app full access to your app secret, and therefore the ability to modify your app.

Inwerpsel: Facebook has a few ways to embed things, maybe we can look at alternatives? The way we're currently using also has some other limitations (for example we're loading the iframe twice because it's not responsive).

Could be we can also use "Client Access Tokens", which are intended to be shipped with clients. https://developers.facebook.com/docs/plugins/oembed#client-access-tokens

mleray: This part of the code is used for the Instagram embed, not the Facebook one, even though the name might be confusing 😅

mleray: Updated the code to use an endpoint for the embed code!

Inwerpsel: Nice! Maybe @sagarsdeshmukh can tell whether we are already using the client token in the Planet 4 settings?

sagarsdeshmukh: Thanks @Pieter for point it 👍  Yes the "App Access Tokens" which is we are currently using is secure token and only recomended to use in backend (serverside code), as Pieter already mention If we want to consume a graph api in client side/frontend we need a "Client Access Tokens" which works exactly same.
@mleray - DM me and I will share you the "Client Access Tokens" :)

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
```diff
@@ -0,0 +1,245 @@
...
+  FACEBOOK_PAGE_TAB_MESSAGES,
+  ALLOWED_OEMBED_PROVIDERS,
+} from './SocialMediaConstants.js';
+
+const { RichText } = wp.blockEditor;
+const { __ } = wp.i18n;
+const { apiFetch } = wp;
+const { addQueryArgs } = wp.url;
+
+const loadScriptAsync = uri => {
```
Inwerpsel, line 25: We could also write this as an `async` function, and use `await`. `useEffect` cannot receive an `async` function, however you can invoke one inside of it. For example: https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/902e5e4322269414a2def75e6399cd679168b61a/assets/src/components/Sidebar/NewThemeSettings.js#L15-L20

mleray: Updated!

Inwerpsel: :+1: I was first also thinking about this function in my comment, however on closer look it might be tricky to write an `async` equivalent of what `tag.onload= () => {resolve()}` does :thinking:

assets/src/blocks/SocialMedia/SocialMediaEditorScript.js
```diff
@@ -0,0 +1,250 @@
...
+  });
+
+  /**
+   * Check if social media corresponding embeds script is loaded and initiliaze it.
+   * Can be used for Facebook, Twitter and Instagram depending on the parameter.
+   */
+  const checkProviderScript = async provider => {
+    const providerData = PROVIDER_SCRIPT_DATA[provider];
+    const script = document.querySelector(`body > script[src="${providerData.script}"]`);
+    if (script === null) {
```
Inwerpsel, line 104: `providerData.initFunction` is called in both cases, so it can be after the if-statement once.

classes/blocks/class-socialmedia.php
```diff
@@ -95,17 +79,22 @@ public function __construct() {
...
+
+		add_action( 'enqueue_block_editor_assets', [ self::class, 'enqueue_editor_assets' ] );
+		add_action( 'wp_enqueue_scripts', [ self::class, 'enqueue_frontend_assets' ] );
 	}
 
 	/**
-	 * Get all the data that will be needed to render the block correctly.
-	 *
-	 * @param array $fields This is the array of fields of this block.
+	 * Required by the `Base_Block` class.
```
Inwerpsel, line 95: Maybe you can just mention here that we're keeping this for existing content.

classes/blocks/class-socialmedia.php
```diff
@@ -95,17 +79,22 @@ public function __construct() {
...
 	 *
-	 * @return array The data to be passed in the View.
+	 * @param array $fields Unused, required by the abstract function.
```
Inwerpsel, line 97: `$fields` is actually used, maybe you meant something else that isn't used? Also the return parameter should still be in the doc comment I think, not sure why it doesn't make the lint job fail :confused:

Inwerpsel (approved):
Looks good :beer:

PR #624, opened 2021-07-02
PLANET-6124: Fix the "open in a new tab" functionality through from settings

GP-Dan-Tovbein:
Fix the "open in a new tab" functionality for external domains on <a> selectors.
Also fixed this functionality on `headings` and `images`

Issue found [here](https://www-dev.greenpeace.org/test-sinope/columns-with-links/)
Issue replicated and solved [here](https://www-dev.greenpeace.org/test-titan/columns-with-links/)
Issue replicated and solved (with images) [here](https://www-dev.greenpeace.org/test-titan/29060-2/)

Linked PR https://github.com/greenpeace/planet4-master-theme/pull/1411

Ref: https://jira.greenpeace.org/browse/PLANET-6124

### DIFF
--- a/assets/src/blocks/Columns/Columns.js
+++ b/assets/src/blocks/Columns/Columns.js
@@ -42,10 +42,10 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample = 
                 : cta_link ?
                   <a
                     href={cta_link}
-                    target={link_new_tab && link_new_tab !== 'false' ? '_blank' : ''}
                     data-ga-category='Columns Block'
                     data-ga-action={columns_block_style === LAYOUT_ICONS ? 'Icon' : 'Image'}
                     data-ga-label={cta_link}
+                    { ...link_new_tab && { target: '_blank' } }
                   >
                     <img src={attachment} alt={title} title={title} loading='lazy' />
                   </a> :
@@ -60,6 +60,7 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample = 
                   data-ga-category='Columns Block'
                   data-ga-action='Title'
                   data-ga-label={cta_link}
+                  { ...link_new_tab && { target: '_blank' } }
                 >
                   {title}
                 </a> :
@@ -72,7 +73,6 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample = 
             {cta_text && cta_link &&
               <a
                 href={cta_link}
-                target={link_new_tab ? '_blank' : ''}
                 className={isCampaign || columns_block_style === LAYOUT_NO_IMAGE ?
                   `btn btn-${isCampaign ? 'primary' : 'secondary'}` :
                   'call-to-action-link'
@@ -80,6 +80,7 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample = 
                 data-ga-category='Columns Block'
                 data-ga-action='Call to Action'
                 data-ga-label={cta_link}
+                { ...link_new_tab && { target: '_blank' } }
               >
                 {cta_text}
               </a>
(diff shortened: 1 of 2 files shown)


planet-4:

:new_moon: [titan](https://www-dev.greenpeace.org/test-titan)

:watch: 2021.07.02 17:31:19

mleray (approved):
I've already suggested different changes for the corresponding master-theme PR, but this one looks good to me 👍

Inwerpsel:
@GP-Dan-Tovbein Do you know how long the previous incorrect behavior was present? The Jira ticket description says: 

> Currently if users leave the "open in a new tab" option unchecked, and the link is an external domain, the link opens in a new tab anyway.

So people never explicitly unchecked the option. Likely opening in a new tab is their desired behavior anyway, at least in some cases (maybe in most), even though they didn't choose it (because they didn't have to).

dantovbein:
@Inwerpsel I'm not sure how long it was present, honestly. And I completely agree with you about the expected behavior for external links, however, I think you should give to the user the chance for choosing if they want to open in a new tab or not as an extra functionality.

Just to take in mind that this pull request solved the inconsistent functionality for `headings` and `images` against `links` because they were working different.

Inwerpsel:
> ... I think you should give to the user the chance for choosing if they want to open in a new tab or not as an extra functionality.

My point was that for content that was created prior to this change, they likely didn't even see the option in most cases, and it defaults to not opening in a new tab. For those cases (only existing content created prior to this change), having it respect the option is possibly not what the author intended to do. So we may want to ask people to review older posts that have a columns block, so that they can enable opening in a new tab if that's what they originally intended.

dantovbein:
Got it. You're completely right!
How should we ask to editors about this change?
Nikos shared me this [reports file](https://docs.google.com/spreadsheets/d/1uAmZLIWYsxrBByqbhoF_vVtSM7WGebYWIc0xftPRPwE/edit#gid=359432436), might we check whose are using the columns block component? Does it make sense?

comzeradd:
> How should we ask to editors about this change?

It's hard to confirm what was the editor's proffered choice for all pre-existing content. Some may indeed missed the option completely, but others could intentionally left this option unchecked. Whatever choice we make it's just an assumption at this point, so I think it's better to just fix the bug and not touch existing content.

Inwerpsel:
We could just reach out to everyone through Slack and ask them to check. Or we could write a query/script that detects external links in columns. Maybe the latter is worth the effort, then we can hand a list to people and they don't need to go manually through a whole lot of columns blocks.

Inwerpsel:
> > How should we ask to editors about this change?
> 
> It's hard to confirm what was the editor's proffered choice for all pre-existing content. Some may indeed missed the option completely, but others could intentionally left this option unchecked. Whatever choice we make it's just an assumption at this point, so I think it's better to just fix the bug and not touch existing content.

@comzeradd Do you mean "not touch" as in we should ensure that existing content doesn't start displaying differently? Or "not touch" as in we shouldn't do an effort to preserve the behavior of existing content?

comzeradd:
> Do you mean "not touch" as in we should ensure that existing content doesn't start displaying differently? Or "not touch" as in we shouldn't do an effort to preserve the behavior of existing content?

The latter :) Because if we check that option for existing content, then it will have undesired behavior for those who intentionally left that option unchecked. Since the option is there already (though buggy) we should expect that people pay attention to block settings.

Inwerpsel:
> > Do you mean "not touch" as in we should ensure that existing content doesn't start displaying differently? Or "not touch" as in we shouldn't do an effort to preserve the behavior of existing content?
> 
> The latter :) Because if we check that option for existing content, then it will have undesired behavior for those who intentionally left that option unchecked. Since the option is there already (though buggy) we should expect that people pay attention to block settings.

:+1: I guess we should at least flag editors that we're changing the behavior of content they already published, so that they don't think the change in behavior is a bug.

PR #625, opened 2021-07-05
PLANET-5662 Split webpack into multiple configs

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-5662

We can merge this before https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/603, then we should be able to remove the webpack changes from that PR and rebase. After that it should be mergeable :crossed_fingers: But before we proceed with that PR, we need to sync with @oekeur to see if this requires changes from their part. Probably they will have to mimic our alias config in both webpack and imports.

* This is needed because we want to have separate aliases for public and
admin JS, so that we can use Preact for the former and WP's bundled
React for the latter.
* Also split off the CSS config, because we can. All config in the
previous mixed object was either applying to JS or to CSS, never to
both.
* Include blocks' entries one by one instead of using the template. This
wouldn't work for the split config, and there were already cases where
not all files were needed for a block. This was previously caught with a
file system check, which can make it hard to figure out what's
happening, and impossible to both have the file local and not build it.
* Remove the plugins that remove certain build artifacts. This was
probably added when the built files were checked into Git. Now that we
build these in CI this shouldn't be needed anymore. These are all small
files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "node-sass": "^4.14.1",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "postcss-loader": "^3.0.0",
-    "remove-files-webpack-plugin": "^1.1.3",
     "sass-loader": "^7.3.1",
     "terser-webpack-plugin": "^1.4.1",
     "webpack-fix-style-only-entries": "^0.3.1"
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,9 +3,7 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
 const TerserJSPlugin = require('terser-webpack-plugin');
 const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
-const RemovePlugin = require('remove-files-webpack-plugin');
 const dashDash = require('@greenpeace/dashdash');
-const fs = require('fs');
 
 const mediaQueryAliases = {
   '(max-width: 576px)': 'mobile-only',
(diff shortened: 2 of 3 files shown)


webpack.config.js
```diff
@@ -111,36 +148,10 @@ module.exports = {
       ignoreOrder: false, // Enable to remove warnings about conflicting order
       filename: './[name].min.css'
     }),
-    new RemovePlugin({
```
Inwerpsel, line 114: Not needed anymore, see commit message and PR description.

planet-4:

:new_moon: [saturn](https://www-dev.greenpeace.org/test-saturn)

:watch: 2021.07.05 11:12:56

webpack.config.js
```diff
@@ -44,20 +78,23 @@ module.exports = {
...
     theme_oil: './assets/src/styles/theme_oil.scss',
     theme_plastic: './assets/src/styles/theme_plastic.scss',
-    ...entryPoints('Accordion'),
-    ...entryPoints('Covers'),
-    ...entryPoints('CarouselHeader'),
-    ...entryPoints('Spreadsheet'),
-    ...entryPoints('Timeline'),
+    AccordionStyle: './assets/src/styles/blocks/Accordion/AccordionStyle.scss',
+    AccordionEditorStyle: './assets/src/styles/blocks/Accordion/AccordionEditorStyle.scss',
+    // CoversStyle: './assets/src/styles/blocks/Covers/CoversStyle.scss',
```
Inwerpsel, line 83: I commented these because the file exists, but it is empty. Creating the empty file was a workaround that we needed to do because the `entryPoints` template function only outputted a fixed set of file names. I'll probably remove them again.

webpack.config.js
```diff
@@ -111,36 +145,10 @@ module.exports = {
...
-                return new RegExp(item, 'm').test(filePath);
-              }).length > 0;
-            }
-          }
-        ]
-      }
-    }),
   ],
   optimization: {
-    ...defaultConfig.optimization,
```
Inwerpsel, line 140: I checked the default config we're currently using, and it didn't have the `optimization` key, so this line didn't add anything. However when I check the latest version of Gutenberg, this file now does include it. https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/config/webpack.config.js#L108-L142

Inwerpsel: Now it still seems better to not include it from the default config for now, otherwise it might result in surprises that could go unnoticed when we update to a later version of WordPress that includes this new base file. Probably a high chance that it doesn't play well with the config we already have.

webpack.config.js
```diff
@@ -3,9 +3,7 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
 const TerserJSPlugin = require('terser-webpack-plugin');
 const FixStyleOnlyEntriesPlugin = require("webpack-fix-style-only-entries");
-const RemovePlugin = require('remove-files-webpack-plugin');
```
mleray, line 6: Should we remove this from `package.json` too? It doesn't seem to be used anywhere else 🤔

webpack.config.js
```diff
@@ -15,27 +13,63 @@ const mediaQueryAliases = {
...
+  },
 };
 
-module.exports = {
-  ...defaultConfig,
+const publicJsConfig = {
+  ...jsConfig,
+  resolve: {
+    alias: {
+      '@hooks': 'preact/hooks',
```
Inwerpsel, line 33: We don't use these yet so it shouldn't be an issue to add the alias mapping already.

mleray (approved):
Tested on local, doesn't seem to break anything 👍

package.json
```diff
@@ -34,7 +34,6 @@
     "node-sass": "^4.14.1",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "postcss-loader": "^3.0.0",
-    "remove-files-webpack-plugin": "^1.1.3",
```
GP-Dan-Tovbein, line 37: Instead of removing, I suggest to uninstall the `remove-files-webpack-plugin` package and then you'll have the package-lock.json updated with the correct dependencies.

GP-Dan-Tovbein: This file has an inconsistency against the `package-lock.json`

webpack.config.js
```diff
@@ -15,27 +13,63 @@ const mediaQueryAliases = {
...
-    [`${blockName}Script`]: `./assets/src/blocks/${blockName}/${blockName}Script.js`,
-    [`${blockName}Style`]: `./assets/src/styles/blocks/${blockName}/${blockName}Style.scss`,
-  };
-  const pathExists = ([,path]) => fs.existsSync(path);
-  return Object.fromEntries(Object.entries(tpl).filter(pathExists));
+const jsConfig = {
+  ...defaultConfig,
+  output: {
+    filename: '[name].js',
+    path: __dirname + '/assets/build'
```
GP-Dan-Tovbein, line 20: It's missing a `,` at the EOL

Inwerpsel: Well spotted, I had just moved this whole object without looking :sweat_smile:

webpack.config.js
```diff
@@ -15,27 +13,63 @@ const mediaQueryAliases = {
...
+      '@hooks': 'preact/hooks',
+      '@render': 'preact',
+      '@compat': 'preact/compat',
+    }
+  },
   entry: {
-    editorIndex: './assets/src/editorIndex.js',
     frontendIndex: './assets/src/frontendIndex.js',
     carouselHeaderFrontIndex: './assets/src/carouselHeaderFrontIndex.js',
+    AccordionScript: './assets/src/blocks/Accordion/AccordionScript.js',
```
GP-Dan-Tovbein, line 41: We can define a `const` for `blocks` instead of duplicating text. It could be `const BLOCKS_PATH='./assets/src/blocks/'`

Inwerpsel: Idk, since a webpack config contains objects that are many lines long, it's actually harder to understand the file with the const being out of view most of the time.

webpack.config.js
```diff
@@ -44,20 +78,20 @@ module.exports = {
     theme_oceans: './assets/src/styles/theme_oceans.scss',
     theme_oil: './assets/src/styles/theme_oil.scss',
     theme_plastic: './assets/src/styles/theme_plastic.scss',
-    ...entryPoints('Accordion'),
-    ...entryPoints('Covers'),
-    ...entryPoints('CarouselHeader'),
-    ...entryPoints('Spreadsheet'),
-    ...entryPoints('Timeline'),
+    AccordionStyle: './assets/src/styles/blocks/Accordion/AccordionStyle.scss',
```
GP-Dan-Tovbein, line 81: You can do sth like this and you'll avoid repeating code

```
{
  entry: {
    theme_oceans: './assets/src/styles/theme_oceans.scss',
    theme_oil: './assets/src/styles/theme_oil.scss',
    theme_plastic: './assets/src/styles/theme_plastic.scss',
    ...['AccordionStyle', 'SpreadsheetStyle'].reduce((b,c) => {
      if(c) {
        return {
          ...b,
          [c]: `./assets/src/styles/blocks/${c}/${c}Style.scss`,
        };
      }
    }, []),
  },
}
```

GP-Dan-Tovbein:
@Inwerpsel Tested locally and works fine! Good job!

GP-Dan-Tovbein (approved):
Tested locally and works fine. Good job!

package.json
Inwerpsel: As discussed on Slack this turned out to be not an issue.

webpack.config.js
Inwerpsel: For a webpack config I'd rather keep it simple. A reducer would make the file significantly harder to read for newcomers.

Inwerpsel: Btw, why is it a reducer and not just a map? :thinking: `if (c)` will always be true for an array of non empty string literals

webpack.config.js
Inwerpsel: Ah, forgot to push this before merging :facepalm:

webpack.config.js
dantovbein: Yes, It's a good point. We can also use `map`, however, take in mind human errors 👀 if you use an Array 😄

PR #628, opened 2021-07-22
Add CSS variables to Articles, Columns and Covers

Inwerpsel:

---

Some variables that are needed to reproduce the themes. CSS structure is still the same, so these changes should be safe. Except for a redundant color rule on ContentCovers which was removed.

It would be better if we clean up the block styles, currently they're very nested. But I added the variables to this structure anyway, since it's the only thing blocking us from removing the legacy.

Somewhat easier [without whitespace changes](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/628/files?diff=split&w=1)

### DIFF
--- a/assets/src/styles/blocks/Articles.scss
+++ b/assets/src/styles/blocks/Articles.scss
@@ -2,9 +2,9 @@
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  margin-top: $space-lg;
 
   --block-articles--article-- {
+    margin-top: $space-lg;
     background-color: transparentize($white, .2);
     box-shadow: 0 3px 10px rgba(0, 0, 0, 0.1);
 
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2021.07.30 10:34:01

assets/src/styles/blocks/Columns.scss
```diff
@@ -234,7 +242,7 @@
       }
 
       .step-info {
-        h5 > a {
+        h5 > a _-- {
```
Inwerpsel, line 245: I kept this one a shorthand because it's not actually used in any lives campaigns. Plastics currently uses it, but Ana will be converting the usages to another style.

Eventually we'll probably want to remove this and unify the style (and heading level) with the other Columns styles.

assets/src/styles/blocks/Articles.scss
```diff
@@ -119,10 +121,10 @@
 .article-list-item-content {
   --block-articles--article--content-- {
     font-size: inherit;
+    width: 80%;
   }
   display: none;
   margin: $space-xs 0 0;
```
mleray, line 127: Can't this margin also be a variable? 🤔

Inwerpsel: It can, but for now I only added variables for spacing changes where a theme needs these (and it made sense). We probably want to clean up some blocks CSS after we converted legacy themes, so the spacing CSS might look different.

assets/src/styles/blocks/Columns.scss
```diff
@@ -204,26 +208,30 @@
         }
 
         @include large-and-up {
+          --block-columns--step-number-- {
+            background-color: $dark-blue;
```
mleray, line 212: Don't you usually prefer to use `background` instead of `background-color` for more flexibility?

Inwerpsel: Changed it :+1:

mleray (approved):
Looks good to me 👍

PR #630, opened 2021-07-27
Un-nest SCSS of ContentCovers

Inwerpsel:
This was already using specific enough class names so that it doesn't need to also be very nested. Too deeply nested SCSS makes code very hard to read. It also increases high specificity, making it challenging to use in child themes.

Not a lot of changes if you view [with whitespace changes ignored](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/630/files?diff=split&w=1).

Ref: 

---

### DIFF
--- a/assets/src/styles/blocks/Covers/ContentCovers.scss
+++ b/assets/src/styles/blocks/Covers/ContentCovers.scss
@@ -31,95 +31,95 @@
     }
   }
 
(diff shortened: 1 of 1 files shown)


planet-4:

:new_moon: [janus](https://www-dev.greenpeace.org/test-janus)

:watch: 2021.07.27 09:42:24

PR #631, opened 2021-07-27
WIP Use h3 for tasks style as well

Inwerpsel:
Other styles use h3 and it's used for the same content. Probably will allow to reduce some duplication. Still need to verify the style looks ok with the designs.

Luckily it turned out not needed for the legacy conversion, as just 2 pages use this block style and they will be converted to use another style.

Ref: 

---


planet-4:

:new_moon: [telesto](https://www-dev.greenpeace.org/test-telesto)

:watch: 2021.07.27 11:28:28

Inwerpsel:
I'll close this as we probably should be looking at the core Columns block for future content.

PR #633, opened 2021-07-27
Remove h1 styles in Gallery block

Inwerpsel:
* It looks like the current code doesn't put an h1 inside of
.carousel-wrap, so this CSS is unused.

I'll check if any more things can be removed before submitting for review.

Ref:


planet-4:

:new_moon: [proteus](https://www-dev.greenpeace.org/test-proteus)

:watch: 2021.07.27 12:09:27

PR #641, opened 2021-08-06
Hubspot Form block base implementation

GP-Dan-Tovbein:
**Description**
---
It's about the new custom [Hubspot Form Block](https://jira.greenpeace.org/browse/PLANET-6190).

For now is only implemented as the default the [Image Full Width](https://p4-designsystem.greenpeace.org/05f6e9516/p/213df0-hubspot-forms/b/99e047) form, but easy to be scalable. 
The rest of the forms will be added once the base implementation is approved. 

-> [Demo page](https://www-dev.greenpeace.org/test-deimos/hubspot-form-block)
-> [Design](https://p4-designsystem.greenpeace.org/05f6e9516/p/213df0-hubspot-forms/b/99e047)

Ref: [Ticket](https://jira.greenpeace.org/browse/PLANET-6190)

If you want to create a new **Hubspot Form** please go [here](https://www-dev.greenpeace.org/dmptest/wp-admin/admin.php?page=leadin_forms&leadin_route%5B0%5D=type&leadin_search=businessUnitId%3D%26dateType%3DLAST_THIRTY_DAYS%26formType%3DALL%26offset%3D0%26parentId%3D0%26query%3D%26sortDirection%3DDESCENDING%26sortingColumn%3DLAST_MODIFIED)

:warning: **Things to keep in mind**
---
- The custom form design will be covered on [this](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/662) pull request.

**Testing**
---
- Make sure you already have a portal id and id related to a Hubspot from before starting.
- Enable `beta` blocks from P4 settings.

**Related PRs** (are in draft)
---
- Use with custom hubspot form styles (overriding the default ones): https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/662
- Implement the Thank you section. https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/663

**Updates**
---
- **September, 9:**
  - Implemented a custom hook to create dynamically the form instead to adding this logic inside the Compoment.
  - Enable to show a custom _Thank you message_ or show the one retrieved by the API.
  - Improve the Hubspot shortcode parser. It can parse 3 different shortcodes (with or without `' and "`)
  ```
  [hubspot type=form portal=8710305 id=85158a32-72b6-4dd7-8012-f4f4d8ef08d1]
  [hubspot type='form' portal='8710305' id='85158a32-72b6-4dd7-8012-f4f4d8ef08d1']
  [hubspot type="form" portal="8710305" id="85158a32-72b6-4dd7-8012-f4f4d8ef08d1"]
  ```
  - ~Show a preloader when the Hubspot form is fetched.~

---

### DIFF
--- a/assets/src/blocks/HubspotForm/HubspotFormBlock.js
+++ b/assets/src/blocks/HubspotForm/HubspotFormBlock.js
@@ -1,0 +1,110 @@
+import ReactDOMServer from 'react-dom/server';
+import { Tooltip } from '@wordpress/components';
+import { HubspotFormEditor } from './HubspotFormEditor';
+import { HubspotFormFrontend } from './HubspotFormFrontend';
+const { registerBlockType } = wp.blocks;
+const { __ } = wp.i18n;
+
+const BLOCK_NAME = 'planet4-blocks/hubspot-form';
+
+const getStyleLabel = (label, help) => (
+  (help)
+    ? <Tooltip text={help}>
+        <span>{label}</span>
+      </Tooltip>
+    : label
+);
+
+export const registerHubspotFormBlock = () => {
+  return registerBlockType(BLOCK_NAME, {
+    title: 'Hubspot Form (beta)',
+    icon: 'feedback',
+    category: 'planet4-blocks-beta',
+    supports: {
+      multiple: false,
+      html: false,
+    },
+    attributes: {
+      blockTitle: {
+        type: 'string',
+      },
+      blockText: {
+        type: 'string',
+      },
+      blockBackgroundImageId: {
+        type: 'integer',
+      },
+      blockBackgroundImageUrl: {
+        type: 'string',
+      },
+      blockStyle: {
+        type: 'string',
+        default: 'image-full-width',
+      },
+      ctaText: {
+        type: 'string',
+      },
+      ctaLink: {
+        type: 'string',
+      },
+      ctaNewTab: {
+        type: 'boolean',
+        default: false,
+      },
+      formTitle: {
+        type: 'string',
+      },
+      formText: {
+        type: 'string',
+      },
+      hubspotShortcode: {
+        type: 'string',
+        default: '',
+      },
+      hubspotThankyouMessage: {
+        type: 'string',
+      },
+      enableCustomHubspotThankyouMessage: {
+        type: 'boolean',
+        default: false,
+      },
+      version: {
+        type: 'integer',
+        default: 1,
+      },
+    },
+    styles: [
+      {
+        name: 'image-full-width',
+        label: getStyleLabel(
+          __('Image full width', 'planet4-blocks-backend'),
+          __('https://p4-designsystem.greenpeace.org/05f6e9516/p/213df0-hubspot-forms/b/99e047', 'planet4-blocks-backend'),
+        ),
+        isDefault: true,
+      },
+    ],
+    edit: HubspotFormEditor,
+    save: (props) => {
+      /**
+       * This parser is added cause the Hubspot plugin takes the shortcode, by a hook,
+       * and converts it into a <script>. In consequence, it fails when it is parsed to json through the hydration.
+       *
+       * This parser is only affected to the hydration.
+       *
+       * Ideally, we should use innerBlocks but it has various reported conflicts using SSR.
+       */
+      const attributes = {...props.attributes};
+      attributes.hubspotShortcode = props.attributes.hubspotShortcode.replace('[', '').replace(']', '');
+
+      const markup = ReactDOMServer.renderToString(
+        <div
+          data-hydrate={BLOCK_NAME}
+          data-attributes={JSON.stringify(attributes)}
+        >
+          <HubspotFormFrontend {...props} />
+        </div>
+      );
+      return <wp.element.RawHTML>{ markup }</wp.element.RawHTML>;
+    },
+  })
+}
--- a/assets/src/blocks/HubspotForm/HubspotFormEditor.js
+++ b/assets/src/blocks/HubspotForm/HubspotFormEditor.js
@@ -1,0 +1,147 @@
+import { Fragment, useEffect } from '@wordpress/element';
+import { RichText } from '@wordpress/block-editor';
+import { Sidebar } from './HubspotFormSidebar';
+import { useToAttribute } from './hooks/useToAttribute';
+import { useBackgroundImage } from './hooks/useBackgroundImage';
+import { getStyleFromClassName } from '../getStyleFromClassName';
+
+const { __ } = wp.i18n;
+
+/**
+ * This component is created as a workaround because the <RichText> doesn't provide
+ * The functionality to set max characters like any content editable element.
+ */
+const MaxLengthHelperComponent = ({ value, maxLength, children, darkTheme = false }) => (
+  <div>
+    { children }
+    {maxLength && <span
+      className={`max-length-message ${darkTheme ? 'dark-theme' : ''} ${value.length >= maxLength ? 'max-length-error' : ''}`}>
+      {value.length >= maxLength
+        ? __('Maximum of characters', 'planet4-blocks-backend')
+        : `${value.length} / ${maxLength}`}
+    </span>}
+  </div>
+)
+
+export const HubspotFormEditor = ({
+  attributes: {
+    blockBackgroundImageId,
+    blockBackgroundImageUrl,
+    blockText = '',
+    blockTitle,
+    blockStyle,
+    ctaText,
+    ctaLink,
+    ctaNewTab,
+    formText = '',
+    formTitle,
+    hubspotShortcode,
+    hubspotThankyouMessage,
+    enableCustomHubspotThankyouMessage,
+    className,
+  },
+  setAttributes,
+}) => {
+  const backgroundImage = useBackgroundImage(blockBackgroundImageUrl);
+  const toAttribute = useToAttribute(setAttributes);
+
+  useEffect(() => {
+    if(className) {
+      setAttributes({ blockStyle: getStyleFromClassName(className) });
+    }
+  }, [ className ]);
+
+  return (
+    <Fragment>
+      <Sidebar {...{
+        ctaLink,
+        ctaNewTab,
+        blockBackgroundImageId,
+        enableCustomHubspotThankyouMessage,
+        hubspotThankyouMessage,
+        setAttributes
+      }} />
+      <section className={`hubspot-form hubspot-form-editor block-wide ${blockStyle}`} style={{...backgroundImage}}>
+        <div className='container'>
+          <div className='block-wrapper'>
+            <div className='block-wrapper-inner block-wrapper-text' style={{...backgroundImage}}>
+              <div className='container'>
+                <RichText
+                  tagName='h1'
+                  className='block-title'
+                  placeholder={__('Enter title', 'planet4-blocks-backend')}
+                  value={blockTitle}
+                  onChange={toAttribute('blockTitle')}
+                  withoutInteractiveFormatting={true}
+                  allowedFormats={[]}
+                  multiline='false'
+                />
+                <MaxLengthHelperComponent value={blockText} maxLength={600} darkTheme={true}>
+                  <RichText
+                    tagName='p'
+                    className='block-text'
+                    placeholder={__('Enter description', 'planet4-blocks-backend')}
+                    value={blockText}
+                    onChange={toAttribute('blockText', 600)}
+                    withoutInteractiveFormatting={true}
+                    allowedFormats={['core/bold', 'core/italic']}
+                  />
+                </MaxLengthHelperComponent>
+                <RichText
+                  tagName='div'
+                  className='block-button'
+                  placeholder={__('Enter CTA text', 'planet4-blocks-backend')}
+                  value={ctaText}
+                  onChange={toAttribute('ctaText')}
+                  withoutInteractiveFormatting
+                  allowedFormats={[]}
+                />
+              </div>
+            </div>
+            <div className='block-wrapper-inner block-wrapper-form'>
+              <div className='container'>
+                <header className='form-header'>
+                  <RichText
+                    tagName='h1'
+                    className='form-title'
+                    placeholder={__('Enter form title', 'planet4-blocks-backend')}
+                    value={formTitle}
+                    onChange={toAttribute('formTitle')}
+                    withoutInteractiveFormatting={true}
+                    allowedFormats={[]}
+                    multiline='false'
+                  />
+                  <MaxLengthHelperComponent value={formText} maxLength={300}>
+                    <RichText
+                      tagName='p'
+                      className='form-text'
+                      placeholder={__('Enter form description', 'planet4-blocks-backend')}
+                      value={formText}
+                      onChange={toAttribute('formText', 300)}
+                      withoutInteractiveFormatting={true}
+                      allowedFormats={['core/bold', 'core/italic']}
+                    />
+                  </MaxLengthHelperComponent>
+                </header>
+                <div className='form-wrapper-editor'>
+                  <div className='form-wrapper-field'>
+                    <label>{__('Paste the Hubspot shortcode here', 'planet4-blocks-backend')}</label>
+                    <RichText
+                      tagName='p'
+                      className='hubspot-form-shortcode-editor'
+                      placeholder={__('[hubspot type="form" portal="XXXXXX" id="XXXX-XXXX-XXXX-XXXX"]', 'planet4-blocks-backend')}
+                      value={hubspotShortcode}
+                      onChange={toAttribute('hubspotShortcode')}
+                      withoutInteractiveFormatting={true}
+                      allowedFormats={[]}
+                    />
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </section>
+    </Fragment>
+  );
+};
--- a/assets/src/blocks/HubspotForm/HubspotFormFrontend.js
+++ b/assets/src/blocks/HubspotForm/HubspotFormFrontend.js
@@ -1,0 +1,70 @@
+import { useRef, useEffect, useState } from '@wordpress/element';
+import { useHubspotForm } from './hooks/useHubspotForm';
+import { useBackgroundImage } from './hooks/useBackgroundImage';
+
+const { __ } = wp.i18n;
+
+export const HubspotFormFrontend = ({
+  formTitle,
+  formText,
+  blockBackgroundImageUrl,
+  blockTitle,
+  blockText,
+  blockStyle,
+  ctaLink,
+  ctaText,
+  ctaNewTab,
+  hubspotShortcode,
+  hubspotThankyouMessage,
+  enableCustomHubspotThankyouMessage,
+}) => {
+  const hubspotFormRef = useRef(null);
+  const [ styleClass, setStyleClass ] = useState('');
+  const { submitted, submittedMessage } = useHubspotForm(hubspotShortcode, hubspotFormRef);
+  const backgroundImage = useBackgroundImage(blockBackgroundImageUrl);
+
+  useEffect(() => {
+    if(blockStyle) {
+      setStyleClass(blockStyle);
+    }
+  }, [ blockStyle ]);
+
+  return (
+    <section className={`hubspot-form block-wide ${styleClass}`} style={{...backgroundImage}}>
+      <div className='container'>
+        <div className='block-wrapper'>
+          <div className='block-wrapper-inner block-wrapper-text' style={{...backgroundImage}}>
+            <div className='container'>
+              <h1 className='block-title'>{ blockTitle }</h1>
+              {blockText && <p className='block-text' dangerouslySetInnerHTML={{ __html: blockText }} />}
+              {(ctaLink && ctaText) && <a
+                href={ctaLink}
+                className='block-button'
+                data-ga-category='Hubspot Forms Block'
+                data-ga-action='custom form'
+                data-ga-label={ctaLink}
+                { ...ctaNewTab && { target: '_blank' } }
+                >{ctaText}</a>}
+            </div>
+          </div>
+          <div className={`block-wrapper-inner block-wrapper-form ${submitted ? 'submitted-form' : ''}`}>
+            <div className='container'>
+              {!submitted && <header className='form-header'>
+                <h1 className='form-title'>{formTitle}</h1>
+                <p className='form-text'>{formText}</p>
+              </header>}
+              {!submitted ? <div className='form-wrapper'>
+                <div id='hubspot-api-form' ref={hubspotFormRef} />
+              </div> : <div className='submitted-message'>
+                <svg className='icon-tick' width="24" height="20" viewBox="0 0 24 20" fill="none" xmlns="http://www.w3.org/2000/svg">
+                  <path d="M2 10.4211L8.4 18L22 2" stroke="#074365" strokeWidth="4" strokeLinecap="round" strokeLinejoin="round"/>
+                </svg>
+                <span>{ enableCustomHubspotThankyouMessage ? hubspotThankyouMessage : submittedMessage }</span>
+              </div>}
+            </div>
+          </div>
+        </div>
+      </div>
+    </section>
+  );
+};
--- a/assets/src/blocks/HubspotForm/HubspotFormSidebar.js
+++ b/assets/src/blocks/HubspotForm/HubspotFormSidebar.js
@@ -1,0 +1,107 @@
+import { CheckboxControl, PanelBody, PanelRow, Button, TextControl } from '@wordpress/components';
+import { MediaUpload, MediaUploadCheck, InspectorControls } from '@wordpress/block-editor';
+import { URLInput } from '../../components/URLInput/URLInput';
+import { useToAttribute } from './hooks/useToAttribute';
+
+const { __ } = wp.i18n;
+
+const PanelRowWrapper = ({ labelText, helpText, children }) => {
+  return (
+    <div className='panel-row-wrapper'>
+      {labelText && <label className='panel-row-label-text'>{labelText}</label>}
+      {helpText && <span className='panel-row-help-text'>{helpText}</span>}
+      {children}
+    </div>
+  )
+}
+
+export const Sidebar = ({
+  ctaLink,
+  ctaNewTab,
+  blockBackgroundImageId,
+  enableCustomHubspotThankyouMessage,
+  hubspotThankyouMessage,
+  setAttributes,
+}) => {
+  const toAttribute = useToAttribute(setAttributes);
+
+  const onSelectImageHandler = ({ id, url }) => {
+    if(url && id) {
+      setAttributes({
+        blockBackgroundImageId: id,
+        blockBackgroundImageUrl: url,
+      });
+    }
+  };
+
+  const onRemoveImageHandler = () => {
+    setAttributes({
+      blockBackgroundImageId: null,
+      blockBackgroundImageUrl: null,
+    });
+  }
+
+  return <InspectorControls>
+    <PanelBody title={__('Settings', 'planet4-blocks-backend')}>
+      <PanelRow>
+        <PanelRowWrapper labelText={__('Call to action', 'planet4-blocks-backend')}>
+          <URLInput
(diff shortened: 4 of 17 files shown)


planet-4:

:new_moon: [deimos](test-deimos) | [admin](https://www-dev.greenpeace.org/test-deimos/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-deimos/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-deimos) | [composer-local.json](https://github.com/greenpeace/planet4-test-deimos/blob/main/composer-local.json)

:watch: 2021.08.12 12:03:21

assets/src/blocks/HubspotForm/HubspotFormEditor.js
```diff
@@ -0,0 +1,260 @@
...
+  },
+  setAttributes,
+}) => {
+
+  const onChangeHandler = (evt) => {
+    evt.preventDefault();
+    setAttributes({ hubspot_portal: evt.currentTarget.value });
+  }
+
+  return useMemo(() => (
```
Inwerpsel, line 37: Why do we have to use `useMemo` here?

assets/src/blocks/HubspotForm/HubspotFormBlock.js
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <Tooltip text={help}>
+        <span>{label}</span>
+      </Tooltip>
+    );
+  }
+  return label;
+};
+
+export class HubspotFormBlock {
```
Inwerpsel, line 22: We don't use class components in new code any more. Can you change this to a function component?

assets/src/blocks/HubspotForm/HubspotFormEditor.js
Inwerpsel: I see it's used pretty much be default. While `useMemo` **can** in some cases be used to address a performance issue if there's a particularly expensive calculation, in most cases it actually reduces performance. Please check  https://medium.com/swlh/should-you-use-usememo-in-react-a-benchmarked-analysis-159faf6609b7 to get an idea of the complexity needed in the calculation for `useMemo` to actually improve performance.

assets/src/blocks/HubspotForm/HubspotFormContext.js
```diff
@@ -0,0 +1,69 @@
...
+    evt.preventDefault();
+    updateBlockHeight();
+  }, []);
+
+  useEffect(() => {
+    console.log(wrapperRef);
+    updateBlockHeight();
+  }, [
+    wrapperRef,
+  ])
```
Inwerpsel, line 40: Why do we need an effect to run on every render to update the block height?

assets/src/editorIndex.js
```diff
@@ -24,32 +24,37 @@ import { SubPagesBlock } from './blocks/SubPages/SubPagesBlock';
...
-registerGuestBookBlock();
-
-addBlockFilters();
-setupImageBlockExtension();
-addButtonLinkPasteWarning();
-replaceTaxonomyTermSelectors();
-setupCustomSidebar();
-setUpCssVariables();
-blockEditorValidation();
+try {
```
Inwerpsel, line 29: Why do we need to wrap this in a try-catch block?

assets/src/styles/blocks/HubspotForm/HubspotFormsOverrides.scss
```diff
@@ -0,0 +1,67 @@
+@mixin hubspot-form {
```
Inwerpsel, line 1: Can't we include the `hbspt-frm` class without putting it in another class?

assets/src/blocks/HubspotForm/HubspotFormThankYou.js
```diff
@@ -0,0 +1,24 @@
...
+        <ShareButtons />
+      </div>
+      <div className='hubspot-form-thank-you__step'>
+        <div className='hubspot-form-thank-you__step-id'>3</div>
+        <h2 className='hubspot-form-thank-you__step-title'>Take more actions</h2>
+        <p className='hubspot-form-thank-you__step-description'>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus in nibh est. Integer tincidunt dolor nec mattis molestie.</p>
+      </div>
+    </section>
+  );
+}
```
Inwerpsel, line 24: Please check your IDE settings for blank lines at the end of files :pray:

assets/src/blocks/HubspotForm/HubspotFormContext.js
GP-Dan-Tovbein: Not on every render, this effect is invoked only when the wrapperRef change. In this case only once or twice but no more.

Inwerpsel: Ah sorry I misread the dependencies.

assets/src/editorIndex.js
GP-Dan-Tovbein: It was suggested by SonarCloud

assets/src/blocks/HubspotForm/HubspotFormContext.js
```diff
@@ -0,0 +1,64 @@
...
+  useEffect,
+  useRef,
+  useState,
+} from '@wordpress/element';
+
+const Context = createContext({});
+Context.displayName = 'HubspotFormContext';
+const { Provider, Consumer } = Context;
+
+const ContextProvider = ({
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: Moved to a hook. However, I left the Context only for taking and returning props instead of passing them through each children `Component`.

classes/blocks/class-hubspotform.php
```diff
@@ -0,0 +1,133 @@
...
+	/**
+	 * Register HubspotForm block.
+	 */
+	public function register_hubspotform_block() {
+		register_block_type(
+			self::get_full_block_name(),
+			[
+				'render_callback' => [ $this, 'front_end_rendered_fallback' ],
+				'attributes' => [
+					'block_title'	=> [
```
lithrel, line 43: You can fix your code style issues automatically with phpcs/phpcbf if you have a running instance on your machine:
- Enter the php container `make php-shell`
- Get into the plugin dir `cd public/wp-content/plugins/planet4-plugin-gutenberg-blocks/`
- `./vendor/bin/phpcs` will show you the issues
- `./vendor/bin/phpcbf` will apply fixes

Inwerpsel (commented):
Still had these old comments I hadn't submitted, I think they're still relevant.

assets/src/blocks/HubspotForm/useDynamicHeight.js
```diff
@@ -0,0 +1,34 @@
+import { useState, useEffect } from 'react';
+
+export const useDynamicHeight = (wrapperRef) => {
```
Inwerpsel, line 3: Can you explain the motivation for this hook? It's not obvious why this is needed, or if it's needed at all.

The current usage sets the height on an element, where the passed in `wrapperRef` is the only descendant of that element. Doesn't that mean the parent element already has the right height?

assets/src/editorIndex.js
Inwerpsel: I think it only tests PRs, so probably it's not historically available for previous commits.

assets/src/blocks/HubspotForm/HubspotFormContext.js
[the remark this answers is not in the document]
Inwerpsel: As discussed on Slack, we should avoid using Context here for now because there's no benefit to using it, it's used in a single place. Context only has benefits for props that get passed deeply, which shouldn't be the case for a block's attributes.

assets/src/blocks/HubspotForm/HubspotFormBlock.js
```diff
@@ -0,0 +1,159 @@
...
+    // edit: (props) => {
+    //   return (
+    //     <Fragment>
+    //       <HubspotFormEditor {...props} />
+    //     </Fragment>
+    //   )
+    // },
+
+    /**
+      The belowed function works correctly using innerBlocks.
```
Inwerpsel, line 96: As discussed on Slack, InnerBlocks doesn't allow us the behavior we want, because it only works well if the block that contains it is also saved. In our case the block needs to be rendered, or hydrated. Neither is compatible yet with InnerBlocks.

However we can achieve the same goal (using a block editor for the thank you message) in a much simpler way: By creating a separate block for the thank you message, which is hidden by default on the page, and then shown by the form submit handler

assets/src/blocks/HubspotForm/useDynamicHeight.js
GP-Dan-Tovbein: You're are right, I've removed it. 
This hook has been used before, now it's not necessary for the current implementation.

assets/src/blocks/HubspotForm/useCreateHubspotForm.js
```diff
@@ -0,0 +1,70 @@
+import { useState, useEffect } from '@wordpress/element';
+
+export const useCreateHubspotForm = (shortcode, targetRef) => {
+  const [ submitted, setSubmitted ] = useState(false);
+
+  const onHubspotMessageHandler = evt => {
+    if(evt.data.type === 'hsFormCallback') {
+      if(evt.data.eventName === 'onFormSubmit') {
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: This block has some tricky implementations. One is that the user can show the submitted message provided by the plugin or use a custom one. Then, we need to detect when the form is submitted.

I've changed these listeners to this [piece of code](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/641/files#diff-a4a59ac7034e42b21e90e1dafbe268a2677eb4be25699b685b1052ddb6a94800R15-R20).

assets/src/blocks/HubspotForm/HubspotFormEditor.js
```diff
@@ -0,0 +1,161 @@
...
+import { Sidebar } from './HubspotFormSidebar';
+import { useToAttribute } from './hooks/useToAttribute';
+import { useBackgroundImage } from './hooks/useBackgroundImage';
+import { getStyleFromClassName } from '../getStyleFromClassName';
+
+const { __ } = wp.i18n;
+
+export const HubspotFormEditor = ({
+  attributes: {
+    block_background_image_id,
```
Inwerpsel, line 13: If this is the main background for the block, we could just use `backgroundImage`.

These attributes can be camelCase, btw. I think the snake_case format in our code is a leftover from the shortcodes. WordPress itself nowadays uses camelCase for all block attributes.

assets/src/blocks/HubspotForm/HubspotFormEditor.js
```diff
@@ -0,0 +1,161 @@
...
+    form_text,
+    form_title,
+    hubspot_shortcode,
+    hubspot_thankyou_message,
+    enable_custom_hubspot_thankyou_message,
+    className,
+  },
+  setAttributes,
+}) => {
+  const wrapperRef = useRef(null);
```
Inwerpsel, line 30: This ref now seems unused, can probably be removed.

Inwerpsel (commented):
I had this other comment still pending about `InnerBlocks`, and 2 small new ones.

assets/src/blocks/HubspotForm/HubspotFormEditor.js
GP-Dan-Tovbein: It's used [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/641/files/02445295ca2c4f40f40f94e456928e42cccd957a#diff-74590d4ebb9c6d57f60a788911755629e1110b04dd2b87c5a18c4c963f26c261R83)

Inwerpsel: Well the reference is passed there, but nothing actually uses it, right?

GP-Dan-Tovbein: Yes, I see your point and you're completely right. It was used by a removed hook. Good catch!

GP-Dan-Tovbein:
@lithrel @Inwerpsel FYI, I've flagged to UAT Passed

assets/src/blocks/HubspotForm/HubspotFormSidebar.js
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  const onRemoveImageHandler = () => {
+    setAttributes({
+      blockBackgroundImageId: null,
+      blockBackgroundImageUrl: null,
+    });
+  }
+
+  const getImageOrButton = (openEvent) => {
```
mleray, line 44: Does this function name still makes sense? It doesn't seem to ever render an image 🤔

assets/src/blocks/HubspotForm/HubspotFormSidebar.js
```diff
@@ -0,0 +1,112 @@
+import { CheckboxControl, PanelBody, PanelRow, Button, TextControl } from '@wordpress/components';
+import { MediaUpload, MediaUploadCheck, InspectorControls } from '@wordpress/block-editor';
+import { URLInput } from '../../components/URLInput/URLInput';
+import { useToAttribute } from './hooks/useToAttribute';
+
+const { __ } = wp.i18n;
+
+const PanelRowWraper = ({ labelText, helpText, children }) => {
```
mleray, line 8: It's not a big deal, but there's a small typo here, it should be `Wrapper` with 2 p's 🙂

classes/blocks/class-hubspotform.php
```diff
@@ -0,0 +1,114 @@
...
+					'enableCustomHubspotThankyouMessage' => [
+						'type'    => 'boolean',
+						'default' => true,
+					],
+				],
+			]
+		);
+
+		add_action( 'enqueue_block_editor_assets', [ self::class, 'enqueue_editor_assets' ] );
+		add_action(
```
mleray, line 89: I don't think that we need Hammer for this block? 🤔

webpack.config.js
```diff
@@ -66,6 +68,7 @@ const adminJsConfig = {
     TimelineEditorScript: './assets/src/blocks/Timeline/TimelineEditorScript.js',
     SocialMediaEditorScript: './assets/src/blocks/SocialMedia/SocialMediaEditorScript.js',
     GalleryEditorScript: './assets/src/blocks/Gallery/GalleryEditorScript.js',
+    HubspotFormEditorScript: './assets/src/blocks/HubspotForm/HubspotFormEditorScript.js',
```
mleray, line 71: Don't we also need to add the CSS files here in the config?

assets/src/blocks/HubspotForm/HubspotFormBlock.js
```diff
@@ -0,0 +1,110 @@
...
+      {
+        name: 'image-full-width',
+        label: getStyleLabel(
+          __('Image full width', 'planet4-blocks-backend'),
+          __('https://p4-designsystem.greenpeace.org/05f6e9516/p/213df0-hubspot-forms/b/99e047', 'planet4-blocks-backend'),
+        ),
+        isDefault: true,
+      },
+    ],
+    edit: (props) => <HubspotFormEditor {...props} />,
```
mleray, line 86: I believe this can be shortened to just `edit: HubspotFormEditor,` 🙂

assets/src/blocks/HubspotForm/hooks/useBackgroundImage.js
```diff
@@ -0,0 +1,13 @@
+import { useState, useEffect } from '@wordpress/element';
+
+export const useBackgroundImage = (image) => {
+  const [ backgroundImage, setBackgroundImage ] = useState();
+
+  useEffect(() => {
+    setBackgroundImage({ backgroundImage: image ? `url(${image})` : 'none' });
+  }, [ image ]);
+
+  return [
```
mleray, line 10: Why do we return an array here? 🤔 It seems it would be simpler to just return the value

assets/src/blocks/HubspotForm/HubspotFormFrontend.js
```diff
@@ -0,0 +1,70 @@
...
+    if(blockStyle) {
+      setStyleClass(blockStyle);
+    }
+  }, [ blockStyle ]);
+
+  return (
+    <section className={`hubspot-form block-wide ${styleClass}`} style={{...backgroundImage}}>
+      <div className='container'>
+        <div className='block-wrapper'>
+          <div className='block-wrapper-inner block-wrapper-text' style={{...backgroundImage}}>
```
mleray, line 36: Is it necessary to use the backgroundImage twice (here and on the `section` element above)?

assets/src/blocks/HubspotForm/hooks/useToAttribute.js
```diff
@@ -0,0 +1,25 @@
...
+        }
+      } else {
+        setAttributes({
+          [attributeName]: value,
+        });
+      }
+    }
+  };
+
+  return [
```
mleray, line 22: Same comment here as earlier, why do we return an array?

webpack.config.js
GP-Dan-Tovbein: What's the main reason to add it also in the config?

mleray: We do it for other blocks in the [cssConfig](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/79bc8812ac1b9cfa4e12ef77c5b288c0d489efb5/webpack.config.js#L74), I thought it was needed to properly split up the blocks' code

GP-Dan-Tovbein: Gotcha! I'll move it.

Just to know, what's the difference between adding to Webpack.config instead of [blocks.scss](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/641/files#diff-cbca468115dd29ef21fbf296256aaf045b3c5fa6d03d09caa7a41ff222f315d4R12)?

mleray: I might be wrong, but I think that when adding them there we only load the styles when the block is actually present on the page, instead of always loading them - so it's better for performance. But don't quote me on that 😬 @Inwerpsel probably knows better!

assets/src/blocks/HubspotForm/HubspotFormFrontend.js
GP-Dan-Tovbein: Yes, since its layout is different between large and small screens.
Large screens -> The background image is set to the main wrapper (including block text and block form)
Smaller screens -> The background image is only set to the block text.

Have a look the [design](https://p4-designsystem.greenpeace.org/05f6e9516/p/213df0-hubspot-forms/b/99e047)

assets/src/blocks/HubspotForm/HubspotFormBlock.js
GP-Dan-Tovbein: Yes, you're right. This is because I thought to potentially render the thank you message editor through this method.

webpack.config.js
Inwerpsel: Entirely correct @mleray

assets/src/blocks/HubspotForm/hooks/useHubspotForm.js
```diff
@@ -0,0 +1,66 @@
...
+      onFormSubmitted: ($form) => {
+        setSubmitted(true);
+        if($form.length) {
+          setSubmittedMessage($form[0].innerText);
+        }
+      }
+    });
+  }
+
+  const fetchHubspotFormsLibrary = () => {
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: @Inwerpsel unfortunately your suggestion is not possible since this method is calling another function `createHubspotForm` which is also inside the hook and is on charge of modifying `submitted` and `submittedMessage` hook state variables and this is why I added inside the hook.

assets/src/blocks/HubspotForm/HubspotFormEditor.js
```diff
@@ -0,0 +1,147 @@
...
+import { useBackgroundImage } from './hooks/useBackgroundImage';
+import { getStyleFromClassName } from '../getStyleFromClassName';
+
+const { __ } = wp.i18n;
+
+/**
+ * This component is created as a workaround because the <RichText> doesn't provide
+ * The functionality to set max characters like any content editable element.
+ */
+const MaxLengthHelperComponent = ({ value, maxLength, children, darkTheme = false }) => (
```
mleray, line 14: We'll need this functionality in other blocks as well, so maybe we could make it into a separate component already? 🤔 Ofc actually using it in other blocks should be taken care of in another ticket

assets/src/blocks/HubspotForm/HubspotFormEditor.js
```diff
@@ -0,0 +1,147 @@
...
+      <Sidebar {...{
+        ctaLink,
+        ctaNewTab,
+        blockBackgroundImageId,
+        enableCustomHubspotThankyouMessage,
+        hubspotThankyouMessage,
+        setAttributes
+      }} />
+      <section className={`hubspot-form hubspot-form-editor block-wide ${blockStyle}`} style={{...backgroundImage}}>
+        <div className='container'>
```
mleray, line 65: You might not need to add the `container` class to the block anymore, since we recently merged changes removing that class from blocks and adding it to the page content element instead 🤔

assets/src/styles/blocks/HubspotForm/HubspotFormEditorStyles.scss
```diff
@@ -0,0 +1,81 @@
...
+@import "../../master-theme/assets/src/scss/base/colors";
+@import "../../master-theme/assets/src/scss/base/mixins";
+
+@import "HubspotFormSidebarStyles";
+
+.hubspot-form-editor {
+  padding-right: 32px;
+  padding-left: 32px;
+
+  & > .container {
```
mleray, line 11: If you do end up removing the `container` class from the block, you won't need this anymore

assets/src/styles/blocks/HubspotForm/HubspotFormStyles.scss
```diff
@@ -0,0 +1,179 @@
...
+  background-position: center center;
+  box-shadow: inset 0 0 0 100vw transparentize($grey-80, .45);
+
+  &.block-wide {
+    width: 100vw;
+    /* A better way is to add these changes into the WideBlocks stylesheet */
+    margin-inline-start: calc(((100vw - 100%) / 2) * -1);
+  }
+
+  & > .container {
```
mleray, line 24: Same comment here as in the other css file, about removing the `container` class 🙂

assets/src/blocks/HubspotForm/HubspotFormEditor.js
```diff
@@ -0,0 +1,147 @@
...
+          <div className='block-wrapper'>
+            <div className='block-wrapper-inner block-wrapper-text' style={{...backgroundImage}}>
+              <div className='container'>
+                <RichText
```
mleray, line 69: Shouldn't the title also have a max length? 🤔

mleray (changes requested):
I've tested the block on the test instance, and I've noticed a couple of things:
- When the sidebar is open, the block preview goes out of the screen
Screenshot 2021-12-08 at 14 00 52

- I find the default text for the custom "Thank you" message a bit strange, I feel like the placeholder alone would be better but ofc the design team should be consulted on that matter 🙂 
Screenshot 2021-12-08 at 14 01 40
Screenshot 2021-12-08 at 14 01 49

- When there's no background image, it just shows a black background but we actually do have an [ImagePlaceholderIcon component](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/components/ImagePlaceholderIcon.js) that we could use instead? It would be more consistent with other blocks such as the CarouselHeader. Also, it's a small thing but the "Remove" button should be disabled when there is no background image selected, as it might be confusing otherwise

assets/src/blocks/HubspotForm/HubspotFormEditor.js
GP-Dan-Tovbein: I completely agree with this but I suggest to merge this implementation as a base one and then continue working on improvements in the future.

mleray: I see your point, I'm just worried that we'll forget this component is defined here when we start implementing this for other blocks... Anyway, it's not a blocker, so I'll leave it up to you 🙂

Inwerpsel: I agree it should probably be a separate component, but let's do that in a follow up so we don't have to rush it.

assets/src/blocks/HubspotForm/HubspotFormEditor.js
Inwerpsel: Agree, but if it requires a lot of CSS changes we can also do it in a follow up.

assets/src/blocks/HubspotForm/hooks/useToAttribute.js
```diff
@@ -0,0 +1,23 @@
...
+  const toAttribute = (attributeName, maxLength = -1) => value => {
+    if(setAttributes) {
+      if(maxLength > -1) {
+        if(value.length < maxLength) {
+          setAttributes({
+            [attributeName]: value,
+          });
+        } else {
+          setAttributes({
+            [attributeName]: value.slice(0, maxLength),
```
Inwerpsel, line 11: We probably don't want to enforce it this way.

What if someone copy pastes text into the block from somewhere? Also if you would start changing a word in the middle of your sentence, it will disappear at the end and won't come back if you delete the extra characters again.

If the text is too long, we should show them a warning, and optionally block the post from saving until they fixed it. We can do this using validation on attributes, [like for slides on the Carousel Header block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/9dd382fc04caa03d56f62b4038e21f9dac3d3cd1/assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js#L26-L35).

But it's also something we can do in a follow up. Definitely should be fixed before using the block in production.

assets/src/blocks/HubspotForm/HubspotFormEditor.js
mleray: Indeed, it's not a blocker!

assets/src/blocks/HubspotForm/hooks/useBackgroundImage.js
```diff
@@ -0,0 +1,11 @@
+import { useState, useEffect } from '@wordpress/element';
+
+export const useBackgroundImage = (image) => {
+  const [ backgroundImage, setBackgroundImage ] = useState();
+
+  useEffect(() => {
+    setBackgroundImage({ backgroundImage: image ? `url(${image})` : 'none' });
```
Inwerpsel, line 7: You could replace this whole function with just this:
```
export const useBackgroundImage = (image) => {
  return { backgroundImage: image ? `url(${image})` : 'none' };
}
```
and it would work just as well, only the first time it wouldn't return `null` for the background.

String interpolation is a very cheap operation. Much cheaper than the overhead the `useState` and `useEffect` hooks add.

Using an effect would only make sense here if you called an API to get the specific background each time, or do some heavy calculation.

Inwerpsel (approved):
Left a few more comments and suggestions but nothing that blocks us from merging it as a beta block.

assets/src/blocks/HubspotForm/HubspotFormEditor.js
comzeradd: @GP-Dan-Tovbein can you make a follow up ticket to start capturing all the things we need to work on afterwards in order to put the block out of beta?

comzeradd:
@GP-Dan-Tovbein rebasing will also fix the failing Acceptance tests

assets/src/blocks/HubspotForm/HubspotFormEditor.js
GP-Dan-Tovbein: For sure @comzeradd

GP-Dan-Tovbein:
I've created this [ticket](https://jira.greenpeace.org/browse/PLANET-6504) to follow up all pending stuffs of coming iterations.

c.c. @mleray feel free to "approve" your requested changes that will allow me to merge this 1st iteration.

mleray (approved):
Looks good enough to be released as beta 👍 However the commits probably need to be squashed before merge? 🤔

PR #643, opened 2021-08-16
Some more blocks variables (ENForm and Counter), bump theme editor

Inwerpsel:
---

This is the last batch of variables needed for themes. I also bumped the theme editor package to a newer version that fixes bugs and adds custom controls for many CSS properties.

All CSS changes are only adding variables and not changing the structure, so these changes should be safe. I changed one variable name (to use the shorthand), which I will update in the single theme (and page) that uses it.

### DIFF
--- a/assets/src/styles/blocks/Counter.scss
+++ b/assets/src/styles/blocks/Counter.scss
@@ -1,6 +1,8 @@
 .counter-block {
+  _-- {
+    font-family: $roboto;
+  }
   text-align: center;
-  font-family: $roboto;
   margin-top: $space-xl;
   margin-bottom: $space-xl;
 
@@ -30,16 +32,18 @@
   }
 
   .counter-target {
+    _-- {
+      color: inherit;
+      font-size: 2rem;
+      font-weight: bold;
+      margin-bottom: $space-lg;
+    }
     display: inline-block;
-
     line-height: 1;
-    font-size: 2rem;
-    font-weight: bold;
-    margin-bottom: $space-lg;
   }
 
   &.counter-style-1 {
-    .counter-target {
+    .counter-target _-- {
       font-size: 3rem;
       font-weight: bold;
       margin-bottom: $space-lg;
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2021.08.16 09:03:09

assets/src/styles/blocks/OldENForm/components/_enform.scss
```diff
@@ -90,7 +90,7 @@
   }
 }
 
-.form-description p:first-child --block-enform--caption--paragraph--first-- {
+.form-description p:first-child _-- {
```
Inwerpsel, line 93: Changing this back to short notation as the only purpose is to ensure continuity for the "climate" theme. We can remove this rule afterwards, as what it's effectively doing is turning the first paragraph into a title, while the block also has a title.

assets/src/styles/blocks/Counter.scss
```diff
@@ -30,59 +32,70 @@
...
+      height: 30px;
+      margin-bottom: $space-lg;
+    }
     overflow: hidden;
   }
 
   .progress-bar {
-    background-color: $dark-blue;
+    _-- {
+      background-color: $dark-blue;
```
mleray, line 65: You probably want to use `background` here instead of `background-color` 🙂

assets/src/styles/blocks/Counter.scss
```diff
@@ -30,59 +32,70 @@
...
+      transform: skewX(-30deg);
+    }
     height: 100%;
-    transform: skew(-30deg, 0deg);
     margin-left: -10px;
   }
 
-  .enform-progress-bar {
+  .enform-progress-bar _-- {
```
mleray, line 72: Same here, you could use `background`

assets/src/styles/blocks/OldENForm/components/_enform.scss
```diff
@@ -171,7 +173,6 @@
 
       p {
         font-size: .75rem;
-        font-family: $roboto !important;
```
mleray, line 174: The `!important` here was not needed after all? 🤔

assets/src/styles/blocks/Counter.scss
Inwerpsel: Nice catch, forgot to change this one. Gradients could work well on the progress bar :ok_hand:

assets/src/styles/blocks/OldENForm/components/_enform.scss
Inwerpsel: I think I tested if it did something on the live content before removing it, but let me re-check :eyes:

Inwerpsel: It already uses Roboto because of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/902e5e4322269414a2def75e6399cd679168b61a/assets/src/styles/blocks/OldENForm/components/_enform.scss#L101-L103

mleray (approved):
Looks good to me 👍

PR #645, opened 2021-08-18
PLANET-6328 WIP Make Submenu block saved in HTML

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6328

* It only needs to change when the post content changes. Add a
subscriber to the block's edit component, so that it auto updates the
content when other blocks change. As a result it doesn't require any
render on the front end anymore.
* Fix a bug with anchors in classic HTML blocks. Previously it was using
the entire classic block's HTML to generate an anchor. Not only could
that be huge, the same string was used for each anchor found in the
classic block, resulting in React complaining about items having the
same `key`.
* Missing piece to the puzzle: generating anchors. This could be tricky,
as it involves updating all blocks that have no anchor at some point to
have one. Challenge is that we want to update all headings that were
auto-generated, for example a user starts typing a word into a new
heading block. The anchor should update as long as the user didn't
specify an anchor themselves.

Perhaps we could add a button to generate all missing anchors on the block. This gets around the challenges of auto updating those whenever a new heading is added. We could add a block validation so that a message is shown to generate missing headers on the submenu block. The button can suggest to manually add those anchors as well.

### DIFF
--- a/assets/src/blocks/Submenu/SubmenuBlock.js
+++ b/assets/src/blocks/Submenu/SubmenuBlock.js
@@ -1,6 +1,7 @@
 import { SubmenuEditor } from './SubmenuEditor.js';
 import { Tooltip } from '@wordpress/components';
 import {example} from './example';
+import { SubmenuFrontend } from './SubmenuFrontend';
 
 const { __ } = wp.i18n;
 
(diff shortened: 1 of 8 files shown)


planet-4:

:new_moon: [jupiter](test-jupiter) | [admin](https://www-dev.greenpeace.org/test-jupiter/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-jupiter/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-jupiter) | [composer-local.json](https://github.com/greenpeace/planet4-test-jupiter/blob/main/composer-local.json)

:watch: 2021.08.18 10:51:21

PR #646, opened 2021-08-19
PLANET-5608 Implement new Take Action cover designs

mleray:
### Description

I've implemented the new designs only for the WYSIWYG version of the block, by adding `-new` to its classes which we'll be able to remove once we remove it from from beta blocks. That way existing blocks should not be affected, and we can merge these changes despite not being ready to merge the related changes for the post page designs.

See https://jira.greenpeace.org/browse/PLANET-5608

### Testing

I've set up a [page](https://www-dev.greenpeace.org/test-iocaste/new-take-action-covers/) for UAT, but you can also test it on local by creating a new "Covers (beta)" block and selecting the "Take Action covers" style. You can also check the non-WYSIWYG Covers block to make sure that it isn't affected by the changes!

### DIFF
--- a/assets/src/blocks/Covers/TakeActionCovers.js
+++ b/assets/src/blocks/Covers/TakeActionCovers.js
@@ -27,7 +27,7 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
             tags,
             image,
             excerpt,
-            button_text
+            button_text,
           } = cover;
           const hideCover = !!initialRowsLimit && index >= row * amountPerRow;
 
@@ -36,30 +36,28 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
           }
 
           return (
-            <div key={title} className='col-lg-4 col-md-6 cover-card-column'>
-              <div className='cover-card card-one' style={{ backgroundImage: `url(${image})` }}>
+            <div key={title} className='col-lg-4 col-md-6'>
+              <div className='cover-card-new'>
                 <a
                   className='cover-card-overlay'
+                  data-ga-category='Take Action Covers'
+                  data-ga-action='Card'
+                  data-ga-label='n/a'
+                  href={button_link}
+                  aria-label={__('Take action cover, link to ' + title, 'planet4-blocks')}
+                />
+                <a
                   data-ga-category='Take Action Covers'
                   data-ga-action='Image'
                   data-ga-label='n/a'
                   href={button_link}
                   aria-label={__('Take action cover, link to ' + title, 'planet4-blocks')}
-                />
+                >
+                  <img src={image} />
+                </a>
                 <div className='cover-card-content'>
-                  {tags && tags.map(tag => (
-                    <a
-                      key={tag.name}
-                      className='cover-card-tag'
-                      data-ga-category='Take Action Covers'
-                      data-ga-action='Navigation Tag'
-                      data-ga-label='n/a'
-                      href={tag.href}
-                    >
-                      <span aria-label='hashtag'>#</span>
-                      {tag.name}
-                    </a>
-                  ))}
+                  {/* Regardless of how many tags there are, we only show the first one */}
+                  {tags && tags.length > 0 && <span className='cover-card-tag'>{tags[0].name}</span>}
                   <a
                     className='cover-card-heading'
                     data-ga-category='Take Action Covers'
@@ -69,10 +67,10 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
                   >
                     {title}
                   </a>
-                  <p>{excerpt}</p>
+                  <p className="cover-card-excerpt">{excerpt}</p>
                 </div>
                 <a
-                  className='btn btn-action btn-block cover-card-btn'
+                  className='btn cover-card-btn btn-primary'
                   data-ga-category='Take Action Covers'
                   data-ga-action='Call to Action'
                   data-ga-label='n/a'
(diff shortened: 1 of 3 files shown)


planet-4:

:new_moon: [iocaste](test-iocaste) | [admin](https://www-dev.greenpeace.org/test-iocaste/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-iocaste/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-iocaste) | [composer-local.json](https://github.com/greenpeace/planet4-test-iocaste/blob/main/composer-local.json)

:watch: 2021.08.19 12:00:20

assets/src/blocks/Covers/TakeActionCovers.js
```diff
@@ -36,30 +36,39 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
           }
 
           return (
-            <div key={title} className='col-lg-4 col-md-6 cover-card-column'>
```
mleray, line 39: Was not needed anymore 🙂

assets/src/blocks/Covers/TakeActionCovers.js
```diff
@@ -36,30 +36,39 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
...
                 <a
                   className='cover-card-overlay'
                   data-ga-category='Take Action Covers'
-                  data-ga-action='Image'
+                  data-ga-action='Card'
                   data-ga-label='n/a'
                   href={button_link}
                   aria-label={__('Take action cover, link to ' + title, 'planet4-blocks')}
                 />
+                <a
```
mleray, line 49: After discussing with Julia we decided to separate click events on the card from the ones on the image, since now the image is no longer the background and therefore not full height anymore

assets/src/styles/blocks/OldCovers/TakeActionCovers.scss
```diff
@@ -223,56 +223,56 @@
...
   }
-}
-
-.cover-card-tag {
-  --block-covers--card-tag-- {
-    color: $yellow;
 
-    &:hover {
-      text-decoration: underline;
+  .cover-card-tag {
```
mleray, line 227: I've had to move these selectors inside the `.cover-card` one, otherwise they were still applied to the new styles even though they shouldn't

assets/src/styles/blocks/Covers/layouts/TakeActionCovers.scss
```diff
@@ -95,186 +59,96 @@
...
+    width: auto;
 
-    @include large-and-up {
-      font-size: 1.5rem;
-      margin-top: 8px;
-      max-width: 100%;
+    img {
+      height: 220px;
+      max-height: 220px;
+      min-height: 220px;
```
GP-Dan-Tovbein, line 131: Why do you need to set these attributes together?

assets/src/blocks/Covers/TakeActionCovers.js
```diff
@@ -36,30 +36,39 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
...
+                  data-ga-label='n/a'
+                  href={button_link}
+                  aria-label={__('Take action cover, link to ' + title, 'planet4-blocks')}
+                >
+                  <img src={image} />
+                </a>
                 <div className='cover-card-content'>
-                  {tags && tags.map(tag => (
+                  {/* Regardless of how many tags there are, we only show the first one */}
+                  {tags && tags.length > 0 && (
```
GP-Dan-Tovbein, line 60: `> 0` is not necessary in this case. `tags.length` is ok.

mleray: I thought about that, but then the block shows `0` in place of the tag :/

Screenshot 2021-08-24 at 09 50 45

assets/src/styles/blocks/Covers/layouts/TakeActionCovers.scss
mleray: Hmmm yeah it's not ideal, it's because otherwise the `height` attribute gets overridden by the value set in [_page.scss](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/pages/_page.scss#L34). This could also be solved by using `!important` but I wanted to avoid that :/

assets/src/blocks/Covers/TakeActionCovers.js
GP-Dan-Tovbein: This is because you are using the logical operator (`&&`) and returns true|1 or false|0`. When you have an empty Array it will return `0`.
Try replacing to 
```
{tags && tags.length > 0 ? (your code) : null}
```

GP-Dan-Tovbein: Updated comment.

mleray: Hmmm yeah I see, not sure that's better than using `> 0` though? It's longer that way and I find it less clear 🤔

GP-Dan-Tovbein:
Yep, I agree that it's more clear using `> 0` but it's really the same!

assets/src/blocks/Covers/TakeActionCovers.js
GP-Dan-Tovbein: Yep, I agree that it's more clear using `> 0` but it's really the same!

GP-Dan-Tovbein:
Code looks good to me!

GP-Dan-Tovbein (approved):
Code looks good to me!

PR #647, opened 2021-08-20
PLANET-6362 Add new version of themes

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6362

This PR does the following:

- Add refactored versions of themes as options of the existing dropdown, and suffix the old theme with `(old)` to remove confusion.
- Exclude these themes from the new themes dropdown (only relevant for test instances which might have a theme with the same name in the options).
- Remove some console logs

To test:
You can switch all content using the old theme to the new version and it should look identical, except for minor spacing changes or over-use of capitalization that was not preserved. Needs to be tested together with https://github.com/greenpeace/planet4-master-theme/pull/1435, else the themes won't exist.

I deployed the latest version to GPI, where I'm currently switching all campaign pages (that use a theme) to the new version. https://www-dev.greenpeace.org/international/?s=&orderby=_score&f[ctype][Campaign]=4 You can compare these to production on https://www.greenpeace.org/international/?s=&orderby=_score&f[ctype][Campaign]=4

Later this day I'll deploy the other dev sites that have campaign pages with a theme.

I did not preserve the over-use of `text-transform: uppercase` in many cases, because it's bad for accessibility. I'll try to confirm with someone whether this is a change we can make.

Keep in mind that deploying this won't automatically switch to the new themes. We'll do this one by one after deploy. So this PR doesn't have to be blocked by minor differences we notice in the themes. I'll try to reach out to authors of the pages to check if these differences should be taken up. I already did this for the Plastic campaign, and almost all differences with the old theme were perceived as improvements.

### DIFF
--- a/assets/src/components/Sidebar/NewThemeSettings.js
+++ b/assets/src/components/Sidebar/NewThemeSettings.js
@@ -6,6 +6,8 @@ import { useDispatch } from '@wordpress/data';
 
 const keysAsLabel = obj => Object.keys(obj).map(k => ({ label: k, value: k }));
 
+const refactoredThemes = ['climate-new', 'forest-new', 'oceans-new', 'plastic-new'];
+
 // Just a lightweight way to have a separate UI element in the sidebar so that it's clear which themes are legacy and
 // which are new. This is likely just for internal usage to facilitate reviewing the refactored version, eventually
 // editors will work with a single select.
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2021.08.20 08:31:25

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -56,12 +81,17 @@ const useAppliedCssVariables = (serverThemes, currentTheme) => {
 export const NewThemeSettings = ({ onChange, currentTheme }) => {
   const [selectedTheme, setSelectedTheme] = useState(currentTheme);
   const {editPost} = useDispatch('core/editor');
-  const serverThemes = useServerThemes();
+  const allServerThemes = useServerThemes();
+  const serverThemes = Object.entries(allServerThemes).reduce((a, [k, t]) => {
+    if (refactoredThemes.includes(k)) {
```
GP-Dan-Tovbein, line 86: This could be another approach
`return (refactoredThemes.includes(k)) ? a : ({ ...a, [k]: t })`

GP-Dan-Tovbein (approved):
Code looks good to me

assets/src/components/Sidebar/NewThemeSettings.js
Inwerpsel: @GP-Dan-Tovbein I also moved the function out and made the params more readable.

PR #627, opened 2021-07-21
PLANET-6122 Use hydration in CarouselHeader WYSIWYG

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6122

PR #648, opened 2021-08-24
Get hydration to work

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6122

Use `ReactDOM.renderToString` in the `save` function of the block, because otherwise it causes hook violations in the editor. This is probably something Gutenberg can fix by using a similar encapsulated render when calling `save`.

For non-migrated content, add a regular render which gets the new attributes from PHP.

After we have deployed the move to the new version, we need to go through all pages with the block, press the button to migrate slide image data, and save the post. This is manageable because we have just around 100 usages of the block.

This was created a PR against https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/627/, however I suggest we just merge this one for simplicity.


planet-4:

:new_moon: [leda](test-leda) | [admin](https://www-dev.greenpeace.org/test-leda/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-leda/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-leda) | [composer-local.json](https://github.com/greenpeace/planet4-test-leda/blob/main/composer-local.json)

:watch: 2021.08.24 10:24:04

classes/blocks/class-carouselheader.php
```diff
@@ -127,15 +127,17 @@ public function prepare_data( $fields ): array {
...
 		if ( ! empty( $attributes['slides'] ) && empty( $attributes['slides'][0]['image_url'] ) ) {
 			$attributes['slides'] = self::get_slides_image_data( $attributes['slides'] );
 		}
 
 		$json = wp_json_encode( [ 'attributes' => $attributes ] );
 
 		// When we remove "beta" from the block name, we'll be able to use as_hydratable_block from class-base-block.php.
-		return '<div data-hydrate="planet4-blocks/carousel-header-beta" data-attributes="' . htmlspecialchars( $json ) . '">'
-		. trim( $content )
+		return '<div data-render="planet4-blocks/carousel-header-beta" data-attributes="' . htmlspecialchars( $json ) . '">'
```
Inwerpsel, line 140: This might be good enough as a temporary solution. Rendering front end would temporarily have worse performance than the hydrated version, so we would want to migrate content after deploy. But it's probably still better performance than the old version.

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
```diff
@@ -86,7 +108,7 @@ export const CarouselHeaderFrontend = ({ slides, carousel_autoplay, className })
               active={currentSlide == index}
               ref={element => slidesRef ? slidesRef.current[index] = element : null}
             >
-              <SlideBackground slide={slide} />
+              <SlideBackground slide={pageLoaded ? slide : slides[0]} />
```
Inwerpsel, line 111: I added this as a quick way to avoid loading all images. Still need to investigate if it makes sense to do or there are better alternatives.

assets/src/functions/hydrateBlock.js
```diff
@@ -1,8 +1,10 @@
 export const hydrateBlock = (blockName, Component) => {
-  document.querySelectorAll( `[data-hydrate="${blockName}"]` ).forEach(
+  const blocks = document.querySelectorAll( `[data-hydrate="${blockName}"]` );
+  console.log(blocks);
+  blocks.forEach(
     blockNode => {
       const attributes = JSON.parse( blockNode.dataset.attributes );
-      ReactDOM.hydrate( <Component { ...attributes.attributes } />, blockNode );
+      ReactDOM.hydrate( <Component { ...attributes } />, blockNode );
```
Inwerpsel, line 7: I removed the extra `attributes` level here. In the code to render front end components it was only still using it because it's hard to change because of block content validation in the editor.

assets/src/blocks/CarouselHeader/EditableBackground.js
```diff
@@ -1,11 +1,13 @@
 import { MediaUpload, MediaUploadCheck } from '@wordpress/block-editor';
 import { ImagePlaceholder } from './ImagePlaceholder';
 import { Button, Dropdown } from '@wordpress/components';
+import { toSrcSet } from '../../../../../../themes/planet4-master-theme/assets/src/js/Components/archivePicker/sizeFunctions';
```
mleray, line 4: Isn't it a bit weird to store this function in the `archivePicker` folder? They're not related 🤔

Inwerpsel: Yes, still need to move it.

Inwerpsel: I ended up copying the function, as I just overlooked that it was not even in the same repo. It's a simple function and we need to make adjustments because the data for `select('core').getMedia(id)` uses different keys than what is returned in `<MediaUpload onSelect/>` :facepalm:

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
```diff
@@ -7,7 +7,21 @@ import { StaticCaption } from './StaticCaption';
 
 const isRTL = document.querySelector('html').dir === 'rtl';
 
+const usePageLoaded = () => {
+  const [pageLoaded, setPageLoaded] = useState(false);
```
mleray, line 11: Why do we need a separate hook here? Couldn't we just write this inside `CarouselHeaderFrontend` directly?

Inwerpsel: What if we want to use it in another component?

assets/src/blocks/CarouselHeader/SlideBackground.js
```diff
@@ -1,12 +1,13 @@
 export const SlideBackground = ({ slide }) => {
-  const { image_url, image_sizes, image_srcset, focal_points } = slide;
+  const { image_url, image_sizes, image_srcset, focal_points, image_alt } = slide;
   return (
     <div className='background-holder'>
       <img
         src={image_url}
         style={{ objectPosition: `${focal_points?.x * 100}% ${focal_points?.y * 100}%` }}
         srcSet={image_srcset}
         sizes={image_sizes || 'false'}
```
mleray, line 9: I think this line can be removed?

assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
Inwerpsel: But also to reduce complexity. `CarouselHeaderFrontend` doesn't care about anything else but that the page was loaded. So we don't need to add the complexity of the inner workings (`useState` and `useEffect`) in that place.

assets/src/blocks/CarouselHeader/CarouselHeaderEditor.js
```diff
@@ -50,30 +63,57 @@ export const CarouselHeaderEditor = ({ setAttributes, attributes }) => {
     goToSlide(currentSlide > lastSlide ? 0 : currentSlide, true);
   }
 
+  const needsMigration = slides.some(slide => !!slide.image && !slide.image_srcset)
+  const migratedSlides = useSelect(select=> slides && slides.map(slide => {
+    if (!needsMigration) {
```
Inwerpsel, line 68: If it doesn't need migration, it will map to a bunch of nulls. That's not a problem because in that case the result is not used.

mleray (approved):
Tested on local, looks good to me 👍 However we should clean up the commits a bit before merging 🙂

Inwerpsel:
> Tested on local, looks good to me  However we should clean up the commits a bit before merging

Rebasing would be a bit tricky unless we squash everything, since the last commits undo some things of the earlier commits. But then we would lose separate meaningful commits. Are you ok with merging like this?

mleray:
> > Tested on local, looks good to me  However we should clean up the commits a bit before merging
> 
> Rebasing would be a bit tricky unless we squash everything, since the last commits undo some things of the earlier commits. But then we would lose separate meaningful commits. Are you ok with merging like this?

@Inwerpsel sure, go for it!

PR #650, opened 2021-08-27
Switch Carousel Header to WYSIWYG version

Inwerpsel:
In the first commit I tried to preserve the old version for safety reasons in case the new version wouldn't work in some cases. It wasn't working yet, and I deployed this branch to develop sites to test the migration. It worked well in all cases, so we can just remove the old block.

After releasing this we need to go through all sites and manually migrate the block (by pressing the "Migrate image data" button). It will temporarily front end render the block if the page was requested before we did this migration.


sagarsdeshmukh (approved):
PR tested on local, LGTM 👍

PR #651, opened 2021-08-27
Fixed the infinite loading on some site's main page (by Nowill Design)

hit1205:
Fixed the infinite loading on some site's main page

User click tags in the tag cloud will make cookies to save the selection, this cause infinite loading animation in shortcake_big_carousel_manual_selection block on some themes, because p4_vars is undefined. These themes are:
* gpea-planet8-child-theme
* planet4-child-theme-hongkong
* planet4-child-theme-korea
* planet4-child-theme-taiwan

This commit updated the first param (script handle name) in wp_localize_script() to the same as planet4-plugin-engagingnetworks's lastest version.


classes/blocks/class-enform.php
```diff
@@ -156,13 +156,13 @@ public static function enqueue_frontend_assets() {
 		parent::enqueue_frontend_assets();
 
 		wp_localize_script(
-			'plugin-engagingnetworks',
+			'plugin-en',
```
Inwerpsel, line 159: Does this mean that the site is currently running both plugins? We moved the functionality of `planet4-plugin-engagingnetworks` into this plugin. Is it possible for you to disable the old en-plugin and see if that also resolves the issue?

hit1205: We cannot to disable `planet4-plugin-engagingnetworks` individually in these sites, because there is another plugin `planet4-gpea-plugin-blocks` depend on it and enabled. The infinite loading is also happend in the block in `planet4-gpea-plugin-blocks` because some JavaScript code in `gpea-planet8-child-theme` (and other three themes) depend on the var `p4_vars` be defined here.

But we have decide to modify the `gpea-planet8-child-theme` to remove the code depend on `p4_vars`, so we will close the pull request.

PR #652, opened 2021-08-27
Deprecate frontendRendered function

Inwerpsel:
This function makes no more sense to use in new code because of issues described in the doc comment. We should convert all previous usages to saved HTML, as it will solve many issues.

The scale of the impact of this became obvious while working on the Carousel Header WYISWYG version. It's using `hydrate` now, however for the migration it has a fallback to `frontendRendered` (to give us time between deploy and migration which creates the static HTML).

This allows us to compare the performance of both techniques. 

The version using `hydrate` does 5 to 10 points better than the old version.

The fallback using `frontendRendered` does **5 to 10 points worse than the old version**, even though it still includes various other performance fixes over the old version. See starting row 945 on https://docs.google.com/spreadsheets/d/1Jv08gzCHEH6QRn03iIc3eykop_5pnk02VAvnGvS9o_Y/edit#gid=0 _(In fact it does almost 5 points worse without GTM than the old version does with GTM. I didn't compare the equal case, hence I extrapolate it should be 5 to 10 points under same conditions)_.

Same page content, same carousel. You can clearly see the render difference if you refresh the page. Even with a regular refresh and no throttling, where all page resources come from cache, it's still very noticeable.
Rendered from scratch: https://www-dev.greenpeace.org/test-phoebe/campaign/covid-19-coronavirus/
Hydrated: https://www-dev.greenpeace.org/test-phoebe/campaign/covid-19-coronavirus-hydrate/

It basically throws our entire performance budget out of the window to start with, and that's before any third party scripts are added by our users.

Furthermore it likely aggravates third party impact, since React is adding DOM nodes at the same moment third party scripts start firing.

Image lazy loading also seems to not work at all, as all page speed test results list all images on the page, while it should only be getting the ones that are above the fold. Likely because the page is very short before our blocks render, so the browser thinks they're all above the fold.

I expect that when we convert all blocks that use it to hydrated markup that we'll see an increase 20 points or more across all sites.


lithrel (approved):
:+1:

PR #654, opened 2021-09-02
Remove characterLimit prop that only results in console warnings

Inwerpsel:
* For some reason this was added to many RichText components, even
though they don't have this prop. So the only effect it had was cause a
whole lot of warnings flooding the console.
* This was copied from a component that was using the
`withCharacterCounter` HOC, that adds the counter and warnings around a
component. Don't use that component, HOCs are legacy, but we still need
to rewrite it. If we need to add a character counter in some places, we
should first rewrite the HOC.

Ref: No ticket, it only prevents warnings and removes non-functioning code.

### DIFF
--- a/assets/src/blocks/Accordion/AccordionEditor.js
+++ b/assets/src/blocks/Accordion/AccordionEditor.js
@@ -46,7 +46,6 @@ const renderView = ({ title, description, tabs, className }, setAttributes, isSe
           onChange={toAttribute('title')}
           keepPlaceholderOnFocus={true}
           withoutInteractiveFormatting
-          characterLimit={60}
           multiline="false"
           allowedFormats={[]}
         />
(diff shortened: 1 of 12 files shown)


planet-4:

:new_moon: [tavros](test-tavros) | [admin](https://www-dev.greenpeace.org/test-tavros/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-tavros/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-tavros) | [composer-local.json](https://github.com/greenpeace/planet4-test-tavros/blob/main/composer-local.json)

:watch: 2021.09.02 10:28:39

lithrel (approved):
:+1:

PR #655, opened 2021-09-03
PLANET-6137 Allow re-ordering slides

Inwerpsel:
Add first implementation of re-ordering to Carousel Header Slides. See commit message and code comments for detailed explanation.
Ref: https://jira.greenpeace.org/browse/PLANET-6137

### DIFF
--- a/assets/src/blocks/CarouselHeader/Sidebar.js
+++ b/assets/src/blocks/CarouselHeader/Sidebar.js
@@ -3,11 +3,29 @@ import {
   PanelBody,
   PanelRow,
 } from '@wordpress/components';
+import { Fragment } from '@wordpress/element';
 
 import { InspectorControls } from '@wordpress/block-editor';
 import { SidebarSlide } from './SidebarSlide';
+import { useState, useRef } from 'react';
 
 const { __ } = wp.i18n;
+
+const reOrderSlides = (slides, oldIndex, newIndex) => {
+  const movingSlide = slides[oldIndex];
+  return slides.reduce((orderedSlides, slide, index) => {
+    if (index === newIndex) {
+      orderedSlides.push(movingSlide);
+    }
+    if (index !== oldIndex) {
+      orderedSlides.push(slide);
+    }
+    if (index === slides.length -1 && newIndex > index) {
+      orderedSlides.push(movingSlide);
+    }
+    return orderedSlides;
+  }, []);
+}
 
 export const Sidebar = ({
   carouselAutoplay,
@@ -16,8 +34,12 @@ export const Sidebar = ({
   currentSlide,
   changeSlideAttribute,
   goToSlide,
-}) => (
-  <InspectorControls>
+}) => {
+  const [dragTarget, setDragTarget] = useState(null);
+  const [draggedSlide, setDraggedSlide] = useState(null);
+  const timeoutRef = useRef(null);
+
+  return <InspectorControls>
     <PanelBody title={__('Setting', 'planet4-blocks-backend')}>
       <CheckboxControl
         label={__('Carousel Autoplay', 'planet4-blocks-backend')}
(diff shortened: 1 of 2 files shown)


planet-4:

:new_moon: [titan](test-titan) | [admin](https://www-dev.greenpeace.org/test-titan/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-titan/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-titan) | [composer-local.json](https://github.com/greenpeace/planet4-test-titan/blob/main/composer-local.json)

:watch: 2021.09.03 11:20:34

assets/src/blocks/CarouselHeader/Sidebar.js
```diff
@@ -3,50 +3,117 @@ import {
...
+        help={ __('Select to trigger images autoslide', 'planet4-blocks-backend') }
+        value={ carouselAutoplay }
+        checked={ carouselAutoplay === true }
+        onChange={ value => setAttributes({ carousel_autoplay: value }) }
       />
     </PanelBody>
-    {slides.map((slide, index) => (
+    { slides.map((slide, index) => <div
+      style={ {
+        transform: index === draggedSlide ? 'scaleY(0)' : 'none',
```
Inwerpsel, line 54: When you start dragging an element, what is shown while dragging is a bitmap snapshot of what the element looked like the moment the drag started. The `.01s` delay should be plenty to ensure that snapshot is taken before we hide the element.

assets/src/blocks/CarouselHeader/Sidebar.js
```diff
@@ -47,6 +114,11 @@ export const Sidebar = ({
...
         </PanelRow>
       </PanelBody>
-    ))}
-  </InspectorControls>
-);
+    </div>) }
+    { draggedSlide !== null && draggedSlide !== slides.length - 1 && <div
+      className={slides.length === dragTarget ? 'carousel-sidebar-insert-before' : ''}
+      onDragOver={ () => setDragTarget(slides.length) }
+      onDragLeave={ () => setDragTarget(null)}
```
GP-Dan-Tovbein, line 121: Please move the lambda function outside the render method. Why? Because in this case you're creating a new functions when the render is called. If you move it outside the render and you'll create it just once


```
const onDragLeaveHandler = () => setDragTarget(nul)

return <InspectorControls>
....
onDragLeave={ onDragLeaveHandler }
</InspectorControls>
```

assets/src/blocks/CarouselHeader/Sidebar.js
```diff
@@ -27,13 +49,58 @@ export const Sidebar = ({
...
+        setDragTarget(null);
+      } }
+    >
       <PanelBody
-        key={index}
-        title={`${__('Slide', 'planet4-blocks-backend')} ${index + 1}`}
-        opened={currentSlide === index}
-        onToggle={() => {
-          if (currentSlide !== index) {
+        key={ index }
```
GP-Dan-Tovbein, line 88: Could you change this key value to another one? Because the `index` value is not recommended to use as `key` through item from lists

assets/src/blocks/CarouselHeader/Sidebar.js
Inwerpsel: @GP-Dan-Tovbein If we add this `const` like you suggest, right before the return statement, the function would be created an equal amount of times.

assets/src/blocks/CarouselHeader/Sidebar.js
Inwerpsel: Good point :+1: This is actually more a problem now slides are re-orderable.

Inwerpsel: Actually, now that it's wrapped in a div it doesn't need a key anymore. I also noticed in some other code it has a key that is not needed, I'll remove those now.

assets/src/blocks/CarouselHeader/Sidebar.js
Inwerpsel: And again, please consider the magnitude of the resources and time it takes to create a function vs. the amount of times this render will be called. You can see in [these lines](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/892b10537e4cddb6082ce0163e8e388e2d24adfb/assets/src/blocks/CarouselHeader/Sidebar.js#L31-L39) all the conditions that will trigger another render. The render will only get called when the attributes change, or when you start dragging/enter or leave a drop target.

Inwerpsel: So we're talking about reducing the legibility of code, so that we can have on the order of 1 part in a billion faster code (supposedly), for something that is only called at max a few times per second, and which already takes a negligible amount of time.

GP-Dan-Tovbein (approved):
Code looks good to me!

PR #656, opened 2021-09-06
Enable Hubspot form block

comzeradd:
This enables Hubspot plugin Gutenberg block that simplifies adding a form.

It doesn't seem to create any issue when the plugin is missing.


mleray (changes requested):
Is there a reason why we don't want to allow the forms in posts? 🤔 Also there are some indentation issues that need to be fixed 🙂

comzeradd:
> Is there a reason why we don't want to allow the forms in posts?

I added Posts too. I thought it usually on pages NROs add petitions, but let's enable it and see how people use it.

mleray (approved):
Looks good to me 👍

PR #660, opened 2021-09-07
Allow theming counter block spacing and display

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6362

Add a few variables needed to style the counter block in ENForm style. These are used in https://github.com/greenpeace/planet4-master-theme/pull/1441/files for the "climate" theme.

### DIFF
--- a/assets/src/styles/blocks/Counter.scss
+++ b/assets/src/styles/blocks/Counter.scss
@@ -23,6 +23,9 @@
     }
 
     .progress-container {
+      _-- {
+        display: block;
+      }
       height: 20px;
     }
   }
--- a/assets/src/styles/blocks/ENForm/components/_enform-side-style.scss
+++ b/assets/src/styles/blocks/ENForm/components/_enform-side-style.scss
@@ -119,7 +119,7 @@
         padding-top: 0;
       }
 
-      .counter-block {
+      .counter-block _-- {
         margin-bottom: 0;
         margin-top: 32px;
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2021.09.07 11:59:28

assets/src/styles/blocks/Counter.scss
```diff
@@ -23,6 +23,9 @@
     }
 
     .progress-container {
+      _-- {
```
Inwerpsel, line 26: Even though the style is called "progress bar in en form", the climate theme hides the progress bar. The result is somewhat non-sensical, and ideally I suppose the block should allow to have only text and still be put in the ENForm. But I'm keeping it to this quick fix, because we probably want to more broadly reconsider the current behavior of the counter block traveling into the ENForm block on page load. Instead these could use inner blocks in the new version.

assets/src/styles/blocks/ENForm/components/_enform-side-style.scss
```diff
@@ -119,7 +119,7 @@
         padding-top: 0;
       }
 
-      .counter-block {
+      .counter-block _-- {
```
Inwerpsel, line 122: In general I tried avoiding making spacing variable, but in this case it looked way off with our default spacing.

mleray (approved):
Looks good to me 👍

PR #665, opened 2021-09-09
Set theme variables on page load

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6363

---

Previously the logic to set the theme variables was only triggered when the sidebar is open. With this change it's always run on page load because the `CampaignSidebar` component is always created when loading the editor.

Unfortunately the sidebar code still needs cleaning up, so I had to squeeze it into some already complex code. The sidebar needs to be entirely rewritten anyway, because it still uses class components and does some weird stuff, so I guess this can be improved at that point.

To test: With this change you should see the theme applied in the editor when you open a page, regardless of which sidebar is open. There could be a small delay before you see the styles applied.

### DIFF
--- a/assets/src/components/Sidebar/CampaignSidebar.js
+++ b/assets/src/components/Sidebar/CampaignSidebar.js
@@ -6,7 +6,7 @@ import isShallowEqual from '@wordpress/is-shallow-equal';
 import { savePreviewMeta } from '../../saveMetaToPreview';
 import { PostParentLink } from './PostParentLink';
 import { LegacyThemeSettings } from './LegacyThemeSettings';
-import { NewThemeSettings } from './NewThemeSettings';
+import { applyChangesToDom, NewThemeSettings, themeJsonUrl } from './NewThemeSettings';
 
 const isLegacy= theme => [
   'default',
@@ -111,6 +111,15 @@ export class CampaignSidebar extends Component {
       ) {
         const theme = await loadTheme(themeName);
         this.setState({ theme: theme });
+        if (themeName) {
+          try {
+            const response = await fetch(`${ themeJsonUrl + themeName.replace('-new', '') }.json`);
+            const theme = await response.json();
+            applyChangesToDom(theme, []);
+          } catch (e) {
+            console.log(`Failed loading config for ${ themeName }`);
+          }
+        }
       }
     } );
     wp.data.subscribe( () => {
--- a/assets/src/components/Sidebar/NewThemeSettings.js
+++ b/assets/src/components/Sidebar/NewThemeSettings.js
@@ -28,7 +28,7 @@ const getAllDefinedProps = () => Object.values(document.documentElement.style).f
   return 'string' === typeof k && k.match(/^--/);
 });
 const baseUrl = window.location.href.split( '/wp-admin' )[ 0 ];
-const themeJsonUrl = `${ baseUrl }/wp-content/themes/planet4-master-theme/themes/`;
+export const themeJsonUrl = `${ baseUrl }/wp-content/themes/planet4-master-theme/themes/`;
 
 const collectTheme = async (a, t) => {
   const response = await fetch(`${ themeJsonUrl + t.replace('-new', '') }.json`);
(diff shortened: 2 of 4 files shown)


assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -111,6 +111,15 @@ export class CampaignSidebar extends Component {
       ) {
         const theme = await loadTheme(themeName);
         this.setState({ theme: theme });
+        if (themeName) {
+          try {
+            const response = await fetch(`${ themeJsonUrl + themeName.replace('-new', '') }.json`);
```
Inwerpsel, line 116: We'll probably migrate content to a name without `-new` now that these are available again. After that we can clean up this logic, and some similar other ones throughout the code.

planet-4:

:new_moon: [phoebe](test-phoebe) | [admin](https://www-dev.greenpeace.org/test-phoebe/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-phoebe/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-phoebe) | [composer-local.json](https://github.com/greenpeace/planet4-test-phoebe/blob/main/composer-local.json)

:watch: 2021.09.09 15:31:29

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -51,31 +51,37 @@ const useJsonThemes = () => {
...
+  });
+
+  if (initialVars.length === 0) {
+    return;
+  }
+
+  const customProps = getAllDefinedProps();
+
+  customProps.forEach(k => {
+    if (!Object.keys(theme).includes(k) && !initialVars.includes(k)) {
```
GP-Dan-Tovbein, line 71: What do you think to change to this?
`if (![...Object.keys(theme), ....initialVars].includes(k))`
In fact, you don't want to have `k` in any of those `Arrays`.

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -51,31 +51,37 @@ const useJsonThemes = () => {
...
+  if (!theme) {
+    return;
+  }
+  Object.entries(theme).forEach(([name, value]) => {
+    // This will only work reliably if no other code is adding new custom properties to the root element after this
+    // component is first rendered. This should be the case in the post editor.
+    document.documentElement.style.setProperty(name, value);
+  });
+
+  if (initialVars.length === 0) {
```
GP-Dan-Tovbein, line 64: Another option could be 
`if (!initialVars.length) {`

assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -111,6 +111,15 @@ export class CampaignSidebar extends Component {
...
+            const theme = await response.json();
+            applyChangesToDom(theme, []);
+          } catch (e) {
+            console.log('Failed loading config for ' + themeName);
```
GP-Dan-Tovbein, line 120: Another approach could be
`console.log('Failed loading config for %s', themeName);`

GP-Dan-Tovbein (approved):
I left some comments, but just suggestions.

assets/src/components/Sidebar/NewThemeSettings.js
Inwerpsel: For now I just moved existing logic, so I'd rather not touch it, but I'll definitely look into that deeper when we address the sidebar class component. It will also involve changing `NewThemeSettings`..

assets/src/styles/editorFonts.scss
```diff
@@ -0,0 +1,42 @@
+@font-face {
```
Inwerpsel, line 1: This includes the contents of [this file](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/commit/1158181f2e7845c4e12403c2beb1b97e56ab8060#diff-6a303c3d9ced23f325fae852cca1af4dc97cf4c2166cb9d3aa97f5fb6e1c1190) and the next one.

GP-Dan-Tovbein (approved):
It's still looking good to me, after the last commit!

PR #667, opened 2021-09-13
Node renderer

Inwerpsel:
Ref: 

---

Trying to create a Cloudflare worker that is able to generate initial HTML with the right version of the code.

To simplify things I didn't set up all build config, and instead made minor adjustments to the code so that it can run without those dependencies.

You currently can run it using 
```
node worker/dist/index.js covers [[attributes]]
```

I included the [built standalone script](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/3971513c4ddcd7375b0ee5e3485fed09ebc78b42/worker/dist/index.js), it's currently at 44.2KB, which includes React, ReactDOM and our code (need to check if there's other deps in it). That's still far away from the allowed 1MB for Cloudflare workers. If we keep front end and editor code apart, it should stay well below that.

Still very WIP, todo:
- create actual Cloudflare worker script
- correctly set up build config (I'll look at https://github.com/flareact/flareact)
- `srcSet` is outputted in the HTML, instead of `srcset`. I assumed React handled that by default, but it probably needs some config. Same for other camel cased properties.
- For Covers it required only minor changes (might be handled by config later), not sure how other blocks would look like.


planet-4:

:new_moon: [oberon](test-oberon) | [admin](https://www-dev.greenpeace.org/test-oberon/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-oberon/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-oberon) | [composer-local.json](https://github.com/greenpeace/planet4-test-oberon/blob/main/composer-local.json)

:watch: 2021.09.13 17:54:54

classes/blocks/class-covers.php
```diff
@@ -420,21 +420,25 @@ private static function populate_posts_for_cfc( &$fields ) {
 
 		if ( ! empty( $posts ) ) {
 
-			foreach ( $posts as $post ) {
+			foreach ( $posts as $post_data ) {
```
Inwerpsel, line 423: Did a quick workaround to avoid `post_content` and more data from being included in the attributes.

It's technically not needed but it made the attributes very big too work with.

Inwerpsel:
This was an idea to address the issue of server side rendered blocks having the same template as the front end.

In last years WordPress moved towards small component blocks doing one specific thing (e.g. "Post Featured Image"). As a result each block just has a small amount of template code. This makes having a JS template for the editor and a PHP template for the front end very manageable. So a back end JS rendered doesn't really make things simpler anymore.

PR #671, opened 2021-09-16
Use hooks instead of HOCs for theme sidebar

Inwerpsel:
Replace HOCs (Higher Order Components) with regular components and hooks. For the conversion I preserved the exact behavior it previously had, so there are still some oddities and things to reconsider.

I also fixed an issue with how the themes are applied in the editor.

I didn't implement a replacement for the `withDefaultLabel` HOC, because that can be handled by including this in the label in the theme config file when it's generated. I might add these to the text of the existing themes.

To test: There are no intended changes in behavior, except for the default labels. Unfortunately we don't have a detailed specification of how it should work, and the live implementation does have some weird aspects we might want to change. Reach out to me if you run into any behavior that seems off.


assets/src/components/Sidebar/CampaignSidebar.js
```diff
@@ -85,7 +85,7 @@ export class CampaignSidebar extends Component {
 
     // Loop through the new theme's fields, and check whether any of the already chosen options has a value that is not
     // available anymore.
-    const invalidatedFields = prevOptions.fields.filter(field => gotInvalidated(field, options, meta));
+    const invalidatedFields = prevOptions?.fields.filter(field => gotInvalidated(field, options, meta)) || [];
```
Inwerpsel, line 88: This was probably broken in my previous PR.

planet-4:

:new_moon: [umbriel](test-umbriel) | [admin](https://www-dev.greenpeace.org/test-umbriel/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-umbriel/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-umbriel) | [composer-local.json](https://github.com/greenpeace/planet4-test-umbriel/blob/main/composer-local.json)

:watch: 2021.09.16 12:32:45

assets/src/components/Sidebar/NewThemeSettings.js
```diff
@@ -61,10 +61,6 @@ export const applyChangesToDom = (theme, initialVars) => {
     document.documentElement.style.setProperty(name, value);
   });
 
-  if (initialVars.length === 0) {
```
Inwerpsel, line 64: I'm not sure why I did this earlier, but it broke switching back to the default theme, so probably a mistake.

assets/src/components/Sidebar/LegacyThemeSettings.js
```diff
@@ -30,108 +44,111 @@ const themeOptions = [
...
     </div>
     <PanelBody
       title={ __( "Navigation", 'planet4-blocks-backend' ) }
       initialOpen={ true }
     >
-      <Radio
-        metaKey='campaign_nav_type'
-        theme={ theme }
+      <RadioControl
+        selected={getValue(LEGACY_FIELDS.navigationType)}
```
Inwerpsel, line 85: Even though the new code is more redundant, you can now easily read the following:
- Which component is actually used.
- Which props are set on that component (was hidden inside of 2 different HOCs previously)
- What happens when you change it.

assets/src/components/Sidebar/LegacyThemeSettings.js
```diff
@@ -30,108 +44,111 @@ const themeOptions = [
...
   const {
     handleThemeSwitch,
     theme,
   } = props;
 
+  const meta = useSelect(select => select('core/editor').getEditedPostAttribute('meta'),[]);
+  const { editPost } = useDispatch('core/editor');
+
+  useEffect(() => {
+    handleThemeSwitch('theme', meta.theme, meta)
```
Inwerpsel, line 57: For now I kept these parameters. I'll simplify it when converting `CampaignSidebar` to a function component.

mleray (approved):
Looks good to me 👍

PR #677, opened 2021-09-20
Prevent rare crash where block is null

Inwerpsel:
* The editor crashed on this code for me because it received `null` for
the block. It happened on WordPress 5.8, but that could be coincidence.
* I was undoing multiple edits involving a Query Loop block and can't
remember which particular one triggered the crash. Haven't managed to
reproduce the issue.

Ref: No ticket

### DIFF
--- a/assets/src/BlockEditorValidation.js
+++ b/assets/src/BlockEditorValidation.js
@@ -32,9 +32,8 @@ export const blockEditorValidation = () => {
 
         return results;
       }, [] );
-      if (results.length > 0) {
-        invalidBlocks.push(...results)
-      }
+
+      invalidBlocks.push(...results);
 
       return invalidBlocks;
     }, []);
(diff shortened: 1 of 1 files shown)


planet-4:

:new_moon: [uranus](test-uranus) | [admin](https://www-dev.greenpeace.org/test-uranus/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-uranus/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-uranus) | [composer-local.json](https://github.com/greenpeace/planet4-test-uranus/blob/main/composer-local.json)

:watch: 2021.09.20 14:40:53

assets/src/BlockEditorValidation.js
```diff
@@ -26,16 +32,16 @@ export const blockEditorValidation = () => {
 
         return results;
       }, [] );
-      if (results.length > 0) {
-        invalidBlocks.push(...results)
-      }
+
+      invalidBlocks.push(...results);
```
Inwerpsel, line 36: I didn't intend to make this change but it was a compromise with Sonarcloud to get back down to the allowed complexity...

Inwerpsel: Sonar is completely right btw that this function is too complex, but I'm avoiding the context switch atm to dive into this logic. I might address it later.

lithrel (approved):

PR #687, opened 2021-10-01
Spreadsheet now sorts numerically

victoria-miltcheva:
Ref: https://github.com/greenpeace/planet4/issues/165

---

## Summary
This PR adds numerical sorting to the spreadsheet table.

## Notes
- I passed in `undefined` as the locale parameter to `localeCompare`, this way the default locale will be used.
- I also passed in an `options` object with `numeric: true` so that it will recognize numbers when sorting and `sensitivity: base` so that the sorting takes case sensitive strings into account. 

[Google spreadsheet used for testing](https://docs.google.com/spreadsheets/d/e/2PACX-1vQMXcRFT6xJBNzZsl_ZKh4YWmYuc0ab3m5uOMb2j5tNxa4HEO-TgHxhE27X3DVJlm7MLqseECbx4YTV/pubhtml)

### Default spreadsheet (no sorting)
default

### Numerical ascending
numerical_asc

### Numerical descending
numerical_desc

### Alphabetical ascending
alphabetical_asc

### Alphabetical descending
alphabetical_desc


Inwerpsel:
Thanks @victoria-miltcheva , the changes look good to me. Looks like with the `numeric: true` argument it both sorts numeric values as expected and otherwise sorts text the same way. :100: 

Can you remove the first 2 commits? As the second one undoes the first one. The easiest would be to use `git rebase -i master`, and then put a `d` (short for `drop`) before the first 2 commits.
Screenshot from 2021-10-07 10-37-18

victoria-miltcheva:
Hi @Inwerpsel , thank you for reviewing my PR. Yeah, the sorting fix ended up being very straightforward.

I've rebased my branch by the way. Thanks for the suggestion.

victoria-miltcheva:
Update: I added a body to my commit so that the commit linting status check will pass.

PR #692, opened 2021-10-11
Add linting for remaining JS, but only on changed files of the branch

Inwerpsel:
Linting previously only ran on `public/js` and `tests/js`.

Because of the [large amount of violations](https://app.circleci.com/pipelines/github/greenpeace/planet4-plugin-gutenberg-blocks/5138/workflows/e641939e-3e2b-45e7-847c-8e31c680c179/jobs/21705), I opted to not auto fix them. There would still be hundreds of violations to fix manually, and the auto fix in many cases changed most of the file. Auto fix possibly does not do the right thing in some cases because we still need to properly setup `.eslintrc.json`.

Instead I changed the lint job so that it **only runs against files that were changed** compared to the common ancestor with the `master` branch.

Before I noticed there were too many violations, I used a glob syntax which I thought matched all JS files, but only did so at a specific directory level, and didn't detect the majority of our blocks' code. That's why the PR does include fixes for those few files, as it seemed manageable to fix everything at that point.

This means that each time we'll change a file, we'll have to fix the violations in the same PR, or first fix the file in a separate PR. In some cases we actually need to properly configure `eslintrc.json` instead, like with parsing errors.

**Example of changing a file with a parsing error**: https://app.circleci.com/pipelines/github/greenpeace/planet4-plugin-gutenberg-blocks/5147/workflows/c703f269-946d-4486-b746-aa7e42f38679/jobs/21771

That file probably has more issues, but the parsing error prevents them from being detected. This particular error is because we use a plugin for optional chaining (`mayNotExist`**`?.`**`value`), and it takes some investigation to find the right config for us so `eslint` can deal with it.

Testing on https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/tree/test-lint-only-changed

Ref: 

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -218,7 +218,12 @@ jobs:
       - run:
           name: Lint JS
           when: always
-          command: eslint --ext .js public/js tests/js assets/src
+          # For now lint only changed files to avoid bulk fixing.
+          command: |
+            changed_files=$(git diff --name-only master...HEAD public/js tests/js assets/src | grep "\.js$")
+            echo "Changed files:"
+            echo $changed_files | tr " " "\n"
+            eslint --ext .js $(echo "$changed_files" | xargs)
       - store_test_results:
           path: /tmp/test-results
 


lithrel:
We have different rules on [master-theme](https://github.com/greenpeace/planet4-master-theme/blob/master/.eslintrc.json) and [plugin](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/.eslintrc.json), maybe we should merge them

PR #701, opened 2021-10-25
PLANET-6069 Take action boxout new version

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6069

---

Switch Take Action Boxout to the "bottom"* style.

_*_ We'll only initially show the block at the bottom, however the block can be placed anywhere in the post.

The fixed style on mobile will be for a follow up to unblock the post design.

Also fixes the following issues:
- Act page image not working in the editor.
- Srcset missing on the front end results in a oversized image.
- Remove unused AB test CSS that we nevertheless served to everyone.
- Remove duplicated CSS from Covers block that has code splitting set up, so it doesn't make sense to include other block's CSS in that file.


assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -162,33 +163,29 @@ export const TakeActionBoxoutEditor = ({
             label={__('Select Take Action Page:', 'planet4-blocks-backend')}
             value={take_action_page}
             options={[
-              { label: __('None', 'planet4-blocks-backend'), value: 0 },
+              { label: __('None (custom)', 'planet4-blocks-backend'), value: 0 },
```
Inwerpsel, line 166: I'll check this with design, but it seems more sensible to show specifically what this option does, more than just not having an act page selected. That way we can also hide the customize label entirely if an act page is selected, which further makes the block less confusing.

mleray (dismissed):
~Could you please remove the `keepPlaceholderOnFocus` prop from the editor file, since it's deprecated now? 🙂~

It's actually already removed on master so rebasing should solve it 👍

Inwerpsel:
@mleray I'll rebase this to a single commit, because the branch that has the new Covers designs has some conflicts. I already did that on a branch, also to check if both together play well https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/717

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutFrontend.js
```diff
@@ -11,12 +11,7 @@ export const TakeActionBoxoutFrontend = ({
...
-  <section
-    className={`cover-card action-card ${className || ''}`}
-    style={{
-      backgroundImage: `linear-gradient(180deg, rgba(0, 0, 0, 0.3), rgba(0, 0, 0, 0)), url(${imageUrl})`
-    }}
-  >
+  <section className={ `boxout ${ className || '' }` }>
     <a
       data-ga-category="Take Action Boxout"
       data-ga-action="Image"
```
mleray, line 17: Maybe something for a follow-up ticket, but the analytics will need to be updated: here for instance they no longer make sense since the image is no longer full-width

PR #710, opened 2021-11-04
Fix display and focus of Articles image link

Inwerpsel:
ATTENTION: The focus state on the [test instance](https://www-dev.greenpeace.org/test-pandora/) still causes layout jumps, which is being addressed in https://github.com/greenpeace/planet4-master-theme/pull/1500/files

* `d-flex` is not needed and broke the display of the link around it.
* Ensure focus around image nicely uses the borders. Without block there
was a protrusion caused by the link being wider but narrower than the
image it contains.

### DIFF
--- a/assets/src/styles/blocks/Articles.scss
+++ b/assets/src/styles/blocks/Articles.scss
@@ -57,6 +57,10 @@
 .article-list-item-image {
   order: 2;
 
+  a {
+    display: block;
+  }
+
   img {
     width: 100% !important;
     object-fit: cover;
--- a/assets/src/blocks/Articles/ArticlePreview.js
+++ b/assets/src/blocks/Articles/ArticlePreview.js
@@ -70,7 +70,7 @@ export class ArticlePreview extends Component {
         data-ga-action="Image"
         data-ga-label="n/a">
           <img
-            className="d-flex topicwise-article-image"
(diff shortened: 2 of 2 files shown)


assets/src/styles/blocks/Articles.scss
```diff
@@ -57,6 +57,10 @@
 .article-list-item-image {
   order: 2;
 
+  a {
```
Inwerpsel, line 60: Before I removed `d-flex`, the focus state of the image link was entirely invisible.

After removing, I had to add this rule. Otherwise the link was actually wider than the image it contains, causing a weird "protrusion" when the link is focused.
Screenshot from 2021-11-04 15-41-21

planet-4:

:watch: 2021.11.04 14:41:59

assets/src/styles/blocks/Articles.scss
Inwerpsel: That being said, this focused state doesn't play very well around image links. In many cases the contrast between the border and the edge of the image is not sufficient or almost 0.

It still seems better than no focus indication at all, so maybe we can proceed with these changes and improve on it later.

mleray (approved):
Tested on local together with the master-theme PR, looks good to me 👍

PR #715, opened 2021-11-09
PLANET-6379 New Posts & Covers & Take Action Boxout block designs

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-6379

This holds already reviewed code merged into this feature branch. Still we should check whether everything integrated nicely.

This currently fails acceptance tests because it requires a change in master theme, which is [introduced in the corresponding PR for these changes](https://github.com/greenpeace/planet4-master-theme/pull/1454).


planet-4:

:watch: 2021.11.09 13:46:45

assets/src/styles/blocks/OldCovers/TakeActionBoxoutEditor.scss
```diff
@@ -1,20 +1,17 @@
...
-  html[dir="rtl"] & {
-    margin-left: 0;
-    margin-right: auto;
-  }
-
   .cover-card-excerpt,
   .cover-card-btn {
     display: block;
   }
+
```
mleray, line 12: Don't we want to remove everything from the `OldCovers` folder? 🤔

public/js/post_action.js
```diff
@@ -1,67 +0,0 @@
-// Force the Cover card to follow scroll
```
mleray, line 1: This file is now empty, I guess we can remove it?

assets/src/styles/blocks/OldCovers/TakeActionBoxoutEditor.scss
mleray: It also seems it's [not even used anymore](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/715/files#diff-1d8227442d5b198c6d5362b74a30d9a58a34c4de87df6535394bdddec936267cL23)

mleray (approved):
I've added a few comments but they're about cleaning up some old files, other than that looks good to me 👍

assets/src/styles/blocks/OldCovers/TakeActionBoxoutEditor.scss
Inwerpsel: Checking, it's actually supposed to be there, because we need the fix added below :thinking: Maybe merging/rebasing went wrong.

public/js/post_action.js
Inwerpsel: I had done that, but probably a rebase or a merge messed with that. Removed it now :+1:

assets/src/styles/blocks/OldCovers/TakeActionBoxoutEditor.scss
Inwerpsel: Ah it was a fix from before I had to change the class names, good catch. Moved it to the right directory and loaded it in the editor.

classes/blocks/class-covers.php
```diff
@@ -381,16 +381,21 @@ private static function populate_posts_for_act_pages( &$fields ) {
 				if ( is_array( $wp_tags ) && $wp_tags ) {
 					foreach ( $wp_tags as $wp_tag ) {
 						$tags[] = [
-							'name' => $wp_tag->name,
+							'name' => html_entity_decode( $wp_tag->name ),
```
Inwerpsel, line 384: Previously this data went to Twig, which simply included it as text in the HTML. However now we use it with React which would output the entity literally (e.g. `&amp;`)

assets/src/styles/blocks/TakeActionBoxout/edit.scss
```diff
@@ -0,0 +1,12 @@
+.boxout {
+  position: relative;
+
+  a {
+    pointer-events: none;
+  }
+
+  // Prevent line clamp on editable stuff, because it's very buggy.
+  [contenteditable] {
+    -webkit-line-clamp: unset !important;
```
Inwerpsel, line 10: `!important` seems justified here, because currently there is no way that it can work properly on `contenteditable` elements.

PR #726, opened 2021-11-16
PLANET-6475 Add mobile sticky behavior to Take Action Boxout

Inwerpsel:
### Description

See https://jira.greenpeace.org/browse/PLANET-6475

This PR also includes some changes in the "normal" TAB (not sticky) to reflect the latest designs and use the 8px grid.

### Testing

On local you can add Take Action Boxout blocks to a post and add the new "sticky behaviour" setting to one of them. Then you can test in the frontend that it behaves as expected, and that it fits the designs in all screen sizes. Otherwise you can test it on the [page I made for UAT](https://www-dev.greenpeace.org/test-pluto/story/47211/test-tab/) 🙂 You should also make sure that only one boxout per post can have the "sticky behaviour" attribute turned on!


planet-4:

:watch: 2021.11.16 11:45:08

assets/src/styles/blocks/TakeActionBoxout/style.scss
```diff
@@ -152,6 +110,74 @@
...
+    .btn {
+      margin-inline-end: 16px;
+      margin-bottom: 16px;
+    }
+  }
+}
+
+@media screen and (max-width: $large-width) {
+  .sticky-bottom-mobile {
+    z-index: 9001;
```
GP-Dan-Tovbein, line 146: I wonder why `9001`?

mleray: I actually don't know either 😬 @Inwerpsel wrote that line

Inwerpsel: Just a random high value, you can change it :sweat_smile:

GP-Dan-Tovbein: Gotcha! 
For me is the same too, but as a convention I would prefer change to something like `z-index: 9999`. it will seem more familiar for developers and we would avoid doubts (like mine 😆).

GP-Dan-Tovbein (commented):
I've tested this functionality and works well, good job! However, and as a personal opinion I don't agree how the take action boxout behaves when is closed. I think it should be move back to the page, like the other boxouts, instead of disappearing. What if I regret to close?

What do you think?

Inwerpsel:
@GP-Dan-Tovbein You're right that we need to think through the closing behavior a bit more. I'm not sure about placing the block inline instead. It could be weird to a user that they want to close something and then see it appearing on the page. It can also make the scroll height jump which can be annoying.

mleray:
@GP-Dan-Tovbein @Inwerpsel maybe we could release this version as a first iteration and then update the closing behaviour if needed after discussing with the rest of the team?

comzeradd:
I agree with @mleray. Since these are the initial design specs, we can iterate after getting some feedback either from the team or the community. Not exactly the same scope, but even the Cookie box is not easy to make it appear again once you dismiss it.

GP-Dan-Tovbein (approved):
Code looks good!

GP-Dan-Tovbein:
Approved as a first iteration, good job @mleray !

PR #729, opened 2021-11-17
Fine tune Carousel Header slide re-ordering design and UX

GP-Dan-Tovbein:
:point_right: [Demo page](https://www-dev.greenpeace.org/test-atlas/demo-page-fine-tune-carousel-header-slide-re-ordering-design-and-ux/)
[Figma](https://www.figma.com/file/IqISN3iLC17g2B2xxy0Clt/Carousel?node-id=301%3A2)

Ref: https://jira.greenpeace.org/browse/PLANET-6376

The current implementation is not working as expected. Not only when user editor wants to drag and drop any slide but also when he wants to edit any other value whenever the Slide is opened.

**What is included**
- Implement the ability to re order the list clicking on the up and down arrow button (**new feature**)
- Modify and fix the drag and drop logic since It has not been working correctly. In addition, someone has reported this issue through this [Slack message](https://greenpeace.slack.com/archives/C0151L0KKNX/p1637874331098400).
- Add ellipsis to texts
- Overrides some functionalities of the PanelBody component. Allowing to click on up and down arrow and stopped the propagation of the [onToggle PanelBody's method](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/panel/body.js#L44-L49).
- Fix the Image focal point logic when the Slide is opened
- Move Slide styles to a separate file for a more specificity
- Implement style logic when the Slide is opened.

### DIFF
--- a/assets/src/blocks/CarouselHeader/Sidebar.js
+++ b/assets/src/blocks/CarouselHeader/Sidebar.js
@@ -1,43 +1,153 @@
+import { useState, useRef, useEffect } from 'react';
 import {
   CheckboxControl,
   PanelBody,
-  PanelRow,
 } from '@wordpress/components';
-import { Fragment } from '@wordpress/element';
-
 import { InspectorControls } from '@wordpress/block-editor';
 import { SidebarSlide } from './SidebarSlide';
-import { useState, useRef } from 'react';
 
 const { __ } = wp.i18n;
 
-const reOrderSlides = (slides, oldIndex, newIndex) => {
-  const movingSlide = slides[oldIndex];
-  return slides.reduce((orderedSlides, slide, index) => {
-    if (index === newIndex) {
-      orderedSlides.push(movingSlide);
-    }
-    if (index !== oldIndex) {
-      orderedSlides.push(slide);
-    }
-    if (index === slides.length -1 && newIndex > index) {
-      orderedSlides.push(movingSlide);
-    }
-    return orderedSlides;
-  }, []);
-}
+const getNode = (index) => (
+  document.querySelector(`.sidebar-slide[data-index="${index}"]`)
+)
 
 export const Sidebar = ({
   carouselAutoplay,
   setAttributes,
   slides,
-  currentSlide,
   changeSlideAttribute,
   goToSlide,
 }) => {
   const [dragTarget, setDragTarget] = useState(null);
   const [draggedSlide, setDraggedSlide] = useState(null);
-  const timeoutRef = useRef(null);
+  const slidesRef = useRef(null);
+
+  const upOrDownHandler = ( evt ) => {
+    evt.stopPropagation();
+
+    if(!slidesRef.current) {
+      return;
+    }
+
+    const currentIndex = parseInt(evt.currentTarget.dataset.index)
+    const currentNode = getNode(currentIndex);
+    let siblingNode, siblingIndex = -1;
+
+    switch(evt.currentTarget.dataset.type) {
+      case 'up':
+        siblingIndex = currentIndex - 1;
+      break;
+      case 'down':
+        siblingIndex = currentIndex + 1;
+      break;
+    }
+
+    siblingNode = getNode(siblingIndex);
+
+    // Get the data-index from sidebar-slide
+    // Set to position absolute and find it sibbling element
+    const currentNodeCloned = currentNode.cloneNode(true);
+    const siblingNodeCloned = siblingNode.cloneNode(true);
+
+    currentNodeCloned.classList.add('cloned-slide');
+    siblingNodeCloned.classList.add('cloned-slide');
+
+    // Append before setting the top
+    slidesRef.current.append(currentNodeCloned);
+    slidesRef.current.append(siblingNodeCloned);
+
+    currentNode.style.opacity = 0;
+    siblingNode.style.opacity = 0;
+
+    currentNodeCloned.style.top = `${currentNode.offsetTop}px`;
+    siblingNodeCloned.style.top = `${siblingNode.offsetTop}px`;
+
+    const timeout = setTimeout(() => {
+      // Switch positions
+      currentNodeCloned.style.top = `${siblingNodeCloned.offsetTop}px`;
+      siblingNodeCloned.style.top = `${currentNodeCloned.offsetTop}px`;
+
+      setTimeout(() => {
+        // Re order slides
+        setAttributes({ slides: slides.reduce((prev, curr, idx) => {
+          if(idx !== currentIndex && idx !== siblingIndex) {
+            if(prev) {
+              return prev.concat(curr)
+            }
+          } else {
+            if(idx === currentIndex) {
+              return prev.concat(slides[siblingIndex]);
+            }
+            if(idx === siblingIndex) {
+              return prev.concat(slides[currentIndex]);
+            }
+            return prev;
+          }
+        }, []) });
+
+        // Add a delay
+        setTimeout(() => {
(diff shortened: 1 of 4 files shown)


planet-4:

:watch: 2021.11.17 13:35:35

assets/src/blocks/CarouselHeader/Sidebar.js
```diff
@@ -49,72 +130,112 @@ export const Sidebar = ({
...
+          title={
+            <div className='slide-item-wrapper'>
+              <div
+                className='slide-item-drag-button'
+                onClick={(evt) => {
+                  evt.stopPropagation();
+                  console.log('Drag')
+                }}
+              >
+                <svg width="8" height="14" viewBox="0 0 8 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
GP-Dan-Tovbein, line 186: This `svg` code will be replaced with an external svg

assets/src/styles/blocks/CarouselHeader/SidebarSlidesEditor.scss
```diff
@@ -0,0 +1,129 @@
+// Doing this for a more specificity
+.sidebar-slides {
+  position: relative;
+
+  .sidebar-slide {
```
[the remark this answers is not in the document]
Inwerpsel: There's only a handful of CSS classes in this file that actually need to be nested (such as `.components-panel__body-toggle`). For all the other ones it's unnecessarily increasing CSS specificity.

You can put all these at the top level
```
sidebar-slide
slide-item-wrapper
slide-item-draggable-button
slide-item-order-arrows
arrow-button <-- definitely this one. It's clear to see that we could use this outside of slide re-ordering (but not if we hard code it to only work when used inside particular combinations of parent elements)
slide-item-image
slide-item-text
cloned-slide
opened-slide

File: assets/src/blocks/CarouselHeader/SidebarSlide.js [hunk 3]
```diff
@@ -23,45 +36,104 @@ export const SidebarSlide = ({
...
-          <div className='row'>
-            <div className='col'>
-              <CheckboxControl
-                label={__('Open in a new tab', 'planet4-blocks-backend')}
-                value={link_url_new_tab}
-                checked={link_url_new_tab}
-                onChange={changeSlideAttribute('link_url_new_tab', index)}
-              />
+            <div className='slide-item-order-arrows'>
+              <div
```
Inwerpsel, line 75: For accessibility we should make this a button. Else you cannot focus it with the keyboard, nor press it.

Then we can also put a `disabled` attribute on the button and don't need the class `disabled`.

Inwerpsel (commented):
I'll put these comments already as they mostly affect code ordering and would move around a lot of lines when addressed.

assets/src/blocks/CarouselHeader/SidebarSlide.js
GP-Dan-Tovbein: I wanted to use a `Button` but I figured out that, because of the Accessibility, you cannot nest interactive elements.

In other words, you cannot do this:
```
<button>
   <button></button>
</button>
```

As you see [here](https://github.com/WordPress/gutenberg/blob/7bf1297d3751dbb959e62c32666321a3af9a03dc/packages/components/src/panel/body.js#L100), the Gutenberg's PanelBody component is developed

Inwerpsel: We could also use something other than the `PanelBody` component as a container. Since it's a button, it's clearly not intended to put stuff in (my bad I didn't spot this earlier).

assets/src/blocks/CarouselHeader/Sidebar.js
```diff
@@ -1,43 +1,153 @@
...
-  currentSlide,
   changeSlideAttribute,
   goToSlide,
 }) => {
   const [dragTarget, setDragTarget] = useState(null);
   const [draggedSlide, setDraggedSlide] = useState(null);
-  const timeoutRef = useRef(null);
+  const slidesRef = useRef(null);
+
+  const getNode = (index) => (
```
Inwerpsel, line 22: This function can be defined outside of render.

GP-Dan-Tovbein: Do you mean `render` as a "scope"? Because this function is not being created through every render method.

GP-Dan-Tovbein: I understand that you're referring to move out of the component, am I right?

Inwerpsel: Yes, I meant render as in "when the component renders".

assets/src/blocks/CarouselHeader/Sidebar.js
```diff
@@ -1,43 +1,153 @@
...
+    document.querySelector(`.sidebar-slide[data-index="${index}"]`)
+  )
+
+  const upOrDownHandler = ( evt ) => {
```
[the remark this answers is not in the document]
Inwerpsel: Can also be done in a follow up

assets/src/blocks/CarouselHeader/SidebarSlide.js
```diff
@@ -23,45 +36,104 @@ export const SidebarSlide = ({
...
+                className={`arrow-button ${isLastItem ? 'disabled' : ''}`}
+                onClick={ upOrDownHandler }
+                data-index={index}
+                data-type='down' />
             </div>
+            <img
+              className='slide-item-image'
+              draggable={ false }
+              srcSet={ image_srcset }
```
Inwerpsel, line 89: If we also add `sizes='50px'` here it will help the browser load the right size.

Inwerpsel (approved):
Works well :+1: I see it uses direct DOM reading and manipulation instead of React for a few things. I'm not sure if each of these is really needed, e.g. maybe the data attributes can be avoided by passing `index` to the handler directly. But then again maybe for drag and drop the approach you chose is more reliable cross browser.

assets/src/blocks/CarouselHeader/SidebarSlide.js
Inwerpsel: We can also do this in a follow up, since avoiding `PanelBody` is not a quick fix, and without the PR there's no buttons at all. But since we're adding the button as the only option for keyboard users, we should make it a button element so that they can definitely reach it.

We could also add a tabindex attribute to this, which might make it focusable. But that's also potentially tricky in React.

GP-Dan-Tovbein:
I've just added the label UAT needed since is required.
FYI and since you've already approved this PR, I won't merge it until the Design team will approve it. Does it make sense @Inwerpsel ?

assets/src/styles/blocks/CarouselHeader/SidebarSlidesEditor.scss
```diff
@@ -0,0 +1,138 @@
...
+    display: none;
+  }
+
+  .components-panel__body-toggle {
+    pointer-events: none;
+    padding-right: 16px;
+    padding-left: 9px;
+  }
+
+  &.cloned-slide {
```
Inwerpsel, line 19: [This part](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/729/files#diff-b11e1bdfff8bc0685a960c674ccbed0d863e5c63327f1c2a0d93f65333b98290R19-R46) doesn't need to be "inside of" `.sidebar-slide` to work. Would save a few bytes in the compiled CSS.

Inwerpsel (approved):

PR #730, opened 2021-11-19
Decouple image hover controls from Columns block

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6471

Since the image buttons will be used for the Take Action Boxout block as well, we should first extract them from the columns block.

This also simplifies some of the surrounding code.

- Extract SVG which was duplicated in Columns and Carousel Header. Will be used for Boxout too.
- Create a separate component for the hover buttons.

I might do a follow up with more decoupling, but this seems like already a sufficient amount to review.

TESTING:
This should produce the same result as before, so if you see any difference with how it behaves in master it's probably not intended.
Columns: https://www-dev.greenpeace.org/test-pandora/wp-admin/post.php?post=24037&action=edit
Carousel Header: https://www-dev.greenpeace.org/test-pandora/wp-admin/post.php?post=70&action=edit

I noticed that undoing seems to be buggy after removing the image, `master` has the same issue. It sometimes works, and sometimes doesn't seem to perform the undo anymore, even if you had other edits before changing the image. Didn't see any related console messages yet. Maybe it's because of the `MediaUpload` render prop.

### DIFF
--- a/assets/src/blocks/Columns/EditableColumns.js
+++ b/assets/src/blocks/Columns/EditableColumns.js
@@ -15,7 +15,7 @@ export const EditableColumns = ({
   columnImages,
 }) => {
   const getImageOrButton = (openMediaModal, index) => {
-    if (columns[index] && columns[index].attachment && (0 < columns[index].attachment)) {
+    if ((0 < columns[index].attachment)) {
 
       return <div className='columns-image-container'>
         <ImageHoverControls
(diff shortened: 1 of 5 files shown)


planet-4:

:watch: 2021.11.19 11:10:00

assets/src/blocks/Columns/EditableColumns.js
```diff
@@ -13,66 +14,45 @@ export const EditableColumns = ({
...
-              </Button>
-              <Button
-                className='remove-image'
-                onClick={() => toAttribute('attachment', index)(0)}
-                icon='trash'
-              />
-            </div>
-            <img
-              src={columnImages[columns[index].attachment]}
-              onClick={openEvent}
```
Inwerpsel, line 38: Didn't work because it's covered by an overlay.

mleray (approved):
Just a minor bug that can be solved in a follow-up PR since it was already the case before the refactor, in the Columns block with the Images style the `Add image` button is not in the center:
Screenshot 2021-11-19 at 14 26 00

PR #732, opened 2021-11-19
PLANET-6471 Use same image controls as Columns block

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6471

* Move buttons-overlay class outside of the columns-specific class so it can be reused.
* Put styles on wrapper div instead of on img in editor only.
* Convert earlier added unintentional double quotes to single.

The image controls should work the same as for the columns block. The alignment issue is resolved, but only for the Boxout block. I'll address that in a follow up because it needs more changes in the Columns block's editor code.

Example on test instance: https://www-dev.greenpeace.org/test-venus/wp-admin/post.php?post=50288&action=edit

### DIFF
--- a/assets/src/components/ImageHoverControls/index.js
+++ b/assets/src/components/ImageHoverControls/index.js
@@ -6,21 +6,31 @@ export const ImageHoverControls = props => {
     onEdit,
     onRemove,
     isCompact,
+    isAdd,
   } = props;
 
-  return <div className="buttons-overlay">
-    <Button
+  return <div className='buttons-overlay'>
+    { isAdd && <Button
       onClick={ onEdit }
-      icon="edit"
+      icon='plus-alt2'
       isPrimary
-      className="edit-image"
+      className='edit-image'
+    >
+      { __('Add image', 'planet4-blocks-backend') }
+    </Button> }
+
+    { !isAdd && <Button
+      onClick={ onEdit }
+      icon='edit'
+      isPrimary
+      className='edit-image'
     >
       { !isCompact && __('Edit', 'planet4-blocks-backend') }
-    </Button>
-    <Button
-      className="remove-image"
+    </Button> }
+    { !isAdd && <Button
+      className='remove-image'
       onClick={ onRemove }
-      icon="trash"
-    />
+      icon='trash'
+    /> }
   </div>;
 };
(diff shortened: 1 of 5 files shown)


planet-4:

:new_moon: [venus](https://www-dev.greenpeace.org/test-venus) | [admin](https://www-dev.greenpeace.org/test-venus/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-venus/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-venus) | [composer-local.json](https://github.com/greenpeace/planet4-test-venus/blob/main/composer-local.json)

:watch: 2021.11.19 16:37:33

assets/src/styles/blocks/ColumnsEditor.scss
```diff
@@ -1,52 +1,52 @@
...
 
-  .buttons-overlay {
-    position: absolute;
-    z-index: 2;
-    align-items: center;
-    justify-content: center;
-    width: 100%;
-    height: 100%;
-    display: flex;
+.buttons-overlay {
```
Inwerpsel, line 5: I'll move this to a separate file later. Didn't change anything inside of `.buttons-overlay`, it all still worked and works for the other block too.

That's why you should avoid needlessly putting classes inside of other classes, it makes the code way less reusable.

assets/src/components/ImageHoverControls/index.js
```diff
@@ -6,21 +6,31 @@ export const ImageHoverControls = props => {
...
+  return <div className='buttons-overlay'>
+    { isAdd && <Button
       onClick={ onEdit }
-      icon="edit"
+      icon='plus-alt2'
       isPrimary
-      className="edit-image"
+      className='edit-image'
+    >
+      { __('Add image', 'planet4-blocks-backend') }
```
Inwerpsel, line 19: This add button is not used by the Columns block yet, because it's conditionally displayed there.
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/09068d6e833ce660389916cd296259a68b660f6c/assets/src/blocks/Columns/EditableColumns.js#L30-L39

mleray (approved):

PR #733, opened 2021-11-22
Workaround to prevent local images CarouselHeader

Inwerpsel:
* We were not able to reproduce the issue yet, but it keeps on occurring
that a local image is used for the largest size of an image. Local means
it's on the PHP server, and doesn't use the CDN URL. This source is
removed when we deploy, resulting in an empty image.

While this doesn't fully address [the issue](https://jira.greenpeace.org/browse/PLANET-6372), it does prevent the worst effect (browser using 404 image).

I also added a blocking alert message, so that the editor in question knows it happens, and can contact the Planet 4 team with reproduction steps.

### DIFF
--- a/assets/src/blocks/CarouselHeader/EditableBackground.js
+++ b/assets/src/blocks/CarouselHeader/EditableBackground.js
@@ -4,6 +4,39 @@ import { Button, Dropdown } from '@wordpress/components';
 import { toSrcSet } from './CarouselHeaderEditor';
 const { __ } = wp.i18n;
 
+const usesCdn = window.p4ge_vars.uses_cdn;
+
+// If the site uses CDN through the stateless plugin, it should never have the local folder in the URL.
+const isWrongSource = src => {
+  return usesCdn && src.includes('wp-content/uploads');
+}
+
+const ALERT_TEXT = `
+An issue with the image path was detected. This usually happens on the largest version of the image.
+Please try removing and re-adding the images until you don't get this alert anymore.
+If the alert persists, please provide the prior steps you did (including upload of image) to the Planet 4 team.
+`;
+
+// Due to a yet unidentified bug, it sometimes happens that the URL and the largest size don't use the CDN URL.
+// This has a nasty consequence: the URL on the server exists, but only temporary. It is removed on deploy. So an editor
+// can not see that this issue will occur. It's probably related to WPML.
+const preventLocalImage = (url, sizes) => {
+  const fixedSizes = sizes.filter(size => !isWrongSource(size.url || size.source_url));
+  let fixedUrl = url;
+  if (isWrongSource(url)) {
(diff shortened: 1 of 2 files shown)


assets/src/blocks/CarouselHeader/EditableBackground.js
```diff
@@ -4,6 +4,31 @@ import { Button, Dropdown } from '@wordpress/components';
 import { toSrcSet } from './CarouselHeaderEditor';
 const { __ } = wp.i18n;
 
+const customCdnUrl = !window.p4ge_vars.stateless_url;
```
Inwerpsel, line 7: This probably needs further checks to see if the custom URL is actually enabled.

planet-4:

:watch: 2021.11.22 13:18:01

sagarsdeshmukh:
Hi @Inwerpsel , This fix looks good to me in mean time, till we reach to the final conclusion of issue. 👍

classes/class-loader.php
```diff
@@ -342,6 +342,7 @@ public function enqueue_editor_scripts( $hook ) {
 			'features'        => [
 				'feature_engaging_networks' => $en_active,
 			],
+			'uses_cdn'        => get_option( 'sm_mode' ),
```
lithrel, line 345: It can be absent but also have different values like `cdn` `backup` or `disabled`, I guess we only want to aim at the `cdn` one, so `'cdn' === get_option( 'sm_mode' )` might be better ?

Inwerpsel: I added the check for both `cdn` and `ephemeral`. I also added `stateless`, which seems to be the old name for `ephemeral`, in case a site would still need that.

lithrel (dismissed):
:+1:

classes/class-loader.php
```diff
@@ -336,12 +336,15 @@ public function enqueue_editor_scripts( $hook ) {
 		$en_active = ! MigrationLog::from_wp_options()->already_ran( M001EnableEnFormFeature::get_id() )
 					|| Features::is_active( Features::ENGAGING_NETWORKS );
 
+		$uses_cdn = is_plugin_active( 'wp-stateless/wp-stateless-media.php' ) && get_option( 'sm_custom_domain' );
```
Inwerpsel, line 339: I had to change the check to this because we use a constant to set the mode, and [it's always `stateless](https://github.com/search?q=org%3Agreenpeace+%22WP_STATELESS_MEDIA_MODE%22&type=code)`.

lithrel (approved):
:+1:

classes/class-loader.php
Inwerpsel: See https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/733#discussion_r766006227

PR #734, opened 2021-11-22
PLANET-5609 Covers block carousel layout

mleray:
### Description

See [PLANET-5609](https://jira.greenpeace.org/browse/PLANET-5609) and the designs [here](https://p4-designsystem.greenpeace.org/05f6e9516/p/66c4dc-covers/b/460472)

### Testing

On local, you can play around with the different styles and layouts available for the Covers block. Otherwise you can see all the possible versions on [this page](https://www-dev.greenpeace.org/test-jupiter/covers-block-layouts/) I created for UAT 🙂


assets/src/blocks/Covers/CoversCarouselButton.js
```diff
@@ -0,0 +1,8 @@
+const { __ } = wp.i18n;
+
+export const CoversCarouselButton = ({ handler, direction, disabled = false }) => (
+  <button className={`carousel-control-${direction}`} disabled={disabled} role="button" onClick={handler}>
```
GP-Dan-Tovbein, line 4: You don't need to set `role="button"` to any `button` since it has already set this value by default.

Take a look the column from here https://www.w3.org/TR/2014/REC-html5-20141028/dom.html#aria-role-attribute
**Language feature**: `button` element 
**Default implicit ARIA semantic**: `button` role

mleray: Removed 👍

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -1,59 +1,51 @@
...
+        <div key={name} className='cover'>
+          <a
+            href={campaignLink}
+            data-ga-category='Campaign Covers'
+            data-ga-action='Image'
+            data-ga-label='n/a'
+            aria-label={__('Check our campaign about ' + name, 'planet4-blocks')}
+          >
+            <div className='thumbnail-large'>
+              {image && image[0] &&
```
GP-Dan-Tovbein, line 34: Why do directly check for length `{image.length && ...`?

assets/src/blocks/Covers/CoversCarouselControls.js
```diff
@@ -0,0 +1,48 @@
+const { __ } = wp.i18n;
+
+const CoversCarouselArrow = ({ handler, direction, disabled = false }) => (
+  <button className={`carousel-control-${direction}`} disabled={disabled} onClick={handler}>
+    <span className={`carousel-control-${direction}-icon`} aria-hidden="true"><i></i></span>
+    <span className="visually-hidden">{__(direction === 'next' ? 'Next' : 'Prev', 'planet4-blocks')}</span>
+  </button>
+);
+
+
```
GP-Dan-Tovbein, line 10: Reme this empty line

assets/src/blocks/Covers/CoversEditor.js
```diff
@@ -1,22 +1,25 @@
...
 
 import { InspectorControls } from '@wordpress/block-editor';
 import TagSelector from '../../components/TagSelector/TagSelector';
 import PostSelector from '../../components/PostSelector/PostSelector';
 import PostTypeSelector from '../../components/PostTypeSelector/PostTypeSelector';
-import { Covers, COVER_TYPES } from './Covers';
+import { Covers } from './Covers';
+import { COVERS_TYPES, COVERS_LAYOUTS, CAROUSEL_LAYOUT_COVERS_LIMIT } from './CoversConstants';
 import { useCovers } from './useCovers';
+import { CoversCarouselButton } from './CoversCarouselButton';
```
GP-Dan-Tovbein: You missed to add this file

assets/src/blocks/Covers/CoversEditor.js
```diff
@@ -110,7 +135,15 @@ const renderView = (attributes, toAttribute) => {
         <div className='EmptyMessage'>
           {__('Block content is empty. Check the block\'s settings or remove it.', 'planet4-blocks-backend')}
         </div> :
-        <Covers {...coversProps} />
+        <div className='covers-container'>
+          <Covers {...coversProps} />
+          {covers.length > 4 && isCarouselLayout && (
+            <CoversCarouselButton direction='prev' handler={() => {}} />
```
GP-Dan-Tovbein: As I mentioned before, you missed to add this Component. However, I would like to know how are you handle this `handler` method.

assets/src/blocks/Covers/CoversFrontend.js
```diff
@@ -1,32 +1,66 @@
...
 export const CoversFrontend = attributes => {
-  const { initialRowsLimit, cover_type, title, description, covers, className } = attributes;
+  const { initialRowsLimit, cover_type, title, description, covers, className, layout } = attributes;
+  const coversContainerRef = useRef(null);
+
+  const {
+    showMoreCovers,
+    row,
+    amountOfCoversPerRow,
+    slideCovers,
```
GP-Dan-Tovbein, line 16: As far as I see this prop is a custom `hook`, so, I'm wondering why you cannot import this hook here instead of passing it as prop?

assets/src/blocks/Covers/TakeActionCovers.js
```diff
@@ -40,65 +36,61 @@ export const TakeActionCovers = ({ initialRowsLimit, covers, row, loadMoreCovers
...
+                <img
+                  alt={alt_text}
+                  src={image}
+                  srcSet={srcset}
+                  sizes={IMAGE_SIZES.takeAction}
                 />
+              </a>
+              <div className='cover-card-content'>
+                {/* Regardless of how many tags there are, we only show the first one */}
+                {tags && tags.length > 0 && <span className='cover-card-tag'>{tags[0].name}</span>}
```
GP-Dan-Tovbein: I think that `{tags && tags.length && ...` or maybe `{tags.length}` are fine too.

assets/src/styles/blocks/Covers/layouts/CoversCarouselLayout.scss
```diff
@@ -0,0 +1,97 @@
...
+
+  .covers {
+    display: flex;
+    overflow-x: scroll;
+    flex-wrap: nowrap;
+    -ms-overflow-style: none;
+    scrollbar-width: none;
+    scroll-behavior: smooth;
+
+    &::-webkit-scrollbar {
```
GP-Dan-Tovbein, line 19: Why Webpack doesn't solve this workaround?

GP-Dan-Tovbein (changes requested):
Code looks good.
I left some comments, however, I will still reviewing since it's very large PR.

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -1,59 +1,51 @@
...
 
-          if (hideCover) {
-            return null;
-          }
+      if (hideCover) {
+        return null;
+      }
 
-          const campaignLink = inEditor ? null : href;
+      const campaignLink = inEditor ? null : href;
```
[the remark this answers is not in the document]
mleray: I think it's ok to keep it as is for now, it's more consistent with the other Covers styles and it might be useful in the future since we will eventually update the [designs for the Campaign Covers](https://p4-designsystem.greenpeace.org/05f6e9516/p/66c4dc-covers/b/937749)

assets/src/styles/blocks/Covers/layouts/CoversCarouselLayout.scss
mleray: I'm not sure... But if I remove this line then the scrollbar is shown 😕

assets/src/blocks/Covers/TakeActionCovers.js
mleray: I think we've already talked about this issue in another PR 😄 but if I just put `tags.length` then it will show "0" when there are no tags

assets/src/blocks/Covers/CampaignCovers.js
mleray: This was already [written like this](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/Covers/CampaignCovers.js#L31), so I'm not sure what the initial logic behind it was but it is shorter anyway to write it like this than it is to write `image && image.length > 0` so I'm ok with it 🤷‍♀️

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -1,59 +1,51 @@
...
-                      src={image[0]}
-                      alt={alt_text}
-                      title={alt_text}
-                    />
-                  }
-                  <span className='yellow-cta'><span aria-label='hashtag'>#</span>{name}</span>
-                </div>
-              </a>
+      return (
+        <div key={name} className='campaign-card-column cover'>
```
GP-Dan-Tovbein, line 25: `name` is always `undefined`. You might to define `defined` values for `keys`, maybe change to `id`?

mleray: It shouldn't be undefined, we also use it in other places on the cover and it works there (such as the hashtag) 🤔

mleray:
Thanks for your testing @GP-Dan-Tovbein 💚 I'll definitely look more into the migration of deprecated blocks, maybe we need a new `coversV2` file 🤔 
For the keys issue, I wouldn't worry about it now as I'm pretty sure it's due to the style icons and that will be fixed in [PLANET-6106](https://jira.greenpeace.org/browse/PLANET-6106)! It's also already happening in the current version of the block, before these changes, so I think it would anyway be best to solve it in a separate PR 🙂

assets/src/blocks/Covers/CoversFrontend.js
mleray: Sorry, I forgot about this comment! But it's not passed down as a prop 🤔

assets/src/blocks/Covers/useCovers.js
```diff
@@ -1,14 +1,52 @@
...
+      }
+      break;
+    case COVERS_TYPES.takeAction:
+      if (layout === COVERS_LAYOUTS.carousel) {
+        setAmountOfCoversPerRow(4);
+      } else {
+        setAmountOfCoversPerRow(isSmallWindow() ? 2 : 3);
+      }
+      break;
+    default:
```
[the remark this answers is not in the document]
mleray: What would you return? 🤔 If it's not one of these types then nothing should render anyway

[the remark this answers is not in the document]
GP-Dan-Tovbein: I would prefer to remove the `default` case in this case since there is no logic here. What do you think?

assets/src/blocks/Covers/CoversCarouselControls.js
```diff
@@ -0,0 +1,46 @@
...
+  const amountOfSlides = Math.ceil(totalAmountOfCovers / amountOfCoversPerRow);
+  const indicators = Array.from(Array(amountOfSlides).keys());
+
+  return (
+    <>
+      <CoversCarouselArrow
+        direction='prev'
+        disabled={currentRow === 1}
+        handler={() => slideCovers('prev')}
```
GP-Dan-Tovbein, line 28: Have you tried something like this?
```
<CoversCarouselArrow
        direction='prev'
        disabled={currentRow === 1}
        {...(slideCovers) && {handler: () => slideCovers('prev')}}
```
And then avoid to pass something like this:
https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/734/files#diff-5e8e8bb129161a155bd56c85e46aea8d5d068edca136765f068b35f9286295c3R150

I haven't checked but I think it would work. What do you think?

assets/src/blocks/Covers/CoversFrontend.js
GP-Dan-Tovbein: Sorry You're right, I was confused with the `CoversCarouselControls` from [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/734/files#diff-93c303aba0166eba7362c8c44643bb67e809e03106503a13337b8e00ce45a7a5R53)

assets/src/blocks/Covers/useCovers.js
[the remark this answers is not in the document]
mleray: Sure, we can remove it 👍

assets/src/blocks/Covers/CoversCarouselControls.js
```diff
@@ -0,0 +1,46 @@
...
+  <button className={`carousel-control-${direction}`} disabled={disabled} onClick={handler}>
+    <span className={`carousel-control-${direction}-icon`} aria-hidden="true"><i></i></span>
+    <span className="visually-hidden">{__(direction === 'next' ? 'Next' : 'Prev', 'planet4-blocks')}</span>
+  </button>
+);
+
+export const CoversCarouselControls = ({
+  totalAmountOfCovers,
+  amountOfCoversPerRow,
+  slideCovers,
```
GP-Dan-Tovbein, line 13: As I mentioned in another comment, another approach could be to import the `useCovers` hook and `destructure the `slideCovers` method, instead of passing it as a prop. It's just a suggestion but it's up to you. 

And then [this](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/734/files#r772401503)

Does it make sense?

assets/src/blocks/Covers/useCovers.js
```diff
@@ -39,11 +77,52 @@ export const useCovers = ({ post_types, tags, cover_type, initialRowsLimit, post
...
     }
-  }, [ cover_type, post_types, tags, posts ]);
+  }, [cover_type, post_types, tags, posts, layout]);
+
+  useEffect(() => {
+    updateRowCoversAmount();
+
+    if (layout !== COVERS_LAYOUTS.carousel) {
+      window.addEventListener('resize', updateRowCoversAmount);
+      return () => window.removeEventListener('resize', updateRowCoversAmount);
```
GP-Dan-Tovbein, line 123: Just an opinion, I would prefer to not `return` this removed listener. maybe this will be better 
```
return () => { window.removeEventListener('resize', updateRowCoversAmount); }
```

GP-Dan-Tovbein (approved):
code looks good to me! 

However, I've left some suggestions to improve but we could follow up in the coming implementations. 

Good job @mleray

assets/src/blocks/Covers/useCovers.js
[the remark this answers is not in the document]
mleray: Actually, nvm, linting does complain indeed if we try to remove it 😬

assets/src/blocks/Covers/useCovers.js
mleray: Hm, I think it's fine as is, we use similar notations in several other places 🙂 Such as [CarouselHeaderFrontend](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js)

assets/src/blocks/Covers/CoversCarouselControls.js
mleray: That's a good point! I've made the changes 👌

GP-Dan-Tovbein (approved):
Code still looking good to me!

classes/blocks/class-covers.php
```diff
@@ -364,7 +364,7 @@ private static function populate_posts_for_campaigns( &$fields ) {
 
 			if ( ! empty( $attachment_id ) ) {
 				$tag_remapped['image']    = wp_get_attachment_image_src( $attachment_id, 'medium_large' );
-				$tag_remapped['src_set']  = wp_get_attachment_image_srcset( $attachment_id, 'medium_large' );
+				$tag_remapped['src_set']  = wp_get_attachment_image_srcset( $attachment_id, 'medium_large' ) ?? 'false';
```
Inwerpsel, line 367: The `??` operator will only use the second operand if the first one is `null`. In this case the function never returns null, but a bool instead. Even if it did, why would this need to be the string `'false'`? :thinking:

mleray: I was getting console warnings in the frontend, that `srcSet` was set to a boolean value and needed a string instead

Inwerpsel: `'false'` is not really better because React also does not know what to do with it for the `srcset` prop.

PR #745, opened 2021-12-10
PLANET-6432: Add reading time to post details

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6432

Cf. https://github.com/greenpeace/planet4-master-theme/pull/1542

### DIFF
--- a/assets/src/blocks/Articles/ArticlePreview.js
+++ b/assets/src/blocks/Articles/ArticlePreview.js
@@ -1,6 +1,7 @@
 import { Component } from '@wordpress/element';
 import { unescape } from '../../functions/unescape';
 import { IMAGE_SIZES } from './imageSizes';
+const { __ } = wp.i18n;
 
 export class ArticlePreview extends Component {
   constructor(props) {
@@ -103,6 +104,7 @@ export class ArticlePreview extends Component {
         post_title,
         post_excerpt,
         date_formatted,
+        reading_time,
       }
     } = this.props;
 
@@ -165,6 +167,16 @@ export class ArticlePreview extends Component {
               <time className="article-list-item-date" dateTime="">
                 {date_formatted}
               </time>
+              }
+
+              {(authorLink || date_formatted) && reading_time &&
+                <span className="article-list-item-bullet" aria-hidden="true">&#8226;</span>
+              }
+
+              {reading_time &&
+                <span className="article-list-item-readtime">
+                  { __( '%d min read', 'planet4-master-theme' ).replace('%d', reading_time) }
+                </span>
               }
             </p>
           </header>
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2021.12.13 08:33:16

GP-Dan-Tovbein (commented):
I was testing this feature and works well, good job! However, I detected an inconsistent behaviour and I would like your opinion.

I know that the reading time is calculated by length of words, thus, I've created 2 different pages with the same content, but in one of them I removed all spaces. (this is a very special case LOL)

[With spaces](https://www-dev.greenpeace.org/test-pandora/blogs/49626/reading-time-demo-with-spaces/): 19 min
[Without spaces](https://www-dev.greenpeace.org/test-pandora/blogs/49630/reading-time-demo-without-spaces/): 2 min

What do you think?

lithrel:
> What do you think?

That's an interesting test :grin:  I can see a few solutions for that, where we adjust the score with the actual total length of the text; but the [ICU](https://icu.unicode.org/) lib that is used is made for a real language usage, with [well defined word/sentence separators, punctuation, etc.](https://github.com/unicode-org/icu/blob/49dda34fb175240a7724c7e039a270126ff7d900/icu4c/source/data/locales/fr.txt), depending on the locale selected. So I don't think it's necessary to handle this case for now, we can work on it if an editor has a real use case. And if we end up writing a manual override of the read time by post, this would probably become irrelevant.

Inwerpsel:
> > What do you think?
> 
> That's an interesting test grin I can see a few solutions for that, where we adjust the score with the actual total length of the text; but the [ICU](https://icu.unicode.org/) lib that is used is made for a real language usage, with [well defined word/sentence separators, punctuation, etc.](https://github.com/unicode-org/icu/blob/49dda34fb175240a7724c7e039a270126ff7d900/icu4c/source/data/locales/fr.txt), depending on the locale selected. So I don't think it's necessary to handle this case for now, we can work on it if an editor has a real use case. And if we end up implementing a manual override of the read time by post, this would probably become irrelevant.

It's also something that struck me when investigating the reading time (see [relevant section in doc](https://docs.google.com/document/d/1D3dUrNwNBKUCyYXHSytd8fg-3b7QdV2SvmFnSI592mE/edit#heading=h.b4cgjfnpd4o9)), as I was familiar with WPM from typing speed which actually does count the characters. But it provides a good approximation in almost all cases.

Inwerpsel:
I guess the submodule update is in here by accident?

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -103,6 +104,7 @@ export class ArticlePreview extends Component {
         post_title,
         post_excerpt,
         date_formatted,
+        reading_time
```
GP-Dan-Tovbein, line 107: Small detail.. missing the `,` at the EOL

PR #752, opened 2021-12-22
PLANET-6106 Use block examples for Covers styles

mleray:
### Description

See [PLANET-6106](https://jira.greenpeace.org/browse/PLANET-6106)
This is to give a better idea of the block styles when the block is empty. This PR also solves the missing keys console warnings that showed up when switching styles, by adding an id to all covers instead of using a different attribute per style such as `name` or `post_title`. This new id also makes sure that there are no duplicates, whereas before the fix there could have been some edge cases with the same post title for example.

### Testing

On your local, you can add a Covers block to a page and check the new styles in the editor, in the sidebar. Otherwise you can check it out on the test instance, for instance on the [Act page](https://www-dev.greenpeace.org/test-nix/act) that already has a Covers block 🙂

### DIFF
--- a/assets/src/blocks/Covers/CampaignCovers.js
+++ b/assets/src/blocks/Covers/CampaignCovers.js
@@ -1,3 +1,4 @@
+import { CoversImagePlaceholder } from './CoversImagePlaceholder';
 import { IMAGE_SIZES } from './imageSizes';
 
 const { __ } = wp.i18n;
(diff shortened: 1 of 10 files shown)


planet-4:

:watch: 2021.12.22 09:35:26

lithrel (approved):
:+1:

PR #754, opened 2021-12-22
Remove covers ids

Inwerpsel:
We can use the URLs of the results as a key instead of adding an ID.

This avoids complicating the front end code even further to support the editor. We should really keep front end and editor separate so that we don't have "Swiss army knife" style files that do everything and are very hard to change.

This should again result in console warnings, but those are harmless and just noisy.

The root cause is that the data for each cover type is different. The URL is called `href`, `link` and `button_link` in the 3 styles.

The styles in the sidebar are using WP's `registerBlockStyle` API, which is only intended to affect which class is used for CSS.

The current code [adds an effect to the class name](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d77070ac49a5116e3b8d3139cb9226a9567adccc/assets/src/blocks/Covers/CoversEditor.js#L175-L187), which then, in [another effect](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d77070ac49a5116e3b8d3139cb9226a9567adccc/assets/src/blocks/Covers/useCovers.js#L81-L85), results in an effect with new API call to get the matching results and data format.

### DIFF
--- a/assets/src/blocks/Covers/CampaignCovers.js
+++ b/assets/src/blocks/Covers/CampaignCovers.js
@@ -24,7 +24,7 @@ export const CampaignCovers = ({
       const campaignLink = inEditor ? null : href;
 
       return (
-        <div key={name} className='campaign-card-column cover'>
+        <div key={href} className='campaign-card-column cover'>
           <a
             href={campaignLink}
             data-ga-category='Campaign Covers'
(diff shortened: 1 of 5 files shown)


planet-4:

:watch: 2021.12.22 14:46:57

mleray (approved):
👍

PR #755, opened 2021-12-23
Use consistent attribute names for all Covers styles 

mleray:
### Description

This was causing console warnings for missing keys issues. It also makes more sense to use the same names everywhere as the style of Covers should only impact the CSS and not the data used.

### DIFF
--- a/assets/src/blocks/Covers/CampaignCovers.js
+++ b/assets/src/blocks/Covers/CampaignCovers.js
@@ -14,37 +14,41 @@ export const CampaignCovers = ({
 }) => (
   <div className='covers'>
     {covers.map((cover, index) => {
-      const { href, image, alt_text, name, src_set } = cover;
+      const { image, alt_text } = cover;
       const hideCover = !isCarouselLayout && !!initialRowsLimit && index >= row * amountOfCoversPerRow;
 
       if (hideCover) {
         return null;
       }
 
-      const campaignLink = inEditor ? null : href;
+      const link = cover.link || cover.href;
+      const srcset = cover.srcset || cover.src_set;
+      const title = cover.title || cover.name;
+
+      const campaignLink = inEditor ? null : link;
 
       return (
-        <div key={href} className='campaign-card-column cover'>
+        <div key={link} className='campaign-card-column cover'>
           <a
             href={campaignLink}
             data-ga-category='Campaign Covers'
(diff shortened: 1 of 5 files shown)


planet-4:

:watch: 2021.12.24 08:30:11

assets/src/blocks/Covers/CampaignCovers.js
```diff
@@ -14,37 +14,37 @@ export const CampaignCovers = ({
 }) => (
   <div className='covers'>
     {covers.map((cover, index) => {
-      const { href, image, alt_text, name, src_set } = cover;
+      const { link, image, alt_text, tag, srcset, src_set, href, name } = cover;
```
Inwerpsel, line 17: Maybe it's easier if we handle this in one place?
```
const link = cover.link || cover.href;
const tag = cover.tag || cover.name;
const srcset = cover.srcset || cover.src_set;

File: assets/src/blocks/Covers/CampaignCovers.js [hunk 2]
```diff
@@ -14,37 +14,41 @@ export const CampaignCovers = ({
...
       const hideCover = !isCarouselLayout && !!initialRowsLimit && index >= row * amountOfCoversPerRow;
 
       if (hideCover) {
         return null;
       }
 
-      const campaignLink = inEditor ? null : href;
+      const link = cover.link || cover.href;
+      const srcset = cover.srcset || cover.src_set;
+      const tag = cover.tag || cover.name;
```
Inwerpsel, line 26: Shouldn't this be called `title`? :thinking:

mleray: I was hesitating about that, but since it's not really a title I decided to give it a more accurate name... But we could indeed use `title` instead

Inwerpsel: That makes sense, but then we again have the problem of the block not receiving the expected data when switching between styles.

That's why we shouldn't use the block styles to control what data is being shown, this API is only intended by WP to put the class name there. I guess this means rethinking how the block works, or converting the `registerBlockStyles` thing to our own component that can actually work with our logic.

mleray: Yes it makes sense, just added a commit with this change!

Inwerpsel (approved):

PR #758, opened 2022-01-11
Feature/turn sharebuttons to a block

GP-Dan-Tovbein:
This PR is about to turn (and improve) the current [ShareButtons](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/tree/master/assets/src/components/ShareButtons) component to a `block`. Initially will be used by the HB thank you message template by a template block, and afterwards could be used by any other by just adding `planet4-blocks/share-buttons`.

Demo page: https://www-dev.greenpeace.org/test-nix/new-sharebuttons-block-instead-of-using-as-a-component/

**Remember that this is a `beta` block**

*Features*
- UTM values can be set by the editor
- Google Analytics share values can be set by the editor
- WhatsApp share values can be set by the editor
- Facebook share values can be set by the editor
- Twitter share values can be set by the editor
- Email share values can be set by the editor
- Show/hide any of this four share buttons
- We will be able to add any other social share button easily afterwards

[Related PR](https://github.com/greenpeace/planet4-master-theme/pull/1581)

Screenshot 2022-01-19 at 09 27 21
Screenshot 2022-01-19 at 09 27 32


planet-4:

:watch: 2022.01.13 09:48:33

classes/blocks/class-sharebuttons.php
```diff
@@ -0,0 +1,75 @@
...
+	];
+
+	/**
+	 * Register ShareButtons block.
+	 */
+	public function register_sharebuttons_block() {
+		register_block_type(
+			self::get_full_block_name(),
+			[
+				'render_callback' => [ $this, 'front_end_rendered_fallback' ],
```
Inwerpsel, line 58: This fallback is not needed here because it's the first version of the block.

classes/blocks/class-sharebutton.php
```diff
@@ -0,0 +1,63 @@
+<?php
```
Inwerpsel, line 1: Not every block needs a PHP file. We only need to add this file once the block needs back end logic (rendering or data fetching). The JS files are sufficient for the block to be usable if it's saved as HTML.

assets/src/blocks/ShareButtons/EmailShare/EmailShareFrontend.js
```diff
@@ -0,0 +1,32 @@
+import { SvgIcon } from "../SvgIcon";
+import { parseUtmParams } from '../utils';
+
+const { __ } = wp.i18n;
+
+export const EmailShareFrontend = ({
```
Inwerpsel: All these provider specific components are mostly the same code, except for a few props (`href`, `className`, `SvgIcon.name`, and the screen reader text).

Wouldn't it be better to create a single `ShareButton` component?
```js
<ShareButton
  // ...paramsThatAreSameForAll,
  href={`mailto:...`}
  className='email'
  icon='envelope'
>

File: assets/src/blocks/ShareButtons/EmailShare/EmailShareFrontend.js [hunk 4]
```diff
@@ -0,0 +1,32 @@
...
+    className="social-share-button email"
+    data-ga-event={gaEvent}
+    data-ga-event-category={gaEventCategory}
+    data-ga-category={gaCategory}
+    data-ga-action={gaAction}
+    data-ga-label={gaLabel}
+    { ...openInNewTab && { target: '_blank' } }
+  >
+    <SvgIcon {...{name: "envelope"}} />
+    <span className="visually-hidden">{__( 'Share via', 'planet4-master-theme' )} Email</span>
```
Inwerpsel: We should make the entire string translatable: `"Share via Email"`. I don't think all languages use the word "email", and it might not come at the end either.

Inwerpsel: The text domain should be `planet4-blocks-backend` for admin strings in the plugin..

assets/src/blocks/ShareButtons/FacebookShare/FacebookShareFrontend.js
```diff
@@ -0,0 +1,31 @@
...
+  utmCampaign,
+  gaEvent,
+  gaEventCategory,
+  gaCategory,
+  gaAction,
+  gaLabel,
+  openInNewTab,
+  baseSharedUrl,
+}) => (
+  <a href={ `${baseSharedUrl}?u=${encodeURIComponent(url)}&${parseUtmParams('facebook', utmMedium, utmContent, utmCampaign)}` }
```
Inwerpsel: I don't see why we would define this `baseSharedUrl` outside of this component, [in the default value of `ShareButtons` `facebook` attribute](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/758/files#diff-e0501445f43571f76b44fabd75c63eff89481ef28f79a2cfb31d203654f86c84R50). It's only ever used here, it could be a simple `const` in this file.

classes/blocks/class-sharebuttons.php
GP-Dan-Tovbein: Oh sorry, I missed to remove it! Thanks for pointed it this

assets/src/blocks/ShareButtons/EmailShare/EmailShareFrontend.js
GP-Dan-Tovbein: It's a really good point!

GP-Dan-Tovbein: It was implemented separately since I initially thought to develop each as a block, but now it doesn't make sense!

assets/src/blocks/ShareButtons/EmailShare/EmailShareFrontend.js
```diff
@@ -0,0 +1,32 @@
...
+  gaCategory,
+  gaAction,
+  gaLabel,
+  openInNewTab,
+  title,
+  body,
+}) => (
+  <a href={ `mailto:?subject=${title}&body=${body ? encodeURIComponent(body) : ''}\n${url}&${parseUtmParams('email', utmMedium, utmContent, utmCampaign)}` }
+    className="social-share-button email"
+    data-ga-event={gaEvent}
```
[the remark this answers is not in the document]
GP-Dan-Tovbein: But the same happens with `event-category`, so I prefer to keep it both and then set a default value. It would be better if the editor prefers to edit all `ga` values.

assets/src/blocks/ShareButtons/FacebookShare/FacebookShareFrontend.js
GP-Dan-Tovbein: Because any base shared url could change in the future, so, I think it would be better to allow to editors to edit them in case of, for example, Facebook updates its API.

Inwerpsel: Can you give me an indication of how much you expect these URLs to change?

assets/src/blocks/ShareButtons/EmailShare/EmailShareFrontend.js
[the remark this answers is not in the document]
Inwerpsel: I don't see why that means we need to mirror the implementation of one into the other? I mean, in the same spirit you can create a whole lot of props that are never really used.

assets/src/blocks/ShareButtons/FacebookShare/FacebookShareFrontend.js
GP-Dan-Tovbein: > Can you give me an indication of how much you expect these URLs to change?
A while time ago I had to lead with the Instagram API that its URL changed to a FB domain. But honestly, it does not happen frequently 🙏

assets/src/blocks/ShareButtons/ShareButtonsBlock.js
```diff
@@ -0,0 +1,87 @@
...
+      utmContent: {
+        type: 'string',
+      },
+      utmCampaign: {
+        type: 'string',
+      },
+      buttons: {
+        type: 'object',
+        default: {
+          facebook: {
```
Inwerpsel, line 48: Is there a reason for using an object over an array, for these buttons?

GP-Dan-Tovbein: Yes, It's more performant than using the `filter` for Arrays. In this case you will only update a `key|value` instead of finding an object, update and replace it into the Array

Inwerpsel: Can you give me an indication about how much performance difference we're talking about here?

GP-Dan-Tovbein: Here is a good post
http://www.andygup.net/fastest-way-to-find-an-item-in-a-javascript-array/

Inwerpsel: I meant rather in this particular case. How much performance do you expect to gain here by using an object instead of an array?

GP-Dan-Tovbein: Don't know exactly the value but is more faster than using an array for filtering.

Take in mind we have (for now) four Social Shares added to the attributes, and at least 2 of them have more than a field, to be exactly we have almost 10 fields to update. If we use a `filter`method for updating these particular values, we'll be iterating 10 times per each update.

Inwerpsel: I agree that objects can be more convenient than using `find` or `filter` on an array.

But if you're basing the implementation on performance considerations, it helps to know the approximate amounts of time and memory we're talking about. Are we talking about 10 times faster? Or just 1/1000000 faster? Or even less?

I did a [quick benchmark with a similar scenario](https://jsfiddle.net/68015uek/) and if I do 1000 iterations it's twice as fast using an object. But it's still just 1ms total, and much more iterations than this code is expected to do.

If I run it with 10 iterations, there is no difference at all most of the time.

GP-Dan-Tovbein: I know that 10 iterations does not affect if you do it in a regular scenario, but we're using react and it's a completely different scenario since aren't ten, are more than ten, since for every change you do on every field it's invoked the `render`method and that is the main performance issue

GP-Dan-Tovbein: I typed (very slowly) these values, about 70 characters, and for each letter that I type is invoked the `render`method, so, in theory if we implement a `filter`method and taking into account that we have 10 fields that needs to be filtered, we will iterate ~700 times that function. it's a big difference if you compare to an `object`.

assets/src/styles/blocks/ShareButtons/ShareButtonsStyles.scss
```diff
@@ -0,0 +1,50 @@
+@mixin sharedStyles($color) {
+  background: $color;
+
+  &:hover,
```
Inwerpsel, line 4: We should never use the same styles for hover, focus and active, these need to be distinct for a11y.

In this case you can definitely leave out the `focus` state, these should get a focus border from our general styles.

assets/src/styles/blocks.scss
```diff
@@ -9,6 +9,7 @@
 @import "blocks/SplitTwoColumns";
 @import "blocks/Submenu";
 @import "blocks/Cookies";
+@import "blocks/ShareButtons/ShareButtonsStyles";
```
Inwerpsel, line 12: This needs to be set up with code splitting if we want to merge it as a beta block. Otherwise we'll be sending this style to everyone for no reason.

assets/src/editorIndex.js
```diff
@@ -21,6 +21,7 @@ import { SubPagesBlock } from './blocks/SubPages/SubPagesBlock';
 import { blockEditorValidation } from './BlockEditorValidation';
 import { ENFormBlock } from './blocks/OldENForm/ENFormBlock';
 import { registerGuestBookBlock } from './blocks/GuestBook/GuestBookBlock';
+import { registerBlock as registerShareButtonsBlock } from './blocks/ShareButtons/ShareButtonsBlock';
```
Inwerpsel, line 24: Also for the JS, we should avoid adding these beta blocks to the main index JS for performance.

assets/src/styles/blocks/ShareButtons/ShareButtonsStyles.scss
GP-Dan-Tovbein: This is how is currently implemented
https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/components/_share-buttons.scss#L25-L62

Inwerpsel: Ah I didn't realize these styles were copied over. Why do we even need to copy them? Can't they use the existing styles?

Inwerpsel: In which case of course we need to fix this issue in the existing styles :smile:

assets/src/styles/blocks/ShareButtons/ShareButtonsStyles.scss
```diff
@@ -0,0 +1,54 @@
+@import "../../master-theme/assets/src/scss/base/colors";
+
+@mixin sharedStyles($color) {
+  background: $color;
+
+  &:hover {
+    background: darken($color, 20%);
+  }
+
+  &:active {
```
Inwerpsel, line 10: Can you remove this from the PR to keep things simple? We don't have a spec for the active state yet.

assets/src/styles/blocks/ShareButtons/ShareButtonsStyles.scss
GP-Dan-Tovbein: Don't you think it's better if we differentiate the styles from the block?

Inwerpsel: Do we have a specification somewhere saying we want a different style of share buttons for this block?

Inwerpsel: I assume that if we don't have a specification, it makes most sense to use the existing style. If we find in the future we want a different style here, we can at that point add separate CSS.

assets/src/editorIndex.js
```diff
@@ -47,3 +47,4 @@ replaceTaxonomyTermSelectors();
 setupCustomSidebar();
 setUpCssVariables();
 blockEditorValidation();
+registerShareButtonsBlock();
```
Inwerpsel, line 50: I think this also needs to be removed?

GP-Dan-Tovbein: Good catch!

assets/src/styles/blocks/ShareButtons/ShareButtonsStyles.scss
GP-Dan-Tovbein: Ok! I'm going to use the current sharbuttons styles, so, should I remove `active` from [here](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/components/_share-buttons.scss#L25-L62), shouldn't I?

Inwerpsel: You can also remove `focus` there, as it already has a focus outline from the general button styles.

assets/src/blocks/ShareButtons/ShareButtonsBlock.js
```diff
@@ -0,0 +1,83 @@
+import { ShareButtonsEditor } from './ShareButtonsEditor';
+import { ShareButtonsFrontend } from './ShareButtonsFrontend';
+
+const { registerBlockType } = wp.blocks;
+const { __ } = wp.i18n;
+
+const BLOCK_NAME = 'planet4-blocks/share-buttons';
+
+export const registerBlock = () => {
+  return registerBlockType(BLOCK_NAME, {
```
Inwerpsel, line 10: You can directly include this in the editor script, like this block does https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/17f159f23992cb58f164fd9c198a1ab07389b64d/assets/src/blocks/Accordion/AccordionEditorScript.js#L35-L56

Inwerpsel (approved):
Looks mostly good. I think we can improve the sidebar code a bit in a follow up, a lot of the logic for each provider's sidebar entry is currently repeated. But probably best to do that when we have a better idea of what to show in the sidebar.

GP-Dan-Tovbein:
I realize after I merged it that I didn't squash all the commits.

PR #763, opened 2022-01-19
Turn sharebuttons to a block using array

GP-Dan-Tovbein:


Inwerpsel:
@GP-Dan-Tovbein I'll close this as we included it in https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/758

PR #764, opened 2022-01-20
PLANET-6496 Happy Point: Add option for code snippet embed

sagarsdeshmukh:
See https://jira.greenpeace.org/browse/PLANET-6496

---

Added the "embed code" options to happy point block editor. -

image

### DIFF
--- a/assets/src/blocks/Happypoint/HappypointFrontend.js
+++ b/assets/src/blocks/Happypoint/HappypointFrontend.js
@@ -6,6 +6,8 @@ export const HappypointFrontend = ({
   mailing_list_iframe,
   iframe_url,
   id,
+  use_embed_code,
+  embed_code,
   className,
 }) => {
   const { imageData } = useHappypointImageData(id);
@@ -16,7 +18,8 @@ export const HappypointFrontend = ({
     background_alt,
     background_sizes,
     default_image,
-    engaging_network_id
+    engaging_network_id,
+    default_embed_code
   } = imageData;
 
   const imgProps = {
@@ -34,6 +37,7 @@ export const HappypointFrontend = ({
   }
 
   const url = iframe_url || engaging_network_id;
+  const happypoint_embed_code = embed_code || default_embed_code;
 
   return (
     <section className={`block block-footer block-wide happy-point-block-wrap ${className ?? ''}`}>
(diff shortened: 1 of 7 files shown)


assets/src/blocks/Happypoint/HappypointFrontend.js
```diff
@@ -6,6 +6,8 @@ export const HappypointFrontend = ({
   mailing_list_iframe,
   iframe_url,
   id,
+  use_embed_code,
+  embed_code,
   className,
```
comzeradd, line 11: Should we also fetch `happy_point_embed_code` from settings just below? Otherwise only the embed code from the sidebar is being used.

assets/src/blocks/Happypoint/HappypointFrontend.js
```diff
@@ -42,19 +44,24 @@ export const HappypointFrontend = ({
...
-                allowtransparency="true"
-                frameBorder={0}
-                scrolling="no"
-                width="100%"
-                loading="lazy"
-              />
-            </div>
-          )}
+          {(use_embed_code && embed_code) ?
+            <div dangerouslySetInnerHTML={{ __html: embed_code }} /> :
```
comzeradd, line 48: That works fine for simple html code, but it doesn't seem to run any `<script>` tags that are part of the code.

assets/src/blocks/Happypoint/HappypointFrontend.js
sagarsdeshmukh: yes indeed, was missed from my head.  PR updated :)

assets/src/blocks/Happypoint/HappypointFrontend.js
```diff
@@ -42,19 +48,25 @@ export const HappypointFrontend = ({
...
-                allowtransparency="true"
-                frameBorder={0}
-                scrolling="no"
-                width="100%"
-                loading="lazy"
-              />
-            </div>
-          )}
+          {use_embed_code
+            ? <div className="col-md-10" id="happy-point" />
```
comzeradd, line 52: Can you also add an `mt-5` class here? Just to make the iframe have some spacing from the top.

comzeradd (approved):
:100:

PR #766, opened 2022-01-20
PLANET-6603 Use default Cookies copy in Cookies block

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6603

Related master-theme PR: https://github.com/greenpeace/planet4-master-theme/pull/1586

**Note**: the code is very repetitive, we could create some components to improve that, but maybe in a separate PR for simplicity's sake 🙂 

### Testing

You can check the new settings on local (`Planet 4 > Cookies`), or on the [neptune test instance](https://www-dev.greenpeace.org/test-neptune/wp-admin/admin.php?page=planet4_settings_cookies_text). Then once you set some values there, you can add a Cookies block to a page and see those values appear as default in the various fields. I've also added some reset buttons for the fields, so that editors can easily revert custom values to the default ones defined in the settings.

### DIFF
--- a/assets/src/blocks/Cookies/CookiesBlock.js
+++ b/assets/src/blocks/Cookies/CookiesBlock.js
@@ -24,27 +24,21 @@ export class CookiesBlock {
         },
         necessary_cookies_name: {
           type: 'string',
-          default: ''
         },
         necessary_cookies_description: {
           type: 'string',
-          default: ''
         },
         all_cookies_name: {
           type: 'string',
-          default: ''
         },
         all_cookies_description: {
           type: 'string',
-          default: ''
         },
         analytical_cookies_name: {
           type: 'string',
-          default: ''
         },
         analytical_cookies_description: {
           type: 'string',
-          default: ''
         },
       },
       edit: CookiesEditor,
--- a/assets/src/blocks/Cookies/CookiesFieldResetButton.js
+++ b/assets/src/blocks/Cookies/CookiesFieldResetButton.js
@@ -1,0 +1,27 @@
+const { __ } = wp.i18n;
+
+import { Tooltip } from '@wordpress/components';
+
+const COOKIES_DEFAULT_COPY = window.p4bk_vars.cookies_default_copy || {};
+
+export const CookiesFieldResetButton = ({ fieldName, toAttribute, currentValue }) => {
+  const defaultValue = COOKIES_DEFAULT_COPY[fieldName] || '';
+
+  if (!currentValue || !defaultValue || currentValue === defaultValue) {
+    return null;
+  }
+
+  return (
+    <div className='field-reset-button'>
+      <span
+        className='cta'
+        onClick={() => toAttribute(fieldName)(undefined)}
+      >
+        {__('Use default value', 'planet4-blocks-backend')}
+      </span>
+      <Tooltip text={__('This value is defined in the settings, in Planet 4 > Cookies', 'planet4-blocks-backend')}>
+        <span className='info'>i</span>
+      </Tooltip>
+    </div>
+  );
+}
--- a/assets/src/blocks/Cookies/CookiesFrontend.js
+++ b/assets/src/blocks/Cookies/CookiesFrontend.js
@@ -1,7 +1,7 @@
-import { Fragment } from '@wordpress/element';
 import { FrontendRichText } from '../../components/FrontendRichText/FrontendRichText';
 import { removeCookie, useCookie, writeCookie } from './useCookie';
 import { useState, useEffect } from 'react';
+import { CookiesFieldResetButton } from './CookiesFieldResetButton';
 
 const { __ } = wp.i18n;
 const CONSENT_COOKIE = 'greenpeace';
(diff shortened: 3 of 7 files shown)


planet-4:

:watch: 2022.01.21 17:12:59

assets/src/blocks/Cookies/CookiesBlock.js
```diff
@@ -24,27 +26,27 @@ export class CookiesBlock {
         },
         necessary_cookies_name: {
           type: 'string',
-          default: ''
+          default: DEFAULT_COOKIES_COPY.necessary_cookies_label
```
Inwerpsel, line 29: This default value is only used inside the editor, it's not saved or used anywhere else. The [back end still has different default values](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/c0d7ca2d7e01d895656f476bdaeba3fc15e294a1/classes/blocks/class-cookies.php#L32-L63).

mleray: Updated! 🙂

assets/src/blocks/Cookies/CookiesBlock.js
```diff
@@ -1,13 +1,16 @@
 import { CookiesEditor } from './CookiesEditor.js';
+const { __ } = wp.i18n;
```
Inwerpsel, line 2: Can we avoid adding this dependency to the front end? It adds a lot of scripts in order to perform the translation. In a few other blocks we added the translations to a back end render to avoid this.

mleray: I guess we could just remove it tbh, it wasn't there before and nobody complained about it 😛 I've added it to be consistent with other blocks, but now that I think about it "Cookies" might be an international term and therefore not need translations 🤔 We could also fix this by splitting the blocks' files into backend and frontend scripts, but that would probably be better done in a separate PR?

assets/src/styles/blocks/Cookies.scss
```diff
@@ -6,4 +6,36 @@
       font-size: $font-size-md;
     }
   }
+
+  .cookies-checkbox-description.rich-text {
```
Inwerpsel, line 10: Shouldn't this be for the editor only?

assets/src/blocks/Cookies/CookiesBlock.js
Inwerpsel: Sounds good, let's just remove the translation there.

Inwerpsel: +1 on splitting back end and front end JS in a separate PR

assets/src/blocks/Cookies/CookiesFrontend.js
```diff
@@ -114,194 +114,235 @@ export const CookiesFrontend = props => {
...
           editable={isEditing}
           allowedFormats={['core/bold', 'core/italic']}
         />
-      </Fragment>
-      }
-    </section>
-  </Fragment>;
+        }
+        {(isEditing || (necessary_cookies_name && necessary_cookies_description)) &&
+          <>
```
Inwerpsel, line 148: Don't we preferably use `<Fragment>`? :thinking:

assets/src/blocks/Cookies/CookiesFieldResetButton.js
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!currentValue || !defaultValue || currentValue === defaultValue) {
+    return null;
+  }
+
+  return (
+    <div className='field-reset-button'>
+      <span
+        className='cta'
+        onClick={() => toAttribute(fieldName)(defaultValue)}
```
Inwerpsel, line 18: I think we actually need to unset the attribute here. Otherwise it will not change along when the option changes.

assets/src/styles/blocks/Cookies.scss
mleray: Indeed, but this block doesn't have split files yet

assets/src/blocks/Cookies/CookiesFrontend.js
mleray: I thought `<>` was the same? But without using an extra import statement

assets/src/styles/blocks/Cookies.scss
Inwerpsel: We can just add the file in this PR, it's an easy way to improve front end perofrmance a bit.

mleray: Good point! Updated 🙂

assets/src/blocks/Cookies/CookiesFrontend.js
Inwerpsel: It's does the same but I vaguely remember disucssing it and deciding to use `<Fragment>`, though it never made it into code style config :thinking:

assets/src/blocks/Cookies/CookiesFieldResetButton.js
mleray: Updated! I've used `undefined` and it seems to work as expected 🙂

PR #768, opened 2022-01-26
Add a block that outputs everything around content

Inwerpsel:
* This block renders uses templates located in the theme repo. It allows
us to specify templates for server generated pages without rewriting
everything first.

### DIFF
--- a/classes/blocks/class-framework.php
+++ b/classes/blocks/class-framework.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace P4GBKS\Blocks;
+
+use P4\MasterTheme\Context;
+use P4\MasterTheme\Post;
+
+/**
+ * A block with the base template from master theme, inside which you can put post content.
+ */
+class Framework extends Base_Block {
+
+	public function __construct() {
+		register_block_type( 'planet4-blocks/framework', [
+			'editor_script'   => 'planet4-blocks',
+			'attributes'      => [],
+			'render_callback' => [ $this, 'render_it' ],
+		] );
+	}
+
+	public function prepare_data( $fields ): array {
+		return [];
+	}
+
+	public function render_it( $attributes, $content, $block ) {
+		$inner_blocks_html = '';
+		foreach ( $block->inner_blocks as $inner_block ) {
+			$inner_blocks_html .= $inner_block->render();
+		}
+
+		\Timber::$locations = P4GBKS_PLUGIN_DIR . '../../themes/planet4-master-theme/templates';
+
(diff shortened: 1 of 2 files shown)


classes/blocks/class-framework.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace P4GBKS\Blocks;
+
+use P4\MasterTheme\Context;
+use P4\MasterTheme\Post;
+
+/**
+ * A block with the base template from master theme, inside which you can put post content.
+ */
+class Framework extends Base_Block {
```
Inwerpsel, line 11: Best name I could come up with for what the block does, but open to alternative.

I do not in any way mean that this is some kind of a framework to build on. In fact it's just a hack to allow us to use the new `block-templates` directory for auto generated pages like the author page.

Without this "hack" we would need to convert all our site's elements to blocks before we can start using it.

classes/blocks/class-framework.php
```diff
@@ -0,0 +1,56 @@
...
+			'attributes'      => [],
+			'render_callback' => [ $this, 'render_it' ],
+		] );
+	}
+
+	public function prepare_data( $fields ): array {
+		return [];
+	}
+
+	public function render_it( $attributes, $content, $block ) {
```
Inwerpsel, line 25: Name is not `render` as a quick fix to make it stop complaining the base classes method has different params.

classes/blocks/class-framework.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * A block with the base template from master theme, inside which you can put post content.
+ */
+class Framework extends Base_Block {
+
+	public function __construct() {
+		register_block_type( 'planet4-blocks/framework', [
+			'editor_script'   => 'planet4-blocks',
+			'attributes'      => [],
+			'render_callback' => [ $this, 'render_it' ],
```
Inwerpsel, line 17: Name is not render as a quick fix to make it stop complaining the base classes method has different params.

planet-4:

:new_moon: [venus](https://www-dev.greenpeace.org/test-venus) | [admin](https://www-dev.greenpeace.org/test-venus/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-venus/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-venus) | [composer-local.json](https://github.com/greenpeace/planet4-test-venus/blob/main/composer-local.json)

:watch: 2022.01.27 10:12:37

Inwerpsel:
See https://github.com/greenpeace/planet4-master-theme/pull/1620#issuecomment-1072308923

PR #771, opened 2022-01-27
PLANET-6599: Fix arrow issue from Carousel Take Action block

GP-Dan-Tovbein:
[Ticket](https://jira.greenpeace.org/browse/PLANET-6599)

[Demo page](https://www-dev.greenpeace.org/test-proteus/carousel-layout/)

**Problem**
The arrows are not visible when the Covers has set just 4 cards.

**What is fixed**
Set to 3 rows to Carousel layout the same as it's set to `grid`.

### DIFF
--- a/assets/src/blocks/Covers/useCovers.js
+++ b/assets/src/blocks/Covers/useCovers.js
@@ -16,15 +16,13 @@ export const useCovers = ({ post_types, tags, cover_type, initialRowsLimit, post
   const [error, setError] = useState(null);
 
   const updateRowCoversAmount = () => {
-    switch (cover_type) {
-    case COVERS_TYPES.campaign:
+    if(cover_type === COVERS_TYPES.campaign || cover_type === COVERS_TYPES.takeAction) {
       if (layout === COVERS_LAYOUTS.carousel) {
         setAmountOfCoversPerRow(3);
       } else {
         setAmountOfCoversPerRow(isSmallWindow() ? 2 : 3);
       }
-      break;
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.01.27 20:35:43

assets/src/blocks/Covers/useCovers.js
```diff
@@ -15,14 +15,21 @@ export const useCovers = ({ post_types, tags, cover_type, initialRowsLimit, post
...
+
   const updateRowCoversAmount = () => {
     switch (cover_type) {
     case COVERS_TYPES.campaign:
-      if (layout === COVERS_LAYOUTS.carousel) {
-        setAmountOfCoversPerRow(3);
-      } else {
-        setAmountOfCoversPerRow(isSmallWindow() ? 2 : 3);
-      }
+      setCoversPerRow();
```
mleray, line 32: Since this is now identical to the `takeAction` style, and that means there are only 2 cases left, I don't think a `switch` structure is still needed, we can probably just use an `if ... else` statement? 🤔 Or at least we can group the 2 identical cases together, such as:
```
case COVERS_TYPES.campaign:
case COVERS_TYPES.takeAction:
```

GP-Dan-Tovbein: Yep, It's a good suggestion!

mleray (approved):
Looks good to me 👍 Next step would be to solve the covers' width issue in XXL screens!

PR #785, opened 2022-02-11
Dev toggle for all blocks

Inwerpsel:
I also cleaned up some obsolete fallback code and simplified the logic a bit.

The fallback code in case the type is none of the specified was throwing warnings. Also the logic made it use only core blocks for those cases. One of them is reusable blocks, which can be created with p4 blocks but then when edited individually would not allow to place them.

To improve this I made it fallback to the page blocks.

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -178,11 +178,12 @@ const BETA_CAMPAIGN_BLOCK_TYPES = [
  * Allowed block types based on post type
  *
  * @param array  $allowed_block_types array of allowed block types.
- * @param object $post current post.
+ * @param object $context Current editor context.
  *
  * @return array of all blocks allowed.
  */
-function set_allowed_block_types( $allowed_block_types, $post ) {
+function set_allowed_block_types( $allowed_block_types, $context ) {
+	$post_type = $context->post ? $context->post->post_type : null;
 	// https://github.com/WordPress/gutenberg/blob/trunk/lib/blocks.php.
 	$wordpress_blocks = [
 		'core/block',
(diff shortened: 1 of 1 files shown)


planet4-gutenberg-blocks.php
```diff
@@ -251,35 +255,12 @@ function set_allowed_block_types( $allowed_block_types, $post ) {
...
 	];
 
-	$allowed_p4_block_types = $all_allowed_p4_block_types[ $post->post_type ];
-
-	if ( empty( $allowed_p4_block_types ) ) {
-		return $wordpress_blocks;
-	}
-
-	$allowed_block_types = array_merge( $wordpress_blocks, $allowed_p4_block_types );
+	$allowed_p4_block_types = $all_allowed_p4_block_types[ $post_type ] ?? $all_allowed_p4_block_types[ 'page' ];
```
Inwerpsel, line 258: Making this fall back to page block types fixes the issue of p4 blocks not being allow in the reusable block editor.

However there's also a chance this results in the block types being allow unintentionally on custom post types created by NROs. If this could be a problem I can preserve the exact previous behavior of only allowing core blocks.

planet4-gutenberg-blocks.php
```diff
@@ -251,35 +255,12 @@ function set_allowed_block_types( $allowed_block_types, $post ) {
...
 	];
 
-	$allowed_p4_block_types = $all_allowed_p4_block_types[ $post->post_type ];
-
-	if ( empty( $allowed_p4_block_types ) ) {
-		return $wordpress_blocks;
-	}
-
-	$allowed_block_types = array_merge( $wordpress_blocks, $allowed_p4_block_types );
+	$allowed_p4_block_types = $all_allowed_p4_block_types[ $post_type ] ?? $all_allowed_p4_block_types['page'];
```
Inwerpsel, line 258: @comzeradd This previously resulted in warnings if there's no post or no known post type.
Afaik this only occurs for "reusable blocks". Those actually can contain blocks from all post types, depending on where they are saved as a reusable block. However we default to adding no P4 blocks. So this prevented you from putting new P4 blocks in an existing reusable block.

comzeradd: Yeah, that's a good fix. So far the editors had to create Reusable blocks only in the Page editor just to be able to add P4 blocks.

comzeradd (approved):

PR #789, opened 2022-02-14
PLANET-6616 Create new block pattern: side image with text and CTA

mleray:
### Description

See https://jira.greenpeace.org/browse/PLANET-6616

### Testing

You can test the new pattern on [this page](https://www-dev.greenpeace.org/test-atlas/side-image-with-text-and-cta-block-pattern/) for example, make sure it looks good in the frontend in all screen sizes. In the backend, you can add find this pattern in the `Patterns > Planet 4` category and check the UX/UI for editing it.


planet-4:

:new_moon: [atlas](https://www-dev.greenpeace.org/test-atlas) | [admin](https://www-dev.greenpeace.org/test-atlas/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-atlas/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-atlas) | [composer-local.json](https://github.com/greenpeace/planet4-test-atlas/blob/main/composer-local.json)

:watch: 2022.02.14 16:58:42

classes/patterns/class-sideimagewithtextandcta.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Side image with text and CTA pattern constructor.
+	 */
+	public function __construct() {
+		if ( ! function_exists( 'register_block_pattern' ) ) {
+			return;
+		}
+
+		register_block_pattern(
+			self::PATTERN_NAME,
```
Inwerpsel, line 34: Ideally we have this logic somewhere else, so that we don't need to add more than just the config for every pattern.

WP core does this by including files that return an array.

https://github.com/WordPress/WordPress/blob/8475350f0e408d00832caedcfba340bed96f6266/wp-includes/block-patterns.php#L21-L37

Inwerpsel: You could add an abstract class with `get_name`, and `get_config` abstract methods? It can include shared logic, and also allows IDEs to auto generate most of the class (at least in PHPStorm, apparently [you need to pay for a premium plugin](https://stackoverflow.com/questions/54883976/vs-code-auto-generate-interface-methods-in-php-class) to have this feature :cry: )

Like we use for [migration scripts](https://github.com/greenpeace/planet4-master-theme/blob/71347e57b872b5c27aaf8cc0c946790cd455f37a/src/MigrationScript.php#L13-L30).

Inwerpsel (approved):

classes/patterns/class-sideimagewithtextandcta.php
Inwerpsel: We can also do this once we add the next pattern.

assets/src/styles/patterns.scss
```diff
@@ -0,0 +1,2 @@
+// P4 Patterns
+@import "patterns/SideImageWithTextAndCta";
```
Inwerpsel, line 2: Patterns are nothing more than copying pasting a collection of blocks inside a post. So it doesn't really make sense to attach custom styles to them.

If the pattern needs to make use of styles, we have the following options:
* Set style options on the blocks before exporting them
* Register a block style and add it to the pattern before exporting to code.

classes/patterns/class-sideimagewithtextandcta.php
```diff
@@ -0,0 +1,62 @@
...
+			[
+				'title'      => __( 'Side image with text and CTA', 'planet4-blocks-backend' ),
+				'keywords'   => [ 'cta', 'side', 'image' ],
+				'categories' => [ 'planet4' ],
+				'content'    => '
+					<!-- wp:media-text -->
```
Inwerpsel, line 40: Probably best to add the class name property here too. Apparently WP will add it automatically when inserting and there's no console error/warning. But still seems best to add it in a valid state already.

Inwerpsel: Screenshot from 2022-02-17 10-22-30

Inwerpsel, line 40: WP seems to add the missing `className` attribute here after inserting. But probably best to add it here anyway.

mleray: Updated!

classes/patterns/class-sideimagewithtextandcta.php
mleray: [I've added a base class for the patterns](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/789/commits/befa20906c7efde0765ddb1dae7edd2c9c5e78f1), similar to what we do for blocks... Anything else you think we should put there? 🤔

assets/src/styles/blocks.scss
```diff
@@ -27,6 +27,7 @@
 @import "blocks/core-overrides/BlockSpacing";
 @import "blocks/core-overrides/Separator";
 @import "blocks/core-overrides/Table";
+@import "blocks/core-overrides/MediaAndText";
```
Inwerpsel, line 30: We should probably use the block name for directories, [like Gutenberg does](https://github.com/WordPress/gutenberg/tree/8536f6f1aa1df9daa98193964aa8e62cc196cb7a/packages/block-library/src/media-text).

classes/patterns/class-base-pattern.php
```diff
@@ -0,0 +1,25 @@
...
+namespace P4GBKS\Patterns;
+
+/**
+ * Class Base_Pattern
+ *
+ * @package P4GBKS\Patterns
+ */
+abstract class Base_Pattern {
+
+	public const NAMESPACE = 'planet4-patterns';
```
Inwerpsel, line 17: It's not needed to define this namespace separately, as it will only be used inside the name. So it's a lot easier to just put the namespace in the name.

classes/patterns/class-base-pattern.php
```diff
@@ -0,0 +1,25 @@
...
+
+	/**
+	 * Returns the pattern name with its namespace prefix, e.g.: planet4-patterns/side-image-with-text-and-cta.
+	 */
+	public static function get_full_pattern_name() {
+		return static::NAMESPACE . '/' . static::PATTERN_NAME;
```
Inwerpsel, line 23: The problem with constants is that you can't require them to be present with an abstract class. Abstract functions on the other hand will need to have an implementation in any extending class, or the code won't even run (or pass CI tests).

In other words, if we create 2 abstract functions (`get_name` and `get_attributes`) is a way of saying ""Every pattern needs to have a name and an array of attributes".

assets/src/styles/blocks.scss
mleray: In our case it's not a directory, it's a file, not sure if that changes anything for your comment 😬 Do you mean a structure like `blocks/core-overrides/media-text/style.scss`, so having a folder per core block?

Inwerpsel: No same folder structure, just using the block actual name for the file name `_media-text.scss` instead of creating our own identifier for it.

mleray: Sure, updated!

assets/src/styles/blocks/core-overrides/_media-text.scss
```diff
@@ -0,0 +1,35 @@
...
+
+    .wp-block-button,
+    .wp-block-button a {
+      width: 100%;
+    }
+  }
+
+  @media (min-width: 601px) {
+    &.has-media-on-the-right .wp-block-media-text__content {
+      padding-inline-start: 0;
```
Inwerpsel, line 16: Why not `$sp-0`? :grin:

Inwerpsel (dismissed):
Looks good. I'm not too sure about the customizations on the core media and text block, but I guess we can remove them once we find the core styles actually make more sense :shrug:

assets/src/styles/blocks/core-overrides/_media-text.scss
mleray: I'm pretty sure 0 will remain 0 if we decide to change the grid in the future 😉

classes/patterns/class-base-pattern.php
Inwerpsel: Can you remove this namespace and include it directly in the block name?

Also, similar to with block name prefixes, we could go with a short prefix `p4/`.

Inwerpsel (changes requested):

assets/src/styles/blocks/core-overrides/_media-text.scss
Inwerpsel: It makes equally much sense to me as with other numbers :shrug:

Inwerpsel (approved):

PR #795, opened 2022-02-21
PLANET-6612 Whitelist the articles block in Post

sagarsdeshmukh:
The 'Include Articles in Post' setting is moved as articles block at the end of post

Ref.: https://jira.greenpeace.org/browse/PLANET-6612

### DIFF
--- a/classes/blocks/class-articles.php
+++ b/classes/blocks/class-articles.php
@@ -314,15 +314,23 @@ class Articles extends Base_Block {
 		// Get all posts with arguments.
 		$args = self::DEFAULT_POST_ARGS;
 
+		// For post page block, so current main post will exclude.
+		if ( $exclude_post_id ) {
+			$args['post__not_in'] = [ $exclude_post_id ];
+
+			// Get current post tags & categories, if not pass.
+			if ( ! $tags ) {
+				$tags = wp_get_post_tags( $exclude_post_id, [ 'fields' => 'ids' ] );
+			}
+			if ( ! $category_id_array ) {
+				$category_id_array = wp_get_post_categories( $exclude_post_id, [ 'fields' => 'ids' ] );
+			}
+		}
+
 		if ( true !== $ignore_categories ) {
 			if ( $category_id_array ) {
 				$args['category__in'] = $category_id_array;
 			}
-		}
-
-		// For post page block so current main post will exclude.
-		if ( $exclude_post_id ) {
-			$args['post__not_in'] = [ $exclude_post_id ];
 		}
 
 		// Add filter for p4-page-type terms.
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -108,6 +108,7 @@ require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
 const POST_BLOCK_TYPES = [
 	'planet4-blocks/accordion',
+	'planet4-blocks/articles',
 	'planet4-blocks/counter',
 	'planet4-blocks/gallery',
 	'planet4-blocks/social-media',
--- a/assets/src/blocks/Articles/ArticlesEditor.js
+++ b/assets/src/blocks/Articles/ArticlesEditor.js
@@ -20,7 +20,8 @@ const { __ } = wp.i18n;
 
 const TextControl = withCharacterCounter(BaseTextControl);
 
-const renderEdit = (attributes, toAttribute) => {
+const renderEdit = ( { attributes, postType }, toAttribute) => {
+  const showPostTagHelp = postType === 'post' ? 'block' : 'none';
 
   return (
     <Fragment>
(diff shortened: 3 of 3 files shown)


planet-4:

:new_moon: [jupiter](https://www-dev.greenpeace.org/test-jupiter) | [admin](https://www-dev.greenpeace.org/test-jupiter/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-jupiter/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-jupiter) | [composer-local.json](https://github.com/greenpeace/planet4-test-jupiter/blob/main/composer-local.json)

:watch: 2022.02.21 12:57:27

planet4-gutenberg-blocks.php
```diff
@@ -116,6 +116,7 @@
 	'planet4-blocks/timeline',
 	'leadin/hubspot-form-block',
 	'gravityforms/form',
+	'planet4-blocks/articles',
```
comzeradd, line 119: Since the rest of the items here are ordered alphabetically, maybe we can move articles to its alphabetical place.

comzeradd (approved):

classes/blocks/class-articles.php
```diff
@@ -314,17 +314,25 @@ function ( $tag ) {
 		// Get all posts with arguments.
 		$args = self::DEFAULT_POST_ARGS;
 
+		// For post page block, so current main post will exclude.
+		if ( $exclude_post_id ) {
+			$args['post__not_in'] = [ $exclude_post_id ];
+
+			// Get current post tags & categories, if not pass.
+			if ( !$tags ) {
```
sagarsdeshmukh, line 322: These are the new code changes after last review.
Reference : https://github.com/greenpeace/planet4-master-theme/pull/1645/files#r823477472

cc: @comzeradd

classes/blocks/class-articles.php
```diff
@@ -314,17 +314,25 @@ function ( $tag ) {
 		// Get all posts with arguments.
 		$args = self::DEFAULT_POST_ARGS;
 
+		// For post page block, so current main post will exclude.
+		if ( $exclude_post_id ) {
+			$args['post__not_in'] = [ $exclude_post_id ];
+
+			// Get current post tags & categories, if not pass.
+			if ( ! $tags ) {
```
sagarsdeshmukh, line 322: Hi @Inwerpsel , I like your idea of adding a new field(checkbox of "inherit filters from post") to offer a more control on articles block on post pagetype(cause this issue only happens on post & not on other post types), but without adding this field the current implementation work as it was working before this change.

I just thought I have to highlight that, if editor wants to add a custom taxonomy filters, then they can easily able to add it, and if we introduce a new field here it will bit confuse editor about behaviour of articles block with and without this field, also I don't think the unchecked checkbox of "inherit filters from post" will be more useful cause mostly the goal is to display the related articles in articles block, so if the taxonomy filter(tags & categories) is empty in articles block then the articles block shows the posts by filtering on current post taxonomy will be the default behaviour.

I may miss some aspects of your suggestion, so I thought I put my thoughts here.

cc: @comzeradd , @suzi-greenpeace

Inwerpsel: @sagarsdeshmukh To be sure I understand correctly, you suggest to make inheriting the filters implicit each time an Articles block has no filters?

That makes some sense, as there's likely no use case for Articles with no filters. Would still be nice to include a notice in the editor about this behavior.

For example if a post has 2 categories and 3 tags. If an editor would then add 1 tag manually to the block, it would stop using the other tags and categories and only filter the manual query. That's fine but if we don't describe the behavior it could be confusing.

I assume we also need to account for that in the block editor code? I mean, with the current behavior the post's current tags and categories are set on the block by our back end, so only when you view the post. If we want consistent behavior in the editor, it would need to update the block every time the post's tags or categories are changed.

sagarsdeshmukh: Hi @Inwerpsel 
> you suggest to make inheriting the filters implicit each time an Articles block has no filters?

yes,

> That makes some sense, as there's likely no use case for Articles with no filters. Would still be nice to include a notice in the editor about this behavior.

The note added as below,
image


> For example if a post has 2 categories and 3 tags. If an editor would then add 1 tag manually to the block, it would 
stop using the other tags **and** categories and only filter the manual query.

Here, it only stop using post tags and use custom tag in filter, but categories filter will still remain as post categories if available.

> I assume we also need to account for that in the block editor code? I mean, with the current behavior the post's current tags and categories are set on the block by our back end, so only when you view the post. If we want consistent behavior in the editor, it would need to update the block every time the post's tags or categories are changed.

That's true, but categories & tags are not frequently changes, but if changes the articles block not refresh automatically till editor refresh the page.

Inwerpsel: > but if changes the articles block not refresh automatically till editor refresh the page

Seems acceptable for a first iteration. But should be possible to fix that using WP's JS data API, we could do a follow up for that.

For example using
```js
const tags = useSelect( select => {
  return select('core/editor').getCurrentPost().tags;
}
```

We could force the block to rerender

PR #796, opened 2022-02-22
Ensure class is not needed to use style

Inwerpsel:
* The code can just as well target img, which allows to use the same CSS
for the Query Loop block. That block doesn't allow putting a class on
the img it outputs.

For now it's the only modification that was needed.


planet-4:

:watch: 2022.02.22 11:56:54

mleray (approved):
Looks good to me 👍

PR #804, opened 2022-03-04
Deprecate tying JS logic to style

Inwerpsel:
* WordPress's block style API is only intended to toggle a class. The
problem is because it's changed outside the component render flow,
currently implementations use "useEffect" directly inside of the
component. This causes tons of issues with WordPress block previews,
which get stuck in a render loop.

I'm still checking if it makes sense to deprecate the entire function. The current usages mostly use `useEffect`, which causes the problems. The block renders an additional time, and the first time it has an invalid state (the block already has the new class but the effect still needs to run).

### DIFF
--- a/assets/src/blocks/getStyleFromClassName.js
+++ b/assets/src/blocks/getStyleFromClassName.js
@@ -1,4 +1,7 @@
 /**
+ * @deprecated Blocks should not attach any further logic to WP's block style mechanism. It leads
+ * to infinite loops, performance issues, buggy editor previews, and more.
+ *
  * Get the name of the block style used from the classname string
  * given by Gutenberg
  *


planet-4:

:watch: 2022.03.04 12:23:40

Inwerpsel:
> Looks good to me +1 Do we need to update the blocks that are currently using this?

I'm still not sure if the whole function should be deprecated. It's mostly a problem if we use it with `useEffect`. If it's used directly without the effect delay it doesn't pose a problem by itself.

I guess it's not needed to immediately fix it in other blocks, unless there's a current need to do it, like a bug. It's not a small change for some blocks, so it may not be worth the effort.

PR #807, opened 2022-03-09
PLANET-6654 Fix editor previews

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6654

The block editor now puts each block preview into an iframe, which it [provides with the editor stylesheets](https://github.com/WordPress/gutenberg/blob/8536f6f1aa1df9daa98193964aa8e62cc196cb7a/packages/block-editor/src/components/block-preview/auto.js#L73-L74).

WordPress provides these to the editor in PHP, by checking each block type for registered styles/editor styles.
https://github.com/WordPress/gutenberg/blob/8536f6f1aa1df9daa98193964aa8e62cc196cb7a/lib/client-assets.php#L627-L639
(It's not the actual same file in WordPress but similar enough).

Additionally WP now allows you to specify the screen size which it will apply to the preview iframe with scaling so that it takes up the same amount of sidebar space.

**Needs to be merged after https://github.com/greenpeace/planet4-master-theme/pull/1643**

TODO:
- [x] ~~Check the previous issue of registering a style causing it to be served on the front end even if the block is not on the page. Likely solved in core by now.~~ Should be fine, we don't register the script if it's not on the page so passing the handle doesn't change the front end.
- [x] Apply to other blocks, maybe reduce duplication.
- [x] Determine the right viewport width for each block type

### Findings

- Preview accuracy significantly improves in most cases.
- The font size gets a bit small, but hard to get around this. I used the minimum preview screen size that resulted in the same layout as our old previews. The reason our previews had larger text in 5.8 with this layout is actually because they're inaccurate.
- Some of our blocks have wildly different styles on mobile, to the point that we should really be showing a preview of the style on each screen for someone to make a reasonable choice. We could look at ways to make this more intuitive.
- Columns block previews not worth investing more time in because people should use the core Columns block, which does not even need a preview.
- Similar for Covers.

### DIFF
--- a/assets/src/blocks/Columns/example.js
+++ b/assets/src/blocks/Columns/example.js
@@ -1,4 +1,5 @@
 export const example = {
+  viewportWidth: 992,
   attributes: {
     isExample: true,
     exampleColumns: [
--- a/assets/src/blocks/Covers/example.js
+++ b/assets/src/blocks/Covers/example.js
@@ -1,4 +1,5 @@
 export const example = {
+  viewportWidth: 992,
   attributes: {
     isExample: true,
     title: 'Covers block',
--- a/assets/src/blocks/Submenu/example.js
+++ b/assets/src/blocks/Submenu/example.js
@@ -1,4 +1,5 @@
 export const example = {
+  viewportWidth: 992,
   attributes: {
     isExample: true,
     title: 'The block title',
--- a/assets/src/styles/editorStyle.scss
+++ b/assets/src/styles/editorStyle.scss
@@ -1,16 +1,8 @@
-@import "~bootstrap/scss/bootstrap-reboot";
-@import "~bootstrap/scss/bootstrap-grid";
-@import "~bootstrap/scss/carousel";
-@import "~bootstrap/scss/tooltip";
-@import "~bootstrap/scss/buttons";
-
 @import "master-theme/assets/src/scss/base/variables";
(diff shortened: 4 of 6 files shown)


planet-4:

:new_moon: [sinope](https://www-dev.greenpeace.org/test-sinope) | [admin](https://www-dev.greenpeace.org/test-sinope/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-sinope/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-sinope) | [composer-local.json](https://github.com/greenpeace/planet4-test-sinope/blob/main/composer-local.json)

:watch: 2022.03.09 13:31:39

lithrel (approved):
:+1: 

> Similar for Covers.

I assume you meant "for Submenus" because it also doesn't have its specific style in preview example when I test it ?

Inwerpsel:
> +1
> 
> > Similar for Covers.
> 
> I assume you meant "for Submenus" because it also doesn't have its specific style in preview example when I test it ?

No I meant similar like Columns it's probably not worth it to get the previews 100% right at this moment, and there's also core block alternatives that don't suffer from this issue.

I also noticed the 1st and 2nd style now look the same. But if I use those styles they also produce the same output in both the editor and the front end, at least for the content I have locally.
Screenshot from 2022-03-16 19-45-27

PR #808, opened 2022-03-11
PLANET-6519 Remove post types and tags from Articles block

mleray:
### Description

See [PLANET-6519](https://jira.greenpeace.org/browse/PLANET-6519)

### Testing


planet-4:

:watch: 2022.03.11 10:25:31

assets/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -36,19 +36,15 @@ export class ArticlesBlock {
           type: 'array',
           default: []
         },
-        post_types: {
```
Inwerpsel, line 39: I don't think we're supposed to remove the `post_types` filter. [The ticket](https://jira.greenpeace.org/browse/PLANET-6519) only mentions

> Remove the post type and tags from the block.

mleray: Yes that is something that was agreed upon on Slack with Mag directly, but as we discussed it's not a good idea to remove it right now without checking existing blocks, it requires more thinking through and therefore [I've added it back](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/808/commits/1101b43aa4ec1fc8bc4dd34618a3cd3e470211d3).

assets/src/blocks/Articles/ArticlesBlock.js
```diff
@@ -36,19 +36,15 @@ export class ArticlesBlock {
...
-          default: __('Load more', 'planet4-blocks')
+          default: __('See all articles', 'planet4-blocks')
         },
-        read_more_link: {
+        see_all_link: {
           type: 'string',
           default: ''
         },
-        button_link_new_tab: {
+        see_all_new_tab: {
```
Inwerpsel, line 51: We could avoid changing this attribute name, if we'd want to preserve the old values. `button_link_new_tab` is still an accurate name.

If we don't want to preserve values, renaming like this makes sense to give everyone the default value..

Probably same for `read_more_text`.

assets/src/blocks/Articles/ArticlePreview.js
```diff
@@ -97,95 +81,77 @@ export class ArticlePreview extends Component {
...
-        page_type_link,
         post_title,
         post_excerpt,
         date_formatted,
         reading_time,
+        category,
       }
     } = this.props;
 
-    const articleClassName = tags.reduce((classname, tag) => classname + ` tag-${tag.slug}`, 'article-list-item');
```
Inwerpsel, line 111: It's possible this is needed even when we don't output tags ourselves. Someone can create a class for a specific post and target any HTML element inside the `article`.

There's no harm in keeping these classes, they don't conflict with the style changes of this PR.

Inwerpsel: It's depended on here, for example https://github.com/greenpeace/planet4-child-theme-switzerland/blob/3c27e85d541ed6f12d375658be96e5595d16a73f/src/scss/variables/_colors.scss#L27

Inwerpsel: Screenshot from 2022-03-25 16-52-00
Screenshot from 2022-03-25 16-51-31

mleray: This is no longer relevant for this PR which focuses on the `See all` functionality

assets/src/blocks/Articles/ArticlesBlock.js
mleray: Yeah I thought about that, but I think that we do want to change the values, since it's no longer about "loading more" articles but it leads to the search page 🤔

mleray: I guess the new tab one could eventually be kept the same, I mostly changed it for consistency

Inwerpsel: Yeah a new name is better, it allows us to preserve the old functionality alongside the new link in the very first iteration.

mleray: It's not the case with the current implementation though, how would that work since both buttons are at the same place in the block? 🤔

PR #810, opened 2022-03-14
PLANET-6658 Fix WPML RTL mayhem

Inwerpsel:
Ref: https://jira.greenpeace.org/browse/PLANET-6658

* Remove "rtl_fix" which only makes things worse.
* Remove filter that caused WPML to call `switch_lang` when it's not
supposed to.

Likely WPML added `url_to_postid` assuming it would only be called in a context where it's safe to switch the language.

Maybe this used to be the case for the REST API. However WordPress calls the REST API "internally" to prevent the editor being delayed by an actual HTTP request. As a result WPML's `url_to_postid` filter will run on every revision that contains any embed block.

After some investigation I found `url_to_postid` is a super old part of WordPress and only used in a few leftover edge cases.

The only usage that could "break" if we remove WPML's filter is when you [embed a post from the same site](https://developer.wordpress.org/reference/classes/wp_oembed_controller/get_item/), which probably no one does anyway.

This branch is pinned on the `ummah` instance ([example page](https://www-dev.ummah4earth.org/wp-admin/post.php?post=1146&action=edit) that had the issue, you can compare with [same page on production](https://ummah4earth.org/wp-admin/post.php?post=1146&action=edit)).

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -397,3 +397,16 @@ function p4_blocks_en_forms_twig_filters( $twig ) {
 */
 P4GBKS\Loader::get_instance();
 \P4GBKS\Rest\Rest_Api::add_endpoints();
+
+$remove_rtl_fix = function (): void {
+	global $sitepress;
+	// This RTL fix does not seem a good idea. Probably it was a bad attempt at solving the issues `url_to_postid`
+	// creates.
+	remove_action( 'wp_head', [ $sitepress, 'rtl_fix' ] );
+	remove_action( 'admin_print_styles', [ $sitepress, 'rtl_fix' ] );
+
+	// This caused `switch_lang` to get called. As a result the RTL fix messed up.
+	remove_filter( 'url_to_postid', [ $sitepress, 'url_to_postid' ] );
+};
+$remove_rtl_fix();
+add_action( 'wpml_after_startup', $remove_rtl_fix, 10, 0 );


planet-4:

:watch: 2022.03.14 13:06:42

docs/WPML-url_to_postid.trace.MD
```diff
@@ -0,0 +1,33 @@
+```
```
Inwerpsel, line 1: I'll probably remove this later but it's nice to put line comments on this here.

docs/WPML-url_to_postid.trace.MD
```diff
@@ -0,0 +1,33 @@
...
+plugin.php:189, apply_filters()
+class-wp-rest-revisions-controller.php:606, WP_REST_Revisions_Controller->prepare_item_for_response()
+class-wp-rest-autosaves-controller.php:408, WP_REST_Autosaves_Controller->prepare_item_for_response()
+class-wp-rest-autosaves-controller.php:300, WP_REST_Autosaves_Controller->get_items()
+class-wp-rest-server.php:1141, WP_REST_Server->respond_to_request()
+class-wp-rest-server.php:988, WP_REST_Server->dispatch()
+rest-api.php:511, rest_do_request()
+rest-api.php:2860, rest_preload_api_request()
+block-editor.php:501, array_reduce()
+block-editor.php:501, block_editor_rest_api_preload()
```
Inwerpsel, line 30: Problem 1: WP's has an "optimization" to preload post revisions before sending the editor HTML. It renders the content of every revision, regardless of how many total.

Inwerpsel:
> Looks good 👍🏻
> 
> Before merging, maybe add the commit hash (or a link to this PR) at the top of the `.md` file, so it's easier to find the reference.

I'll remove the MD file, it was mostly to assist the PR and should not be relevant anymore after addressing the bug.

PR #811, opened 2022-03-15
PLANET-6638: Issues block pattern

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6638

Design: [Figma](https://www.figma.com/file/99X4IXpkQIgNKtKvHEyI1P/IA-%26-nav_MVP-prototype?node-id=31%3A219)

[Demo page](https://www-dev.greenpeace.org/test-sinope/issues-pattern)

**Testing**
Go to the instance > Create new page > Click on "+" > Search Issues pattern

**There are some UI pattern's limitations, ~that is why we add some styles through an external CSS file (see the class deep-dive-topic class)~**

### DIFF
--- a/admin/js/editor.js
+++ b/admin/js/editor.js
@@ -32,4 +32,33 @@ wp.domReady(() => {
   ];
 
   registerBlockStyle('core/button', styles);
+
+  ['core/media-text', 'core/group', 'core/column'].forEach(
+    block => registerBlockStyle(block, [
+      {
+        name: 'small-padding',
+        label: __('Small padding', 'planet4-blocks-backend'),
+      }, {
+        name: 'medium-padding',
+        label: __('Medium padding', 'planet4-blocks-backend'),
+      }, {
+        name: 'large-padding',
+        label: __('Large padding', 'planet4-blocks-backend'),
+      },
+    ]),
+  );
+
+  ['core/group'].forEach(block => {
+    registerBlockStyle(block, {
+      name: 'space-evenly',
+      label: __('Space evenly', 'planet4-blocks-backend'),
+    });
+  });
+
+  ['core/paragraph'].forEach(block => {
+    registerBlockStyle(block, {
+      name: 'roboto-font-family',
+      label: __('Roboto font family', 'planet4-blocks-backend'),
+    });
+  });
 });
--- a/classes/patterns/class-issues.php
+++ b/classes/patterns/class-issues.php
@@ -29,8 +29,8 @@ class Issues extends Block_Pattern {
 	public static function get_media_text_template(): string {
 
 		return '
-			<!-- wp:media-text {"mediaWidth":20,"mediaSizeSlug":"thumbnail","isStackedOnMobile":false,"verticalAlignment":"center","imageFill":false,"backgroundColor":"white","className":"is-style-large-padding"} -->
-			<div class="wp-block-media-text alignwide is-vertically-aligned-center has-white-background-color is-style-large-padding has-background" style="grid-template-columns:20% auto">
+			<!-- wp:media-text {"mediaWidth":15,"mediaSizeSlug":"thumbnail","isStackedOnMobile":false,"verticalAlignment":"center","imageFill":false,"backgroundColor":"white","className":"is-style-large-padding"} -->
+			<div class="wp-block-media-text alignwide is-vertically-aligned-center has-white-background-color is-style-large-padding has-background" style="grid-template-columns:15% auto">
 			<figure class="wp-block-media-text__media"></figure>
 			<div class="wp-block-media-text__content">
 			<!-- wp:paragraph {"align":"left","placeholder":"' . __( 'Enter text', 'planet4-blocks-backend' ) . '","style":{"typography":{"fontSize":"1rem","fontStyle":"normal","fontWeight":"700"}},"className":"is-style-roboto-font-family"} -->
--- a/classes/patterns/class-issues.php
+++ b/classes/patterns/class-issues.php
@@ -1,0 +1,102 @@
+<?php
+/**
+ * Issues class.
+ *
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Patterns;
+
+/**
+ * Issues pattern includes:
+ * Column, Image, Heading, Paragraph, Media & Text.
+ *
+ * @package P4GBKS\Patterns
+ */
+class Issues extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/issues';
+	}
+
+	/**
+	 * Returns the template for one media-text.
+	 */
+	public static function get_media_text_template(): string {
+
+		return '
+			<!-- wp:media-text {"mediaWidth":20,"mediaSizeSlug":"thumbnail","isStackedOnMobile":false,"verticalAlignment":"center","imageFill":false,"backgroundColor":"white","className":"is-style-large-padding"} -->
+			<div class="wp-block-media-text alignwide is-vertically-aligned-center has-white-background-color is-style-large-padding has-background" style="grid-template-columns:20% auto">
+			<figure class="wp-block-media-text__media"></figure>
+			<div class="wp-block-media-text__content">
+			<!-- wp:paragraph {"align":"left","placeholder":"' . __( 'Enter text', 'planet4-blocks-backend' ) . '","style":{"typography":{"fontSize":"1rem","fontStyle":"normal","fontWeight":"700"}},"className":"is-style-roboto-font-family"} -->
+			<p class="has-text-align-left is-style-roboto-font-family" style="font-size:1rem;font-style:normal;font-weight:700"></p>
+			<!-- /wp:paragraph --></div></div>
+			<!-- /wp:media-text -->
+		';
+	}
+
+	/**
+	 * Returns the pattern config.
+	 */
+	public static function get_config(): array {
+		return [
+			'title'      => __( 'Issues', 'planet4-blocks-backend' ),
+			'categories' => [ 'planet4' ],
+			'content'    => '
+				<!-- wp:group {"align":"full","backgroundColor":"grey-10"} -->
+				<div class="wp-block-group alignfull has-grey-10-background-color has-background">
+
+					<!-- wp:group {"className":"container"} -->
+					<div class="wp-block-group container">
+
+						<!-- wp:spacer {"height":"60px"} -->
+						<div style="height:60px" aria-hidden="true" class="wp-block-spacer"></div>
+						<!-- /wp:spacer -->
+
+						<!-- wp:heading {"textAlign":"center","level":1, "placeholder":"' . __( 'Enter title', 'planet4-blocks-backend' ) . '"} -->
+						<h1 class="has-text-align-center"></h1>
+						<!-- /wp:heading -->
+
+						<!-- wp:paragraph {"align":"center", "placeholder":"' . __( 'Enter description', 'planet4-blocks-backend' ) . '"} -->
+						<p class="has-text-align-center"></p>
+						<!-- /wp:paragraph -->
+
+						<!-- wp:spacer {"height":"37px"} -->
+						<div style="height:37px" aria-hidden="true" class="wp-block-spacer"></div>
+						<!-- /wp:spacer -->
+
+						<!-- wp:group {"className":"is-style-space-evenly","layout":{"type":"flex","allowOrientation":false}} -->
+						<div class="wp-block-group is-style-space-evenly">
+						' . self::get_media_text_template() . '
+						' . self::get_media_text_template() . '
+						' . self::get_media_text_template() . '
+						' . self::get_media_text_template() . '
+						</div>
+						<!-- /wp:group -->
+
+						<!-- wp:spacer {"height":"37px"} -->
+						<div style="height:37px" aria-hidden="true" class="wp-block-spacer"></div>
+						<!-- /wp:spacer -->
+
+						<!-- wp:buttons {"placeholder":"' . __( 'Enter text', 'planet4-blocks-backend' ) . '","layout":{"type":"flex","justifyContent":"center"}} -->
+						<div class="wp-block-buttons">
+						<!-- wp:button -->
+						<div class="wp-block-button is-style-secondary"><a class="wp-block-button__link"></a></div>
+						<!-- /wp:button --></div>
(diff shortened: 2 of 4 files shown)


planet-4:

:new_moon: [sinope](https://www-dev.greenpeace.org/test-sinope) | [admin](https://www-dev.greenpeace.org/test-sinope/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-sinope/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-sinope) | [composer-local.json](https://github.com/greenpeace/planet4-test-sinope/blob/main/composer-local.json)

:watch: 2022.03.23 11:32:45

classes/patterns/class-issues.php
```diff
@@ -0,0 +1,176 @@
...
+						<div class="wp-block-group">
+							<!-- wp:columns -->
+							<div class="wp-block-columns">
+								<!-- wp:column -->
+								<div class="wp-block-column">
+									<!-- wp:media-text -->
+										<div class="wp-block-media-text alignwide is-stacked-on-mobile">
+										<figure class="wp-block-media-text__media"></figure>
+										<div class="wp-block-media-text__content">
+										<!-- wp:paragraph {"placeholder":"Description","style":{"typography":{"fontSize":"17px"}}} -->
```
mleray, line 55: We should add translation strings for placeholders, you can see [how I added them for the Highlighted CTA pattern](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/814/files#diff-5c0d4b33eaa8eb18a44ba1f1e0fb04b29c8c3492b6b4dfdc95f5dd4edeaf8c26R44)

mleray:
Do we need to initialise it with so many columns? 🤔 On my local it looks a bit weird in the editor:

Screenshot 2022-03-24 at 11 45 48

assets/src/styles/blocks/Issues.scss
```diff
@@ -0,0 +1,107 @@
+.issues-block-pattern {
```
mleray, line 1: Ideally, we should avoid adding custom classes to the patterns, because the second it gets added to a page it's basically just a block... We can style the root block (for example the WP core `Columns` block, I created a [ticket](https://jira.greenpeace.org/browse/PLANET-6683) to get design recommendations for it), but then it will be styled for all occurrences of said block, not just the one used in the pattern. It's actually better for consistency too.
There are some native ways to style the blocks that are used in the pattern:
- You can usually define color and background color in the attributes when initialising a block (such as `Media & Text`).
- For spacing, it's nice to use the `Spacer` block or [Bootstrap classes](https://getbootstrap.com/docs/5.0/utilities/spacing/) that we already use in other places.

It is less customisable than adding a custom class, so it means sometimes we'll have to compromise, but I think it makes more sense for patterns 🙂

assets/src/styles/blocks/DeepDiveTopic.scss
```diff
@@ -0,0 +1,25 @@
+.deep-dive-topic {
```
GP-Dan-Tovbein, line 1: I need to add this styles since the editor does not provide some of them, like `box-shadow`, `min-height`, among others.

c.c. @mleray

comzeradd: We can leave this customizations out of this implementation for now, if there is no other way of applying them, and let the design team know. Most of these patterns are under assessment as pilots will start using them. So better not add any specific css classes just yet.

classes/patterns/class-issues.php
```diff
@@ -0,0 +1,144 @@
...
+		return 'p4/issues';
+	}
+
+	/**
+	 * Returns the pattern config.
+	 */
+	public static function get_config(): array {
+		return [
+			'title'      => __( 'Issues', 'planet4-blocks-backend' ),
+			'keywords'   => [ 'column', 'image', 'heading', 'paragraph', 'media', 'text' ],
```
mleray, line 32: Those keywords need to be translated 🤔 Also, who decided on which ones to add?

classes/patterns/class-issues.php
```diff
@@ -0,0 +1,144 @@
...
+				<!-- wp:columns -->
+				<div class="wp-block-columns">
+
+				<!-- wp:column {"verticalAlignment":"center","width":"25%","backgroundColor":"white"} -->
+				<div class="wp-block-column is-vertically-aligned-center has-white-background-color has-background" style="flex-basis:25%"><!-- wp:group {"layout":{"wideSize":"90%","contentSize":"90%"}} -->
+				<div class="wp-block-group"><!-- wp:spacer {"height":"16px"} -->
+				<div style="height:16px" aria-hidden="true" class="wp-block-spacer"></div>
+				<!-- /wp:spacer -->
+				<!-- wp:media-text {"mediaWidth":20,"isStackedOnMobile":false,"imageFill":false} -->
+				<div class="wp-block-media-text alignwide" style="grid-template-columns:20% auto"><figure class="wp-block-media-text__media"></figure><div class="wp-block-media-text__content"><!-- wp:paragraph {"align":"left","placeholder":"' . __( 'Enter text', 'planet4-blocks-backend' ) . '","style":{"typography":{"fontSize":"17px","fontStyle":"normal","fontWeight":"700"}},"textColor":"grey-80"} -->
```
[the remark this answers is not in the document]
mleray: Also, maybe you don't know this (I sure didn't 😬) but you can actually use `rem` values instead of `px` for the `fontSize` attribute, I'm not sure about this specific case but it might be better?

classes/patterns/class-issues.php
GP-Dan-Tovbein: > who decided on which ones to add?
Nobody, I've added them since these are the used blocks 😶

mleray: Or we could ask the design team if they have any UX/UI recommendations for this

mleray (changes requested):
It seems that there is an error somewhere in the markup, when testing on local I get this when inserting a new Issues pattern:
Screenshot 2022-04-13 at 14 22 23

classes/patterns/class-issues.php
```diff
@@ -0,0 +1,118 @@
...
+
+	/**
+	 * Returns the pattern config.
+	 */
+	public static function get_config(): array {
+		return [
+			'title'      => __( 'Issues', 'planet4-blocks-backend' ),
+			'categories' => [ 'planet4' ],
+			'content'    => '
+				<!-- wp:group {"backgroundColor":"grey-10","className":"block-wide"} -->
```
mleray, line 63: We should now use the core class `alignfull` class instead of `block-wide` (see this [previously merged PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/820)). I think that you should also put the `block` class since [it adds our spacing values](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/layout/_blocks.scss) 🤔

classes/patterns/class-issues.php
```diff
@@ -0,0 +1,115 @@
...
+
+	/**
+	 * Returns the pattern config.
+	 */
+	public static function get_config(): array {
+		return [
+			'title'      => __( 'Issues', 'planet4-blocks-backend' ),
+			'categories' => [ 'planet4' ],
+			'content'    => '
+				<!-- wp:group {"backgroundColor":"grey-10","className":"alignfull"} -->
```
Inwerpsel, line 60: Can we do this without needing 2 groups? The block's complexity would be already a bit more manageable with 1 level less.

GP-Dan-Tovbein: The first group is the wide block that needs to be fit to the screen's width. And the second group is `container` class.

Inwerpsel: Yeah this one makes sense and doesn't complicated it that much.

It's mostly a problem in the row of items below, as there the complexity is repeated for every element. That makes operations like selecting the right block in the editor very annoying. The paragraphs are currently 8 levels deep. While for a user it seems like the block would only need 3 or 4.
Screenshot from 2022-04-19 22-02-00

Inwerpsel: I created the [same block using the "Row" block](https://www-dev.greenpeace.org/gutenberg/wp-admin/post.php?post=27114&action=edit) (which is enabled because on the gutenberg instance we don't severely limit the core block functionality).
With just 1 additional CSS rule, it looks identical to what's currently in the PR.
Screenshot from 2022-04-19 22-02-29

GP-Dan-Tovbein: I completely agree to use `Row` instead of `Column`, in fact that was my initial plan. But the problem was that we needed -at least- a bit customization and it wasn't what we wanted to achieve. 

If we're fine to go with a bit customization then I suggest to use this [deleted file](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/811/commits/a917eaf2373318fa8cacc1cc27aac36a7c4696b6#diff-ac498fa0f16734d2b3a1af452585c6f5828b5612676c2e2b6c6dddc0409781b5) (with a couple of changes) and then provide to the NRO, the CSS in case they need it just by adding the classname through the editor panel (additional settings). 

What do you think?

Inwerpsel: > But the problem was that we needed -at least- a bit customization and it wasn't what we wanted to achieve.

Can you elaborate? Which customization, and why "wasn't it what we wanted to achieve"?

Inwerpsel: > If we're fine to go with a bit customization then I suggest to use this deleted file (with a couple of changes) and then provide to the NRO, the CSS in case they need it just by adding the classname through the editor panel

In my example you see I only had to add 1 CSS rule to set a out padding on the media and text block. That seems a pretty common use case, so we could register it as a block style. Then we could use it every time instead of nesting blocks to achieve padding.

Inwerpsel:
@GP-Dan-Tovbein Isn't each item supposed to link to a page?

classes/patterns/class-issues.php
```diff
@@ -0,0 +1,115 @@
...
+				<div class="wp-block-group alignfull has-grey-10-background-color has-background">
```
Inwerpsel, line 61: A separate issue as the previous line, but putting it here to start a new thread.

We should use
```
"align":"full",
```
in the attributes instead of only putting the `.alignfull` class on it. It uses the editor's alignment control for that.
Screenshot from 2022-04-20 13-46-21

classes/patterns/class-issues.php
Inwerpsel, line 60: I [added a few classes here as a style to media and text](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/832) (and some other core types that don't use any styles yet), and other blocks. https://www-dev.greenpeace.org/test-venus/issues-pattern-with-padding-style/

https://user-images.githubusercontent.com/7604138/164225326-d87c7a86-1456-405f-91ca-cbfbfe9aa7b2.mp4

GP-Dan-Tovbein: > Can you elaborate? Which customization, and why "wasn't it what we wanted to achieve"?
For example `box-shadow`, `border-radius`, `font-family`, `margin` between image and text. In fact, I had to add those styles to [this GPBR page](https://www-dev.greenpeace.org/brasil/trabalhe-conosco-2/) because they requested that.

comzeradd: > With just 1 additional CSS rule, it looks identical to what's currently in the PR.

Wouldn't that require a css class explicitly for this pattern?

Inwerpsel: No, we could register the style on the Media and Text block type, it currently doesn't have any block styles. It's often the case when you use this block that it needs padding. For now we had to add additional blocks just to get this padding right. With a style it's just 1 click in the sidebar. See also [this message](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/811#discussion_r854049107).

GP-Dan-Tovbein:
> @GP-Dan-Tovbein Isn't each item supposed to link to a page?

Nope, they aren't links.

classes/patterns/class-issues.php
GP-Dan-Tovbein: I really like to register styles, but I think that the padding classes [you added here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/832) are very specific to this block. What do you think?

Inwerpsel: You mean this style is not reusable in other cases? :thinking:

Inwerpsel: You can register the same style for multiple blocks. I added that to the PR now, should be on the test instance soon.

GP-Dan-Tovbein: > You mean this style is not reusable in other cases?

Yes, but if you want to apply more padding to another `media-text` you should override it.

Inwerpsel: Ah you mean the exact padding values. Well we do have a sort of design system, so we can create some values that are likely to work in a lot of cases.

For now I kept it simple in the PR and set a single padding of 8 / 16 / 32 px. I also included `align-self: stretch` as it seemed safe to do. How do you suggest making this more generic?

GP-Dan-Tovbein: Let's continue the padding conversation on this [PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/832/files)

admin/js/editor.js
```diff
@@ -32,4 +32,33 @@ wp.domReady(() => {
...
+  );
+
+  ['core/group'].forEach(block => {
+    registerBlockStyle(block, {
+      name: 'space-evenly',
+      label: __('Space evenly', 'planet4-blocks-backend'),
+    });
+  });
+
+  ['core/paragraph'].forEach(block => {
```
GP-Dan-Tovbein, line 58: This style allows to editor to choose the `Roboto` font family. Also defined through the design system.

Inwerpsel (approved):

comzeradd:
@dantovbein make sure to squash commits before merge.

GP-Dan-Tovbein:
@Inwerpsel I've updated the style and it's now rendered correctly through small and medium screens.
Screenshot 2022-05-04 at 18 59 12
Screenshot 2022-05-04 at 18 59 33

Screenshot 2022-05-04 at 18 59 03

GP-Dan-Tovbein:
@Inwerpsel FYI, Houssam has approved the UAT! So, let me know if this PR is fine again!

Inwerpsel (approved):

PR #814, opened 2022-03-17
PLANET-6668 Implement Highlighted CTA pattern

mleray:
### Description

See [PLANET-6668](https://jira.greenpeace.org/browse/PLANET-6668)

### Testing


planet-4:

:watch: 2022.03.17 10:28:20

classes/class-loader.php
```diff
@@ -633,6 +633,11 @@ public function set_color_palette() {
 					'slug'  => 'yellow',
 					'color' => '#ffd204',
 				],
+				[
```
mleray, line 636: It looks like SonarCloud is complaining about these lines being duplicated, but I don't see a way around it 🤔

sagarsdeshmukh (approved):
Tested on test instance & local, LGTM 👍

classes/patterns/class-highlightedcta.php
```diff
@@ -0,0 +1,65 @@
...
+						<!-- wp:column -->
+							<div class="wp-block-column">
+								<!-- wp:image {"align":"center"} -->
+									<div class="wp-block-image">
+										<figure class="aligncenter">
+											<img alt=""/>
+										</figure>
+									</div>
+								<!-- /wp:image -->
+								<!-- wp:heading {"placeholder":"Enter text","align":"center","className":"has-text-align-center","level":3} -->
```
sagarsdeshmukh, line 44: Can we translate this string.

sagarsdeshmukh (approved):
PR Looks good to me 👍 we need to squash the small commit, also need to fix the commitlint error, otherwise all good :)

classes/patterns/class-highlightedcta.php
mleray: Good catch! [Fixed](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/814/commits/d9ec3d7ac1c22e91dd7cdf39432ec4417af4f8f6)

mleray: I've actually updated the [commit](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/814/commits/93a42c78f80a2784c8cd07a30026875bc89f79db) in order to add translations to the first pattern that we created, the Side image with CTA and text, because we forgot them there 😬 I've also removed the placeholder from the Button there, because this block actually doesn't have a placeholder attribute at the moment.

classes/patterns/class-sideimagewithtextandcta.php
```diff
@@ -28,22 +28,21 @@ public static function get_name(): string {
...
+							<!-- wp:heading {"placeholder":"' . __( 'Enter title', 'planet4-blocks-backend' ) . '"} -->
 								<h2></h2>
 							<!-- /wp:heading -->
-							<!-- wp:paragraph {"placeholder":"Enter description"} -->
+							<!-- wp:paragraph {"placeholder":"' . __( 'Enter description', 'planet4-blocks-backend' ) . '"} -->
 								<p></p>
 							<!-- /wp:paragraph -->
 							<!-- wp:buttons -->
 								<div class="wp-block-buttons">
-									<!-- wp:button {"placeholder":"Enter CTA text"} /-->
```
mleray, line 46: It looks like the Button core block currently doesn't support placeholders, adding it doesn't do anything, so I've removed it

classes/class-loader.php
Inwerpsel: We can now define color palette's in `theme.json`, I think it should have the same result as defining them here.

Inwerpsel: https://developer.wordpress.org/themes/advanced-topics/theme-json/#preset-values

Inwerpsel: Not sure what that means for translations, though.

classes/patterns/class-highlightedcta.php
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	/**
+	 * Returns the pattern config.
+	 */
+	public static function get_config(): array {
+		return [
+			'title'      => __( 'Highlighted CTA', 'planet4-blocks-backend' ),
+			'categories' => [ 'planet4' ],
+			'content'    => '
```
Inwerpsel, line 32: It's probably nicer if we create a function that returns only content, since these patterns get quite complex, defining them inside a key of an array is kind of inconvenient.
Maybe it's even better to put the content in a `.html` file :thinking:

PR #815, opened 2022-03-22
PLANET-6667 Create Reality Check block pattern

mleray:
### Description

See [PLANET-6667](https://jira.greenpeace.org/browse/PLANET-6667), the [example page](https://www-dev.greenpeace.org/gutenberg/block-patterns/) and the [designs](https://www.figma.com/file/leueo1LMrlPabJkZYs2NgW/IA%26nav_Final-mockups?node-id=1%3A1650).
This is basically a Columns block with specific blocks per column (image, title and description). Houssam asked that we initialise the pattern with 3 columns so that is how it's implemented here, but editors can easily remove and/or duplicate columns.

### Testing

You can either test the new pattern on local, or on [this page](https://www-dev.greenpeace.org/test-titan/reality-check-tests/) I created for UAT. Also make sure to try adding a new one in the backend so that you can see the initial state.

### DIFF
--- a/classes/patterns/class-realitycheck.php
+++ b/classes/patterns/class-realitycheck.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * Reality Check pattern class.
+ *
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Patterns;
+
+/**
+ * Class Reality Check.
+ *
+ * @package P4GBKS\Patterns
+ */
+class RealityCheck extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/reality-check';
+	}
+
+	/**
+	 * Returns the template for one column.
+	 */
+	public static function get_column_template(): string {
+		return '
+			<!-- wp:column -->
+				<div class="wp-block-column">
+					<!-- wp:image {"align":"center","className":"mb-0 force-no-lightbox"} -->
+						<div class="wp-block-image mb-0 force-no-lightbox">
+							<figure class="aligncenter">
+								<img alt="" />
+							</figure>
+						</div>
+					<!-- /wp:image -->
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2022.03.23 09:23:57

classes/patterns/class-realitycheck.php
```diff
@@ -0,0 +1,72 @@
...
+				<div class="wp-block-column">
+					<!-- wp:image {"align":"center","className":"mb-0 force-no-lightbox"} -->
+						<div class="wp-block-image mb-0 force-no-lightbox">
+							<figure class="aligncenter">
+								<img alt="" />
+							</figure>
+						</div>
+					<!-- /wp:image -->
+					<!-- wp:heading {"style":{"typography":{"fontSize":"64px"}},"align":"center","className":"has-text-align-center mb-0","placeholder":"' . __( 'Enter title', 'planet4-blocks-backend' ) . '"} -->
+						<h2 style="font-size:64px;" class="has-text-align-center mb-0"></h2>
```
lithrel, line 40: I'm always confused, should we express font-size in `rem` or `px` ?

lithrel (approved):
:+1:

classes/patterns/class-realitycheck.php
mleray: We use a mix of both in our codebase, so it is confusing 😬 But I think ideally we should use [our fonts](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/base/_fonts.scss) (and therefore `rem`) when possible... Here however we are using WordPress' native functionality and that uses pixels, so we don't really have a choice:

Screenshot 2022-03-31 at 10 25 44

lithrel: Hooo didn't realize that, good to know thanks :+1:

mleray: Nevermind what I said, turns out you can select a unit there so I will ask the design team what they would rather use! Good catch @lithrel 👍

mleray: @lithrel I've updated the pattern to use `rem` after discussing with Mag

PR #828, opened 2022-04-14
PLANET-6534 Create Deep Dive block pattern

mleray:
### Description

See [PLANET-6534](https://jira.greenpeace.org/browse/PLANET-6534)
I've yet to figure out how to implement [this transition](https://www-dev.greenpeace.org/gutenberg/deep-dive-topics-links/) requested by the design team, maybe it will be for a second iteration?

### Testing

On local, you'll need to set master-theme to the `new-pattern-placeholders` branch to see the correct image placeholders both in the editor and in the frontend ([corresponding PR](https://github.com/greenpeace/planet4-master-theme/pull/1717)).
Otherwise you can test the new pattern on [this page](https://www-dev.greenpeace.org/test-nix/deep-dive-block-pattern/) for example 🙂

### DIFF
--- a/classes/patterns/class-deepdive.php
+++ b/classes/patterns/class-deepdive.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Deep Dive pattern class.
+ *
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Patterns;
+
+/**
+ * Class Deep Dive.
+ *
+ * @package P4GBKS\Patterns
+ */
+class DeepDive extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/deep-dive';
+	}
+
+	/**
+	 * Returns the template for one column.
+	 */
+	public static function get_column_template(): string {
+		return '
+			<!-- wp:column -->
+				<div class="wp-block-column">
+					<!-- wp:group {"className":"group-stretched-link"} -->
+						<div class="wp-block-group group-stretched-link">
+							<!-- wp:image {"align":"center","className":"force-no-lightbox force-no-caption is-style-rounded-180"} -->
+								<div class="wp-block-image is-style-rounded-180 force-no-lightbox force-no-caption">
+									<figure class="aligncenter">
+										<img src="' . esc_url( get_template_directory_uri() ) . '/images/placeholders/placeholder-180x180.jpg" alt="' . __( 'Default image', 'planet4-blocks-backend' ) . '" />
+									</figure>
+								</div>
+							<!-- /wp:image -->
+							<!-- wp:heading {"level":5,"style":{"typography":{"fontSize":"1rem"}},"textAlign":"center","className":"is-style-chevron","placeholder":"' . __( 'Enter topic', 'planet4-blocks-backend' ) . '"} -->
+								<h5 style="font-size:1rem;" class="has-text-align-center is-style-chevron"></h5>
+							<!-- /wp:heading -->
+							<!-- wp:spacer {"height":"16px"} -->
+								<div style="height:16px" aria-hidden="true" class="wp-block-spacer"></div>
+							<!-- /wp:spacer -->
+						</div>
+					<!-- /wp:group -->
+				</div>
+			<!-- /wp:column -->
+		';
+	}
+
+	/**
+	 * Returns the pattern config.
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2022.04.15 07:54:10

classes/patterns/class-deepdive.php
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @package P4GBKS\Patterns
+ */
+class DeepDive extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/reality-check';
```
GP-Dan-Tovbein, line 22: Wrong name here

classes/patterns/class-deepdive.php
```diff
@@ -0,0 +1,84 @@
...
+						</div>
+					<!-- /wp:group -->
+				</div>
+			<!-- /wp:column -->
+		';
+	}
+
+	/**
+	 * Returns the pattern config.
+	 * We start with 3 columns, but editors can easily remove and/or duplicate them.
```
GP-Dan-Tovbein, line 56: > We start with 3 columns, but editors can easily remove and/or duplicate them.
But it renders 4

classes/patterns/class-deepdive.php
```diff
@@ -0,0 +1,84 @@
...
+	public static function get_column_template(): string {
+		return '
+			<!-- wp:column -->
+				<div class="wp-block-column">
+					<!-- wp:group {"className":"group-stretched-link"} -->
+						<div class="wp-block-group group-stretched-link">
+							<!-- wp:image {"align":"center","className":"force-no-lightbox is-style-rounded"} -->
+								<div class="wp-block-image is-style-rounded force-no-lightbox">
+									<figure class="aligncenter">
+										<img <img src="' . esc_url( get_template_directory_uri() ) . '/images/placeholders/placeholder-180x180.jpg" alt="' . __( 'Default image', 'planet4-blocks-backend' ) . '" />
```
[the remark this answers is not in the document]
mleray: Good catch, fixed! Crazy it didn't complain in the editor 🤯

classes/patterns/class-deepdive.php
```diff
@@ -0,0 +1,84 @@
...
+	 */
+	public static function get_config(): array {
+		return [
+			'title'      => __( 'Deep Dive', 'planet4-blocks-backend' ),
+			'categories' => [ 'planet4' ],
+			'content'    => '
+				<!-- wp:group {"className":"block","align":"full","backgroundColor":"grey-05"} -->
+					<div class="wp-block-group alignfull block has-grey-05-background-color has-background">
+						<!-- wp:spacer {"height":"24px"} -->
+							<div style="height:24px" aria-hidden="true" class="wp-block-spacer"></div>
```
GP-Dan-Tovbein, line 66: What do you think to add a `wp:group` with a `container` class and wrap these [elements](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/828/files#diff-ad859d4c0b075e75438c0346f87e068c4d21863211f43c267e78bb185e9ded21R65-R78) like [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/patterns/class-issues.php#L54-L55).

Doing this it will avoid to set 2 `container` classes, to the `wp:heading` and the `wp:columns`, which does the same behavior. In addition, it will improve the performance since you'll be using only the `@media-query` once instead of 2 (`wp:heading` and the `wp:columns`).

mleray: Yeah I had considered it, I wanted to avoid adding yet another block (it's already quite complicated as it is) so I don't know which solution would be best 😕

GP-Dan-Tovbein: From my point of view, as the best solution I think that it would be better if you wrap into a `.container`, for some reasons:
Pros:
- Easy to understand. It will be more cleaner and with less `.container` classes.
- Easy to maintain. In case if someone request to add a new block inside and you'll forget to set the `.container` to the new element.
- Better performance. Because this pattern will save to check a couple of `@media` less. 

Cons:
- Adding a new `core/group`.

But if you think that it's not necessary then go ahead with the current implementation 👍

GP-Dan-Tovbein (approved):
Code looks good!

However, I've left [this suggestion](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/828/files#r878073102) to check but it's not blocking the approval, just an improvement.

PR #831, opened 2022-04-18
PLANET-6520: Implement Quick Links block pattern

GP-Dan-Tovbein:
### Description

See [PLANET-6520](https://jira.greenpeace.org/browse/PLANET-6520)
I've yet to figure out how to implement [this transition](https://www-dev.greenpeace.org/gutenberg/deep-dive-topics-links/) requested by the design team, maybe it will be for a second iteration?

### Testing

On local, you'll need to set master-theme to the `new-pattern-placeholders` branch to see the correct image placeholders both in the editor and in the frontend ([corresponding PR](https://github.com/greenpeace/planet4-master-theme/pull/1717)).
Otherwise you can test the new pattern on [this page](https://www-dev.greenpeace.org/test-neptune/quick-links-patterns-block/) for example 🙂

### DIFF
--- a/assets/src/styles/blocks/core-overrides/_columns.scss
+++ b/assets/src/styles/blocks/core-overrides/_columns.scss
@@ -1,5 +1,5 @@
 .wp-block-columns.is-not-stacked-on-mobile.is-style-mobile-carousel {
-  @include mobile-only {
+  @include small-and-less {
     flex-wrap: nowrap;
     overflow-x: auto;
     scrollbar-width: none;
--- a/assets/src/styles/blocks/core-overrides/_columns.scss
+++ b/assets/src/styles/blocks/core-overrides/_columns.scss
@@ -1,4 +1,4 @@
-.is-style-mobile-carousel {
+.wp-block-columns.is-not-stacked-on-mobile.is-style-mobile-carousel {
   @include mobile-only {
     flex-wrap: nowrap;
     overflow-x: auto;
@@ -9,11 +9,8 @@
     }
 
     .wp-block-column {
-      min-width: 30vw;
-    }
-
-    &:not(.is-not-stacked-on-mobile) .wp-block-column:not(:last-child) {
-      margin-inline-end: $sp-2;
+      flex-basis: auto;
+      flex-shrink: 0;
     }
   }
 }
--- a/classes/patterns/class-block-pattern.php
+++ b/classes/patterns/class-block-pattern.php
@@ -38,6 +38,7 @@ abstract class Block_Pattern {
 			RealityCheck::class,
 			Issues::class,
 			DeepDive::class,
+			QuickLinks::class,
 		];
 
 		/**
--- a/classes/patterns/class-quicklinks.php
+++ b/classes/patterns/class-quicklinks.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Quick Links pattern class.
+ *
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Patterns;
+
+/**
+ * Class Quick Links.
+ *
+ * @package P4GBKS\Patterns
+ */
+class QuickLinks extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/quick-links';
+	}
+
+	/**
+	 * Returns the template for one column.
+	 */
+	public static function get_column_template(): string {
+		return '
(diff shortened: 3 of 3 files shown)


planet-4:

:watch: 2022.04.27 11:16:18

GP-Dan-Tovbein:
@mleray feel free to use this PR if you want!

GP-Dan-Tovbein:
Hey Maud! The code looks good but before approving it I need to validate with you some stuffs.

- Since we're currently using WP6, why if we change this template to `core/row` instead of `core/column`? (This is not blocking the merge).
- I don't see it well on `mobile-only` & `small-and-up` screens. Am I correct?

Screenshot 2022-06-10 at 09 57 32

Screenshot 2022-06-10 at 09 55 24

mleray:
> Hey Maud! The code looks good but before approving it I need to validate with you some stuffs.
> 
> * Since we're currently using WP6, why if we change this template to `core/row` instead of `core/column`? (This is not blocking the merge).
> * I don't see it well on `mobile-only` & `small-and-up` screens. Am I correct?
> 
> Screenshot 2022-06-10 at 09 57 32
> 
> Screenshot 2022-06-10 at 09 55 24

Thanks for the review @GP-Dan-Tovbein 🙂
- Why do you think it would be better to use `row` instead of `column`? Also currently the mobile carousel style is implemented for the Columns block only, although we could easily add it to other core blocks if needed
- For `mobile-only`, the columns' width is [set to `30vw`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/core-overrides/_columns.scss#L12) which looks good on really small screens but not that good on larger mobiles, maybe we could lower it to `25vw`? 🤔 
- Good catch for `small-and-up`! I'm not sure how we could best solve it, or even if we should, since it's an issue with the core block and not with our changes 😕

GP-Dan-Tovbein:
> > Hey Maud! The code looks good but before approving it I need to validate with you some stuffs.
> > 
> > * Since we're currently using WP6, why if we change this template to `core/row` instead of `core/column`? (This is not blocking the merge).
> > * I don't see it well on `mobile-only` & `small-and-up` screens. Am I correct?
> > 
> > Screenshot 2022-06-10 at 09 57 32
> > Screenshot 2022-06-10 at 09 55 24
> 
> Thanks for the review @GP-Dan-Tovbein 🙂
> 
>     * Why do you think it would be better to use `row` instead of `column`? Also currently the mobile carousel style is implemented for the Columns block only, although we could easily add it to other core blocks if needed
> 
>     * For `mobile-only`, the columns' width is [set to `30vw`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/core-overrides/_columns.scss#L12) which looks good on really small screens but not that good on larger mobiles, maybe we could lower it to `25vw`? 🤔
> 
>     * Good catch for `small-and-up`! I'm not sure how we could best solve it, or even if we should, since it's an issue with the core block and not with our changes 😕

I've prepared this [demo page](https://www-dev.greenpeace.org/test-neptune/quick-links-pattern-row/) using `core/row` instead. 

**PROS**
- The template that is smaller and more readable.
Screenshot 2022-06-10 at 11 52 13
Screenshot 2022-06-10 at 11 52 26

- Don't get the stretching issue because you don't need to define columns and solve this:
> For mobile-only, the columns' width is [set to 30vw](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/core-overrides/_columns.scss#L12) which looks good on really small screens but not that good on larger mobiles, maybe we could lower it to 25vw? 🤔
- We're using a new feature.
- Setting `display: grid` to each item you'll be able to use `gap` and then avoid adding a spacer.

**CONS**
- We need to adapt it with the styles you've applied to the `core/column`

assets/src/styles/blocks/core-overrides/_columns.scss
```diff
@@ -1,5 +1,5 @@
-.is-style-mobile-carousel {
-  @include mobile-only {
+.wp-block-columns.is-not-stacked-on-mobile.is-style-mobile-carousel {
+  @media (max-width: 767px) {
```
lithrel, line 2: I think we have a mixin `@mixin small-and-less` that fits this ? not sure if we should use it/multiply mixins but it's there

mleray: Good point, I didn't know it existed! Thanks, updated 👍

PR #833, opened 2022-04-20
Add group variation for full block link

Inwerpsel:
### Description

This should work well on any block that can use `position: relative`, and as long as the link is not within a block with `position: relative` inside the container.

~~A drawback about this method is a block can only have 1 block style at a time. I have another PR that uses [block styles to add padding on the same blocks](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/832). The only way to use styles for more things is to add 1 style for every possible combination. For more than 2 things that would probably get out of hand.~~

A possible alternative is to create a [variation of the core group block](https://github.com/WordPress/gutenberg/blob/29c0c4239c5e16fd3a32bc4ea82aa0135450b344/packages/block-library/src/group/variations.js#L7-L43), similar to the new "row" block, which has this CSS. That would make it much more obvious what's happening than a style. That also frees up the block styles for actual styles.

I ended up using a variation because styles are already used for padding and equally spacing inside groups. A nice plus of the variation is that it shows up with the variation name, also in the block list.

Screenshot from 2022-05-03 11-08-24

### Testing

Example: https://www-dev.greenpeace.org/test-phoebe/expanded-link-block/

### DIFF
--- a/assets/src/editorIndex.js
+++ b/assets/src/editorIndex.js
@@ -47,3 +47,18 @@ replaceTaxonomyTermSelectors();
 setupCustomSidebar();
 setUpCssVariables();
 blockEditorValidation();
+
+const { registerBlockVariation } = wp.blocks;
+const { __ } = wp.i18n;
+
+registerBlockVariation('core/group', {
+  name: 'group-stretched-link',
+  title: __('Stretched Link', 'planet4-blocks-backend'),
+  description: __('Make the entire block contents clickable, using the first link inside.', 'planet4-blocks-backend'),
+  attributes: { className: 'group-stretched-link' },
+  scope: ['inserter', 'transform'],
+  isActive: (blockAttributes) => {
+    return blockAttributes.className === 'group-stretched-link';
+  },
+  icon: 'admin-links',
+});
--- a/assets/src/styles/blocks.scss
+++ b/assets/src/styles/blocks.scss
@@ -23,7 +23,6 @@
 
 // Other
 @import "blocks/WideBlocks";
-@import "blocks/Editor";
 
 .is-style-small-padding {
   padding: 8px;
@@ -58,3 +57,6 @@
 .is-style-roboto-font-family {
   font-family: $roboto;
 }
+
+// Variations
+@import "../variations/stretched-link";
--- a/assets/src/styles/blocks/Editor.scss
+++ b/assets/src/styles/blocks/Editor.scss
@@ -1,4 +1,0 @@
-.block-edit-mode-warning {
-  padding: $sp-2;
-  margin: 0;
-}
--- a/assets/src/styles/editorStyle.scss
+++ b/assets/src/styles/editorStyle.scss
@@ -15,6 +15,7 @@
 @import "blocks/TakeActionBoxout/edit";
 @import "blocks/CookiesEditor";
 @import "blocks/core-overrides/HeadingEditor";
+@import "../variations/stretched-link/edit";
 
 @import "components/LayoutSelector";
 @import "components/EmptyMessage";
@@ -27,3 +28,8 @@
 @import "components/FormField";
 @import "components/HTMLSidebarHelp";
 @import "editorFonts";
+
+.block-edit-mode-warning {
+  padding: $sp-2;
+  margin: 0;
+}
--- a/assets/src/variations/stretched-link/edit.scss
+++ b/assets/src/variations/stretched-link/edit.scss
@@ -1,0 +1,4 @@
+.group-stretched-link {
+  outline: 1px dashed grey;
+  outline-offset: 2px;
+}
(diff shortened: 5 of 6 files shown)


planet-4:

:watch: 2022.04.20 13:54:50

mleray (changes requested):
This doesn't work for external links, because they also use the `::after` pseudo-element 😕 Also it looks weird in this case:
Screenshot 2022-04-21 at 10 31 13

mleray:
Looks like [it will also be an issue for PDF links](https://github.com/greenpeace/planet4-master-theme/blob/a8a1a7cfce184bccd0eeb08ba237404e61f203e5/assets/src/scss/base/_icons.scss)

Inwerpsel:
> This doesn't work for external links, because they also use the `::after` pseudo-element confused Also it looks weird in this 

I made it use `:before` instead, which preserves both special link styles. However neither of them look particularly well inside of this block (though they might in others).

Could make sense to disable these styles on links that are expanded?

mleray:
> Could make sense to disable these styles on links that are expanded? I guess we can also follow up on this once somebody starts using a external/pdf link.

I agree that it makes sense to remove them in this case, since we already don't show them in headings for example. 🤔  Ideally we would check that before adding the classes in JS, but that can be done in a follow-up!

mleray:
I see that with this implementation since we still add the external link class, we also add a title, not sure if that's an issue 🤔 
Screenshot 2022-04-21 at 11 01 07

Inwerpsel:
> we also add a title, not sure if that's an issue

Well it's still accurate, but maybe the it doesn't look that great in some cases. Let's get back to it once we get more feedback?

Inwerpsel:
> Ideally we would check that before adding the classes in JS

In fact, ideally we wouldn't even be doing this in JS, but server side. Doing it in JS gives the browser more work and causes re-rendering during page load.

assets/src/variations/expanded-link/edit.scss
```diff
@@ -0,0 +1,10 @@
+.is-style-expand-link {
+  &.is-selected a {
+    outline: 1px dashed grey;
+    outline-offset: 2px;
+  }
+
+  a:not(:first-of-type) {
```
mleray, line 7: This is not working, since in the editor links are within another element (paragraph, heading, etc) so technically `a` tags are always `first-of-type`... I don't know how that can be solved 🤔

Inwerpsel: Yeah I should maybe have documented this better. It's only a "best attempt" to catch some cases, like a paragraph with multiple links inside it. Not sure if it's possible with CSS only to catch the other cases.

mleray: It would be nice to make a difference in the editor between this variation and a regular group block 🤔 Maybe the dotted outline should be around the whole block to show that it will be clickable?

lithrel:
Is there a clickable way to revert the expanding link ? I can't unselect it once selected on a grouo.
I'm a bit confused on how to use/test it, the [example](https://www-dev.greenpeace.org/test-phoebe/expanded-link-block/) has expanding link in columns, but the style is added to core/group so I can only use it on group/row

mleray:
> Is there a clickable way to revert the expanding link ? I can't unselect it once selected on a grouo. I'm a bit confused on how to use/test it, the [example](https://www-dev.greenpeace.org/test-phoebe/expanded-link-block/) has expanding link in columns, but the style is added to core/group so I can only use it on group/row

It is added as a variation of the Group block, so it's used as such. In the example I've added a Columns block and within each column I've added an `Expanded Link`, that's the way to do it! Not ideal, but I don't really see how else we can do 😕 
Good point about reverting it, I also can't figure it out 😅 Will look into it!

Also, Houssam spotted [this Bootstrap functionality](https://getbootstrap.com/docs/4.3/utilities/stretched-link/) which does exactly what we want, maybe we could use that instead of adding custom CSS, but I don't see how to do it without JS so maybe it's not better 🤔

mleray:
@lithrel the only way I found to "revert" the variation is by removing the corresponding CSS class from the `Additional CSS class(es)` in the sidebar, in the `Advanced` tab

lithrel (approved):
:+1:

assets/src/editorIndex.js
```diff
@@ -47,3 +48,18 @@ replaceTaxonomyTermSelectors();
...
+registerBlockVariation('core/group', {
+  name: 'group-stretched-link',
+  title: __('Stretched Link', 'planet4-blocks-backend'),
+  description: __('Make the entire block contents clickable, using the first link inside.', 'planet4-blocks-backend'),
+  attributes: { className: 'group-stretched-link' },
+  scope: ['inserter', 'transform'],
+  isActive: (blockAttributes) => {
+    return blockAttributes.className === 'group-stretched-link';
+  },
+  icon: <Fragment>🔗</Fragment>,
```
comzeradd, line 64: We could use the [admin-links](https://developer.wordpress.org/resource/dashicons/#admin-links) dash icon for visual consistency.

mleray: Updated! 👍

PR #835, opened 2022-04-22
Add Query and Query pagination to post report

Inwerpsel:
* I need to easily see which pages use these on GP Brasil site, so we
can check the scale of the image resolution issue.
* Also add Query Pagination, they're not allowed but we'd definitely
want to know if someone starts using, even on the dev site.


lithrel (approved):
:+1:

PR #837, opened 2022-04-25
PLANET-6742 Add right sizes attributes for some core blocks

Inwerpsel:
Ref: [PLANET-6742](https://jira.greenpeace.org/browse/PLANET-6742)

We found while checking the Brasil site that image handling in core is pretty terrible, not just on the Query Loop block. With some small fixes we can [reduce the amount of unnecessary data being sent](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/837#issuecomment-1108816046) by a lot.

These 2 blocks are relatively easy to provide the right `sizes` attribute, which helps browsers avoid downloading oversized images. Image handling is [still an open issue in core](https://github.com/WordPress/gutenberg/issues/6177) and doesn't seem to have much progress lately.

Investigating:
* Is the replacement safe to do? Should fail "gracefully" the fix would just not be applied, visual result is identical.
* It gets hard when multiple blocks that force a width are nested in each other. In that case it might be possible to access block context, at least for a few cases. For now this is not handled, only top level Media and Text / Query Loop grid will have proper `sizes`.

### How I tested these changes
https://github.com/ausi/respimagelint can scan a page's images and report on missing/incorrect data in srcset and sizes attributes.

On the master branch, running this bookmarklet clearly shows errors of images being multiple times smaller than their sizes attributes indicates.
Screenshot from 2022-04-25 14-15-41

On this branch, it should not give any more errors for images inside Query Loop (grid) and Media and Text. (page is identical)

Screenshot from 2022-04-25 14-17-01

### Media and Text
This one is easier and likely has more benefits. However if the block is inside a container that forces a width, it's still way off. At least page header images.

### Query Loop

* Makes the assumption that the only images inside the block are post featured images spanning the full width of the container.
* It gets hard when multiple blocks that force a width are nested in each other. In that case it might be possible to access block context, at least for a few cases.

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -454,7 +454,7 @@ add_filter(
 					$container    = $breakpoint['width'];
 					$column_count = isset( $breakpoint['collapse'] ) ? 1 : $column_count;
 
-					return "(min-width: ${screen}px) calc(($container / $column_count) - 1.25em * ($column_count - 1)";
+					return "(min-width: ${screen}px) calc(($container / $column_count) - 1.25em * ($column_count - 1))";
 				},
 				$breakpoints
 			);
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -432,11 +432,6 @@ $breakpoints = [
 		'screen' => 600,
 		'width'  => '540px',
 	],
-	[
-		'screen'   => 576,
-		'width'    => '540px',
-		'collapse' => true,
-	],
 ];
 
 add_filter(
@@ -450,16 +445,16 @@ add_filter(
 
 			$sizes = array_map(
 				function ( $breakpoint ) use ( $column_count ) {
-					$screen       = $breakpoint['screen'];
-					$container    = $breakpoint['width'];
-					$column_count = isset( $breakpoint['collapse'] ) ? 1 : $column_count;
+					$screen    = $breakpoint['screen'];
+					$container = $breakpoint['width'];
+					$cols_minus_one = $column_count - 1;
 
-					return "(min-width: ${screen}px) calc(($container / $column_count) - 1.25em * ($column_count - 1))";
+					return "(min-width: ${screen}px) calc($container / $column_count - 1.25em * $cols_minus_one)";
 				},
 				$breakpoints
 			);
 
-			$sizes_attr = 'sizes="' . implode( ', ', $sizes ) . ', calc(100vw - 24px)"';
+			$sizes_attr = 'sizes="' . implode( ', ', array_merge($sizes, ['calc(100vw - 24px)']) ) . '"';
 
 			// Assume all images are full width in a container.
 			$block_content = preg_replace( '/sizes=".*"/', $sizes_attr, $block_content );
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.04.25 12:13:26

Inwerpsel:
Interestingly the benefit is way bigger on desktop than it is on mobile. On mobile the images affected here are shown almost full screen width, so the `srcset` was already not that far off.

On desktop however....
before
Screenshot from 2022-04-25 18-54-22

after
Screenshot from 2022-04-25 18-54-35

planet4-gutenberg-blocks.php
```diff
@@ -410,3 +410,109 @@ function ( $std_class_object ) {
...
+		'screen' => 577,
+		'width'  => '540px',
+	],
+];
+
+add_filter(
+	'render_block',
+	function ( $block_content, $block, WP_Block $instance ) use ( $breakpoints ) {
+		// Query param toggle to test difference.
+		if ( isset( $_GET['nofix'] ) ) {
```
Inwerpsel, line 445: I'll remove this once review is passed. That's also the only failing test in PHP tests.

planet4-gutenberg-blocks.php
```diff
@@ -410,3 +410,109 @@ function ( $std_class_object ) {
...
+							return "(min-width: ${screen}px) calc(($container - 24px) / $fraction)";
+						},
+						$breakpoints
+					)
+				);
+
+				$sizes_attr = "sizes=\"{$sizes}, {$default}\"";
+			}
+
+			$image_class_start = "class=\"wp-image-$media_id ";
```
GP-Dan-Tovbein, line 505: Is it intentional the empty space at the end of the String `"wp-image-$media_id "`?

planet4-gutenberg-blocks.php
```diff
@@ -410,3 +410,109 @@ function ( $std_class_object ) {
...
+			$media_id    = $instance->attributes['mediaId'];
+			$media_width = $instance->attributes['mediaWidth'] ?? 50;
+
+			$srcset = wp_get_attachment_image_srcset( $media_id, 'full' );
+
+			if ( 'full' === $instance->attributes['align'] ) {
+				$sizes = ! $instance->attributes['isStackedOnMobile'] ? "${media_width}vw"
+					: "(min-width: 601px) {$media_width}vw, 100vw";
+
+				$sizes_attr = "sizes=\"{$sizes}\"";
```
GP-Dan-Tovbein, line 482: Couldn't you use a mix of `"` and `'` to this `string` `"sizes=\"{$sizes}\""`?

GP-Dan-Tovbein (approved):
Code looks good to me! 
I've left some comments that could be reviewed in case of them apply.

planet4-gutenberg-blocks.php
Inwerpsel: Yes, it's possible that it's running for image 123, but image 1234 also exists and it's inside of 1 of the inner blocks. The markup should (:crossed_fingers: ) always have another class there so it's safe to assume there's going to be a space.

planet4-gutenberg-blocks.php
Inwerpsel: Can you give an example?

GP-Dan-Tovbein: Maybe `$sizes_attr = "sizes='{$sizes}'";` 🤔 ?

Inwerpsel: Then the output HTML would start using single instead of double quotes. Would work but it's nicer to have consistent output.

PR #853, opened 2022-05-13
PLANET-6699: Remove purple visited state from Split Two Columns links

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6699

> Remove purple visited state from Split Two Columns links

Set visited colors as non-visited colors for issue link and tag link

## Test

Check a _Split two columns_ block on the frontend
- Toggle the `:visited` state for the blue issue link and yellow tag link
- Color should not change


mleray (approved):

PR #864, opened 2022-05-27
Hide image credits when caption is empty

mleray:
### Description

This will be useful in block patterns in particular, where we don't want to show caption nor credits, and currently credits cannot be manually removed (unlike the caption).
There are issues with having it automatically implemented like it is in this PR: editors cannot manually remove the `no-caption` class, it will be applied automatically, which might be confusing as to why there are credits in the library but they don't show. Also, I am worried that there might be use cases where editors want the credits to show even if there's no caption, maybe it could be a legal issue?

I thought of creating a block style for this, but we cannot combine styles and we already need one for the image to be rounded... Or maybe we could hide the caption as part of the rounded styles, but I don't think that makes a lot of sense 🤔 

Maybe the best option would be to just add the `no-caption` class to pattern Image blocks and hide the caption only then, and editors could manually add/remove it. It would be similar to our `force-no-lightbox` [class](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/core-overrides/Image.scss#L20), that editors can use if they want to disable the lightbox.

### Testing

You can test this behaviour on [this page](https://www-dev.greenpeace.org/test-leda/images-without-credits/) for example. Also on local, you just need to have an image that has credits but no caption, and in that case the credits should not show in the editor or in the frontend.

### DIFF
--- a/assets/src/components/Image/ImageBlockEdit.js
+++ b/assets/src/components/Image/ImageBlockEdit.js
@@ -12,16 +12,20 @@ export const ImageBlockEdit = (BlockEdit) => {
     }
 
     const { attributes, clientId } = props;
-    const { id, caption } = attributes;
+    const { id, caption, className } = attributes;
 
     // Get image data
     const image = useSelect(select => id ? select('core').getMedia(id) : null);
     const credits = image?.meta?._credit_text;
     // Compile data for insertion
-    const image_credits = credits && credits.length > 0 && ! caption.includes(credits)
+    const image_credits = credits && credits.length > 0 && caption && ! caption.includes(credits)
       ? (credits.includes('©') ? credits : `© ${credits}`)
       : null;
     const block_id = clientId ? `block-${clientId}` : null;
+
+    if (!caption && credits && (!className || !className.includes('no-caption'))) {
+      attributes.className = className ? `${className} no-caption` : 'no-caption';
+    }
 
     return (
       <>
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2022.05.27 10:19:55

assets/src/components/Image/ImageBlockEdit.js
```diff
@@ -12,17 +12,21 @@ export const ImageBlockEdit = (BlockEdit) => {
...
     const { attributes, clientId } = props;
-    const { id, caption } = attributes;
+    const { id, caption, className } = attributes;
 
     // Get image data
     const image = useSelect(select => id ? select('core').getMedia(id) : null);
     const credits = image?.meta?._credit_text;
     // Compile data for insertion
-    const image_credits = credits && credits.length > 0 && ! caption.includes(credits)
+    const image_credits = credits && credits.length > 0 && caption && ! caption.includes(credits)
```
GP-Dan-Tovbein, line 21: Just thinking..

What do you think to change to this:
```
const credits = image?.meta?._credit_text ?? [];
```
Then
```
const image_credits = credits.length && ! caption.includes(credits);
```
And finally (from [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/864/files#diff-fa236ef56a43d882aec0fcdf53bd1cfcb08666dfc9f4f4bc5eb08649b7a04a63R26))
```
if (!caption && credits.length && (!className || !className.includes('no-caption'))) {
```
I think it will make the code more cleaner.

GP-Dan-Tovbein (approved):
Code looks good! However, I've left a suggestion about a code improvement.

mleray:
Closing in favor of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/866, after discussing with Mag

PR #863, opened 2022-05-27
Add Image block rounded styles with specific sizes

mleray:
### Description

These have specific width and height, to facilitate creating block patterns such as [Deep Dive](https://jira.greenpeace.org/browse/PLANET-6534) and [Quick Links](https://jira.greenpeace.org/browse/PLANET-6520). I can't think of another way to force images to have a specific size and aspect ratio.

### Testing

You can test the new styles on your local environment or on the oberon test instance ([example page](https://www-dev.greenpeace.org/test-oberon/rounded-images-with-custom-sizes/)). When selecting either the `Rounded 90x90` or `Rounded 180x180` style, you should see the width and height adapt accordingly in the sidebar. Editors can still change these values manually, which might result in some confusion, but as far as I know it's not possible to disable them 😕

PR #867, opened 2022-05-31
Enforce hiding Image block caption with CSS

mleray:
### Description

We need to use `!important`, otherwise in the editor the caption will sometimes still show (in cases where the image has an alignement class for example, because the CSS selectors then change).

### Testing

On local, add an Image block and select an image that has a caption and/or some credits assigned in the library. Then add alignment to the block (such as `Center`) and the `force-no-caption` class in the advanced settings. On this branch the caption should be hidden in both frontend and backend, whereas on master branch the caption will still show in the editor.

### DIFF
--- a/assets/src/styles/blocks/core-overrides/Image.scss
+++ b/assets/src/styles/blocks/core-overrides/Image.scss
@@ -31,7 +31,7 @@
   }
 
   &.force-no-caption figcaption {
-    display: none;
+    display: none !important;
   }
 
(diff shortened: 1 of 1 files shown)


lithrel (approved):
:+1:

mleray:
Closing since we're now using block styles instead (https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/863)

PR #871, opened 2022-06-16
PLANET-6782: Pattern usage report

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6782

> Since we keep adding block patterns, it would be a good data point for assessing them at a later point to know how often each pattern is being used.

## Pattern report

Screenshot from 2022-07-05 11-40-33
Pattern report accessible in _Blocks > Pattern report (Beta)_

- List of patterns, with post informations
- Filter by pattern
- Group by pattern, post title, post id

### Pattern detection

This version uses 2 types of detection:
- by class name
  Each pattern has a `is-pattern-{pattern name}` class assigned to it, so that it is easily found in the database.
  This allows to see usage _and_ find modifications applied to patterns.
- by structure
  We match the blocks and inner blocks of the pattern to a post content, to see if the pattern is in the post.
  We generate a "signature" from the structure of blocks (basically a json_encode of a tree of blocks), and try to find that signature in a post.
  It is a lot more CPU intensive, but seems to be the only way to detect patterns inserted before the use of a specific class names.

The short term goal is to append the proper class to the existing patterns in production, so we can drop the "structure matching" detection. This will remove the current double detection issue (a pattern will appear twice if it's detected by both methods).

## Test

Test [on Nix](https://www-dev.greenpeace.org/test-nix/wp-admin/admin.php?page=plugin_patterns_report)
Also test the block rest API, it should include a `block_patterns` section
```console
> curl -k https://www-dev.greenpeace.org/test-nix/wp-json/plugin_blocks/v3/plugin_blocks_report/ | jq
```

### DIFF
--- a/classes/patterns/class-block-pattern.php
+++ b/classes/patterns/class-block-pattern.php
@@ -27,6 +27,29 @@ abstract class Block_Pattern {
 	abstract public static function get_config( $params = [] ): array;
 
 	/**
+	 * Returns the pattern classname used for tracking patterns.
+	 */
+	public static function get_classname(): string {
+		return 'is_pattern-' . preg_replace( '#[^_a-zA-Z0-9-]#', '_', static::get_name() );
+	}
+
+	/**
+	 * Patterns list.
+	 */
+	public static function get_list(): array {
+		return [
+			BlankPage::class,
+			DeepDive::class,
+			GetInformed::class,
+			HighlightedCta::class,
+			Issues::class,
+			QuickLinks::class,
+			RealityCheck::class,
+			SideImageWithTextAndCta::class,
+		];
+	}
+
+	/**
 	 * Pattern constructor.
 	 */
 	public static function register_all() {
@@ -34,16 +57,7 @@ abstract class Block_Pattern {
 			return;
 		}
 
-		$patterns = [
-			BlankPage::class,
-			SideImageWithTextAndCta::class,
-			HighlightedCta::class,
-			RealityCheck::class,
-			Issues::class,
-			DeepDive::class,
-			QuickLinks::class,
-			GetInformed::class,
-		];
+		$patterns = self::get_list();
 
 		/**
 		 * @var $pattern self
--- a/classes/patterns/class-block-pattern.php
+++ b/classes/patterns/class-block-pattern.php
@@ -7,6 +7,8 @@
 
 namespace P4GBKS\Patterns;
 
+use P4GBKS\Search\Pattern\PatternData;
+
 /**
  * Class Base_Pattern
  *
@@ -30,7 +32,7 @@ abstract class Block_Pattern {
 	 * Returns the pattern classname used for tracking patterns.
 	 */
 	public static function get_classname(): string {
-		return 'is_pattern-' . preg_replace( '#[^_a-zA-Z0-9-]#', '_', static::get_name() );
+		return PatternData::make_classname( static::get_name() );
 	}
 
 	/**
--- a/classes/patterns/class-deepdive.php
+++ b/classes/patterns/class-deepdive.php
@@ -58,12 +58,13 @@ class DeepDive extends Block_Pattern {
 	 * @param array $params Optional array of parameters for the config.
 	 */
 	public static function get_config( $params = [] ): array {
+		$classname = self::get_classname();
 		return [
 			'title'      => __( 'Deep Dive', 'planet4-blocks-backend' ),
 			'categories' => [ 'planet4' ],
 			'content'    => '
-				<!-- wp:group {"className":"block","align":"full","backgroundColor":"grey-05"} -->
-					<div class="wp-block-group alignfull block has-grey-05-background-color has-background">
+				<!-- wp:group {"className":"block ' . $classname . '","align":"full","backgroundColor":"grey-05"} -->
+					<div class="wp-block-group alignfull block ' . $classname . ' has-grey-05-background-color has-background">
 						<!-- wp:group {"className":"container"} -->
 							<div class="wp-block-group container">
 								<!-- wp:spacer {"height":"24px"} -->
--- a/classes/search/block/query/class-parameters.php
+++ b/classes/search/block/query/class-parameters.php
@@ -39,6 +39,11 @@ class Parameters {
 	/**
 	 * @var string[]
 	 */
+	private $post_type;
+
+	/**
+	 * @var string[]
+	 */
 	private $order;
 
 	/**
@@ -88,6 +93,7 @@ class Parameters {
 				'content'     => $text_search,
 				'attributes'  => $request['attributes'] ?? [ $text_search ],
 				'post_status' => $request['post_status'] ?? self::DEFAULT_POST_STATUS,
+				'post_type'   => $request['post_type'] ?? null,
 				'order'       => $request['order'] ?? null,
 			]
 		);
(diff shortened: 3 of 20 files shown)


classes/patterns/class-block-pattern.php
```diff
@@ -27,23 +27,37 @@ abstract public static function get_name(): string;
...
 	/**
-	 * Pattern constructor.
+	 * Returns the pattern classname used for tracking patterns.
 	 */
-	public static function register_all() {
-		if ( ! function_exists( 'register_block_pattern' ) ) {
-			return;
-		}
+	public static function get_classname(): string {
+		return 'is_pattern-' . preg_replace( '#[^_a-zA-Z0-9-]#', '_', static::get_name() );
```
GP-Dan-Tovbein, line 33: Thinking of CSS class names WP conventions, what if we use `-` instead of `_`? For instance the `is-style-*` WP's rule.

classes/patterns/class-deepdive.php
```diff
@@ -58,12 +58,13 @@ public static function get_column_template(): string {
...
 	 */
 	public static function get_config( $params = [] ): array {
+		$classname = self::get_classname();
 		return [
 			'title'      => __( 'Deep Dive', 'planet4-blocks-backend' ),
 			'categories' => [ 'planet4' ],
 			'content'    => '
-				<!-- wp:group {"className":"block","align":"full","backgroundColor":"grey-05"} -->
-					<div class="wp-block-group alignfull block has-grey-05-background-color has-background">
+				<!-- wp:group {"className":"block ' . $classname . '","align":"full","backgroundColor":"grey-05"} -->
```
GP-Dan-Tovbein, line 66: What do you think if we use directly `self::get_classname()` instead of defining a new variable immutable class name value?

classes/search/block/query/class-parameters.php
```diff
@@ -167,6 +173,13 @@ public function post_status(): ?array {
 		return $this->post_status ? $this->post_status : self::DEFAULT_POST_STATUS;
 	}
 
+	/**
+	 * List of required post types.
+	 */
+	public function post_type(): ?array {
+		return $this->post_type ? $this->post_type : null;
```
GP-Dan-Tovbein, line 180: What do you think if we return `$this->post_type ?? null;`?

GP-Dan-Tovbein: Also `post_type` is defined [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/871/files#diff-fea718b863cbbd69a7354610f017c389dae8078205f63da9d69ca366792b5b3bR42) as a `string` but this function returns an `array` or a nully value. Am I right?

lithrel: haha yes sorry, I think the PHPCS rules refuse `??` at the moment so I end up with stuff like that :facepalm: I'll change the rule instead of complying blindly :pray:

classes/search/pattern/class-patternusageapi.php
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * Get all patterns names
+	 */
+	private function pattern_names(): array {
+		return array_filter(
+			array_column(
+				( WP_Block_Patterns_Registry::get_instance() )->get_all_registered(),
+				'name'
+			),
+			fn ( $name ) => 'p4/blank-page' !== $name
```
GP-Dan-Tovbein, line 81: Is this check should be applied to the `p4/blank-page` pattern or to all the pattern pages?

classes/search/block/query/class-parameters.php
lithrel: Ha, so in fact post_type is an array of strings to holds multiple types, like post_status holds multiple statuses; the declaration `string[]` means an array `[]` of `string`, it's a way to [specify the content of the array](https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html#arrays) 
There are [other type declarations](https://phpstan.org/writing-php-code/phpdoc-types#general-arrays) that we can use once we plug something like phpstan on the code.

classes/patterns/class-deepdive.php
lithrel: I'm using a variable to store result so that I don't call the same function multiple times. But maybe it would be better to manually declare that classname as a static property or constant for each pattern class ?

classes/search/pattern/class-patternusageapi.php
GP-Dan-Tovbein: @lithrel we've just renamed to [`p4/blank-page-pattern-layout`](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/891/files#diff-b38cca8eccfb381ce2f1d2dd175ad88062072c86eff8469108296637ba28229fR22).

lithrel: Updated using BlankPage::get_name() function instead of a string, so it stays up to date :grin: 
Also added a class to GetInformed so it's tracked

classes/patterns/class-block-pattern.php
```diff
@@ -27,23 +27,37 @@ abstract public static function get_name(): string;
...
 	/**
-	 * Pattern constructor.
+	 * Returns the pattern classname used for tracking patterns.
 	 */
-	public static function register_all() {
-		if ( ! function_exists( 'register_block_pattern' ) ) {
-			return;
-		}
+	public static function get_classname(): string {
+		return 'is_pattern-' . preg_replace( '#[^_a-zA-Z0-9-]#', '-', static::get_name() );
```
GP-Dan-Tovbein, line 33: Linked to this [comment](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/871#discussion_r917881514)

classes/patterns/class-block-pattern.php
lithrel: Oh sorry I misunserstood and misread, you were talking about is_pattern -> is-pattern or about the char replacement '_' ?

classes/search/pattern/class-patternusagetable.php
```diff
@@ -0,0 +1,403 @@
...
+
+	/**
+	 * Pattern select
+	 */
+	private function patternname_dropdown() {
+		sort( $this->pattern_names );
+		$filter = $this->search_params->name() ?? [];
+
+		echo '<select name="name" id="filter-by-name">';
+		echo '<option value="">- All patterns -</option>';
```
GP-Dan-Tovbein, line 174: It would be great if we translate this text !

classes/search/pattern/class-patternusagetable.php
```diff
@@ -0,0 +1,403 @@
...
+	 * Available grouping as views.
+	 */
+	protected function get_views() {
+		$link_tpl        = '<a href="%s">%s</a>';
+		$active_link_tpl = '<a class="current" href="%s">%s</a>';
+		return [
+			'pattern_name' => sprintf(
+				'pattern_name' === $this->group_by ? $active_link_tpl : $link_tpl,
+				add_query_arg( 'group', 'pattern_name' ),
+				'Group by pattern name'
```
GP-Dan-Tovbein, line 280: Same here (i18n) !

classes/patterns/class-block-pattern.php
GP-Dan-Tovbein: I meant to the entire `string` because of a CSS `classnames`'s conventions and using `_` it's more related to BEM. What do you think?

lithrel: Yep good point, works for me :+1:

GP-Dan-Tovbein (approved):
Code looks good to me ! Excellent job 🥇

PR #778, opened 2022-02-01
PLANET-6617: Implement resize handler to Cover carousel layout

GP-Dan-Tovbein:
Ticket: https://jira.greenpeace.org/browse/PLANET-6617

This behavior is not implemented on production. 
We should add this for a better navigation.

[Demo page](https://www-dev.greenpeace.org/test-pandora/covers-carousel-layout-resize-handler/)

PR #872, opened 2022-06-16
PLANET-6617 Implement resize handler for Covers block carousel

mleray:
### Description

See [PLANET-6617](https://jira.greenpeace.org/browse/PLANET-6617)
Without this handler, resizing the window sometimes breaks the design because the covers' width changes based on screen width.

### Testing

You can test the functionality on [this page](https://www-dev.greenpeace.org/test-pandora/covers-with-resize-handler/) for example, the covers should slide back to the start when resizing the window!

### DIFF
--- a/assets/src/blocks/Covers/CoversFrontend.js
+++ b/assets/src/blocks/Covers/CoversFrontend.js
@@ -1,5 +1,5 @@
 import { Covers } from './Covers';
-import { useRef } from '@wordpress/element';
+import { useRef, useEffect } from '@wordpress/element';
 import { COVERS_LAYOUTS, COVERS_TYPES } from './CoversConstants';
 import { useCovers } from './useCovers';
 import { CoversCarouselControls } from './CoversCarouselControls';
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.06.16 12:29:33

assets/src/blocks/Covers/CoversFrontend.js
```diff
@@ -58,14 +58,27 @@ export const CoversFrontend = attributes => {
...
     }
   };
 
+  const onResizeHandler = () => {
+    slideCovers('', 1);
+  }
+
   const showLoadMoreButton = !isCarouselLayout && !!initialRowsLimit && covers.length > (amountOfCoversPerRow * row);
 
+  useEffect(() => {
```
GP-Dan-Tovbein, line 73: What do you think to implement [this other approach](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/778/files#diff-93c303aba0166eba7362c8c44643bb67e809e03106503a13337b8e00ce45a7a5R70-R82) added to the `useEffect`? It would be probably better since you're running all the resizing logic only if it's a carousel layout and the handler method is being created once. 

If you compare with this new implementation, on one hand you're checking if it's a carousel layout for adding the `resize` event handler and on the other hand you're creating the resize the `onResizeHandler` no matter if it's listening added the `resize` handler.

Does it make sense?

mleray: Yes, it makes sense! Thanks, I've updated the PR 🙂

GP-Dan-Tovbein (approved):
Code looks good to me!

PR #876, opened 2022-06-23
Add Pages block pattern category

mleray:
### Description

This is needed for the new pages templates, such as [PLANET-6529](https://jira.greenpeace.org/browse/PLANET-6529) and [PLANET-6523](https://jira.greenpeace.org/browse/PLANET-6523)

**Note:** it won't appear in the editor yet because it's empty for now, I'm adding it in a separate PR because we might need to make similar changes for the Page Header patterns.

### DIFF
--- a/classes/class-loader.php
+++ b/classes/class-loader.php
@@ -87,7 +87,7 @@ final class Loader {
 		new Blocks\HubspotForm();
 
 		/**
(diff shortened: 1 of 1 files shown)


lithrel (approved):
:+1:

PR #879, opened 2022-06-24
PLANET-6529 Create Action pattern layout

mleray:
### Description

See [PLANET-6529](https://jira.greenpeace.org/browse/PLANET-6529)

~There's a [related PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/890) to fix the block validation errors that occur with the Covers block, it's in a separate PR since it's also needed for other templates such as the Homepage one.~ Merged!

In order to hide this pattern in post types other than Action pages, we need to add a dummy block (literally called `ActionDummyBlock`, but that's open to discussion 😁) that will be allowed only in this post type. It's hidden from the block list/inserter too, so the only way to add it is via code like we do here in the pattern.

### Testing

On local or on the umbriel test instance, you can create a new **Action** page and add the `Action` pattern to it, to see the initial state of the template and then see how you can edit the various blocks within it!
The template should not appear in other post types (Page, Post, etc) and also not in the page creation modal.
The new `Action Dummy Block` used to hide the pattern in other post types should not be available in the block list/inserter.

### DIFF
--- a/assets/src/blocks/ActionPageDummy/ActionPageDummyBlock.js
+++ b/assets/src/blocks/ActionPageDummy/ActionPageDummyBlock.js
@@ -1,0 +1,13 @@
+const { registerBlockType } = wp.blocks;
+
+export const registerActionPageDummyBlock = () => {
+  registerBlockType('planet4-blocks/action-page-dummy', {
+    title: 'Action Page Dummy',
+    category: 'planet4-blocks',
+    supports: {
+      inserter: false,
+    },
+    edit: () => null,
+    save: () => null,
+  });
+};
--- a/classes/blocks/class-actionpagedummy.php
+++ b/classes/blocks/class-actionpagedummy.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * ActionPageDummy block class.
+ *
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Blocks;
+
+/**
+ * Class ActionPageDummy
+ *
+ * @package P4GBKS\Blocks
+ */
+class ActionPageDummy extends Base_Block {
+
+	/**
+	 * Block name.
+	 *
+	 * @const string BLOCK_NAME.
+	 */
+	const BLOCK_NAME = 'action-page-dummy';
+
+	/**
+	 * ActionPageDummy constructor.
+	 */
+	public function __construct() {
+		add_action( 'init', [ $this, 'register_actionpagedummy_block' ] );
+	}
+
+	/**
+	 * Register ActionPageDummy block.
+	 */
+	public function register_actionpagedummy_block() {
+		register_block_type(
+			self::get_full_block_name(),
+		);
+	}
+
+	/**
+	 * Required by the `Base_Block` class.
+	 *
+	 * @param array $fields Unused, required by the abstract function.
+	 */
+	public function prepare_data( $fields ): array {
+		return [];
+	}
+}
--- a/classes/patterns/class-action.php
+++ b/classes/patterns/class-action.php
@@ -1,0 +1,81 @@
+<?php
+/**
+ * Action pattern class.
+ *
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Patterns;
+
+use P4GBKS\Patterns\Templates\Covers;
+
+/**
+ * Class Action.
+ *
+ * @package P4GBKS\Patterns
+ */
+class Action extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/action-pattern-layout';
+	}
+
+	/**
+	 * Returns the pattern config.
+	 *
+	 * @param array $params Optional array of parameters for the config.
+	 */
+	public static function get_config( $params = [] ): array {
+		$media_link = esc_url( get_template_directory_uri() ) . '/images/placeholders/placeholder-546x415.jpg';
+
+		return [
+			'title'      => __( 'Action', 'planet4-blocks-backend' ),
+			'categories' => [ 'layouts' ],
+			'content'    => '
+				<!-- wp:group {"className":"block"} -->
+					<div class="wp-block-group">
+						<!-- wp:media-text {"mediaLink":"' . $media_link . '","mediaType":"image"} -->
+							<div class="wp-block-media-text is-stacked-on-mobile">
+								<figure class="wp-block-media-text__media">
+									<img src="' . $media_link . '" alt="' . __( 'Default image', 'planet4-blocks-backend' ) . '"/>
+								</figure>
+								<div class="wp-block-media-text__content">
+									<!-- wp:gravityforms/form /-->
+								</div>
+							</div>
+						<!-- /wp:media-text -->
+						<!-- wp:group {"backgroundColor":"grey-05","align":"full","className":"is-style-large-padding"} -->
+							<div class="wp-block-group alignfull has-grey-05-background-color has-background is-style-large-padding">
+								<!-- wp:group {"className":"container"} -->
+									<div class="wp-block-group container">
+										' . SideImageWithTextAndCta::get_config(
+										[
+											'media_position'    => 'right',
+											'title_placeholder' => __( 'The problem', 'planet4-blocks' ),
+										]
+									)['content'] . '
+									</div>
+								<!-- /wp:group -->
+							</div>
+						<!-- /wp:group -->
+						' . Covers::get_content( [ 'cover_type' => 'take-action' ] ) . '
+						<!-- wp:separator {"backgroundColor":"grey-20","className":"has-text-color has-grey-20-color has-grey-20-background-color has-background is-style-wide"} -->
+						<hr class="wp-block-separator has-text-color has-grey-20-color has-alpha-channel-opacity has-grey-20-background-color has-background is-style-wide"/>
+						<!-- /wp:separator -->
+						' . QuickLinks::get_config(
+							[
+								'background_color'  => 'white',
+								'title_placeholder' => __( 'Explore by topics', 'planet4-blocks' ),
+							]
+						)['content'] . '
+					</div>
+					<!-- wp:planet4-blocks/action-page-dummy /-->
+				<!-- /wp:group -->
+			',
+		];
+	}
+}
--- a/assets/src/editorIndex.js
+++ b/assets/src/editorIndex.js
@@ -22,6 +22,7 @@ import { blockEditorValidation } from './BlockEditorValidation';
 import { registerGuestBookBlock } from './blocks/GuestBook/GuestBookBlock';
 import { registerBlock as registerShareButtonsBlock } from './blocks/ShareButtons/ShareButtonsBlock';
 import { registerPageHeaderBlock } from './blocks/PageHeader/PageHeaderBlock';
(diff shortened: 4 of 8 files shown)


classes/patterns/class-action.php
```diff
@@ -0,0 +1,82 @@
...
+	public static function get_name(): string {
+		return 'p4/action';
+	}
+
+	/**
+	 * Returns the pattern config.
+	 *
+	 * @param array $params Optional array of parameters for the config.
+	 */
+	public static function get_config( $params = [] ): array {
```
comzeradd, line 30: You can add a `blockTypes` here to add the pattern to Page Creation modal (like [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/patterns/class-blankpage.php#L33))

mleray: [Updated](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/commits/ae8a556d0c2d40e79f5369728ab59a14057b7694)!

classes/patterns/class-action.php
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @package P4GBKS\Patterns
+ */
+class Action extends Block_Pattern {
+
+	/**
+	 * Returns the pattern name.
+	 */
+	public static function get_name(): string {
+		return 'p4/action';
```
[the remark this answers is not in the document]
lithrel: `-template` makes sense but also [means](https://developer.wordpress.org/themes/basics/template-files/) [something](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-templates/) in wordpress :sob: I'll add my proposition to the confusion `-page-content`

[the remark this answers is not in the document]
mleray: `-page-pattern` sounds good to me! @lithrel what do you think?
Whichever we decide, we will also need to update the [Get Informed template](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/patterns/class-getinformed.php#L22) that has already been merged 😬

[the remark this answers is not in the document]
lithrel: Good for me :+1:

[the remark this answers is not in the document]
mleray: Thank you both for your input! [Updated](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/commits/01a63331e270a4ab26a1b2632fbce3bac96640f8) 😄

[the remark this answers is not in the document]
comzeradd: Hey! Coming late to the <s>party</s> discussion. There is actually a proposal by suzi to name these as "Layout Block Patterns" (or "Layouts" for short), because "Templates" has indeed a different meaning in WP and "Pages" is just one post type.

So maybe we could use `-layout-pattern` instead :grin:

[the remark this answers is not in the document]
mleray: I'm fine with either of those 🤷‍♀️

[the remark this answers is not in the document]
mleray: [Updated](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/commits/7040332e5b4649d2ead50cc1e66d11a1f5439598) after confirming with @GP-Dan-Tovbein

assets/src/blocks/ActionPageDummy/ActionPageDummyBlock.js
```diff
@@ -0,0 +1,13 @@
+const { registerBlockType } = wp.blocks;
+
+export const registerActionPageDummyBlock = () => {
+  registerBlockType('planet4-blocks/action-page-dummy', {
+    title: 'Action Page Dummy',
```
GP-Dan-Tovbein, line 5: Wouldn't we need to translate this text?

classes/patterns/class-action.php
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+namespace P4GBKS\Patterns;
+
+/**
+ * Class Action.
+ *
+ * @package P4GBKS\Patterns
+ */
+class Action extends Block_Pattern {
```
GP-Dan-Tovbein, line 16: Just thinking.. shouldn't we rename the `class` for a another more specific name? SInce `Action` sounds a general stuff. Maybe `ActionLayout`?

classes/blocks/class-actionpagedummy.php
```diff
@@ -0,0 +1,49 @@
...
+
+	/**
+	 * ActionPageDummy constructor.
+	 */
+	public function __construct() {
+		add_action( 'init', [ $this, 'register_actionpagedummy_block' ] );
+	}
+
+	/**
+	 * Register HubspotForm block.
```
GP-Dan-Tovbein, line 33: We will need to edit this comment !

classes/patterns/class-action.php
```diff
@@ -0,0 +1,83 @@
...
+										[
+											'media_position'    => 'right',
+											'title_placeholder' => __( 'The problem', 'planet4-blocks' ),
+										]
+									)['content'] . '
+									</div>
+								<!-- /wp:group -->
+							</div>
+						<!-- /wp:group -->
+						<!-- wp:planet4-blocks/covers {
```
GP-Dan-Tovbein, line 63: If you change to something like [this](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/884/files#diff-25f0e4e3db0b77616a6f7b7662e425dc8343525cac2a6b8a478461a2ef08b376R86-R88) it would fix the issue we have with this block. 
We should definitely follow up this issue and find the best solution because it happens not only with the Cover but also with other `p4/blocks`, i.e. the Carousel Header.

assets/src/blocks/ActionPageDummy/ActionPageDummyBlock.js
mleray: I don't think it's necessary in this case, since the block isn't meant to be visible in the editor 🤔 It can only be added via code

mleray: It would probably be confusing for editors to have this string to translate even though they cannot see it anywhere!

classes/blocks/class-actionpagedummy.php
mleray: [Done](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/commits/25a895d910a365e01626db4e6ea715f10358a739)!

classes/patterns/class-action.php
mleray: [Done](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/commits/f793f2921f8b39a0401651a29bcdf56ae6d1bd95)! Indeed it's not an ideal solution, but it seems to work

classes/patterns/class-action.php
mleray: Maybe all of these pattern layouts' classes should indicate it in their name, so here for example `ActionPatternLayout`? 🤔

GP-Dan-Tovbein: I agree to change to `ActionPatternLayout`. Not now, but we should rename the other layout patterns too.

classes/patterns/class-action.php
```diff
@@ -0,0 +1,85 @@
...
+							</div>
+						<!-- /wp:media-text -->
+						<!-- wp:group {"backgroundColor":"grey-05","align":"full","className":"is-style-large-padding"} -->
+							<div class="wp-block-group alignfull has-grey-05-background-color has-background is-style-large-padding">
+								<!-- wp:group {"className":"container"} -->
+									<div class="wp-block-group container">
+										' . SideImageWithTextAndCta::get_config(
+										[
+											'media_position'    => 'right',
+											'title_placeholder' => __( 'The problem', 'planet4-blocks' ),
```
GP-Dan-Tovbein, line 56: Why did you define the domain to `planet4-blocks` but from [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/files#diff-ff56a7b10430fcb2a283a3994a55fad63d259fb799ad2539f4da4f42798a16e6R42) you defined to `planet4-blocks-backend`?

mleray: I've left `planet4-blocks` here because it could still appear in the frontend in case editors don't change it

classes/patterns/class-action.php
```diff
@@ -0,0 +1,81 @@
...
+										]
+									)['content'] . '
+									</div>
+								<!-- /wp:group -->
+							</div>
+						<!-- /wp:group -->
+						' . Covers::get_content( [ 'cover_type' => 'take-action' ] ) . '
+						<!-- wp:separator {"backgroundColor":"grey-20","className":"is-style-wide"} -->
```
lithrel, line 66: The separator ends up different in the editor (it doesn't have the backgorund color UI selected and gets an opacity property)
```
<!-- wp:separator {"opacity":"css","className":"has-text-color has-grey-20-color has-grey-20-background-color has-background is-style-wide"} -->
<hr class="wp-block-separator has-css-opacity has-text-color has-grey-20-color has-grey-20-background-color has-background is-style-wide"/>
<!-- /wp:separator -->
```
I tested repasting it after setting the bg color, maybe the alpha-channel class fixes the thing ? seems to work :shrug: 
```
<!-- wp:separator {"backgroundColor":"grey-20","className":"has-text-color has-grey-20-color has-grey-20-background-color has-background is-style-wide"} -->
<hr class="wp-block-separator has-text-color has-grey-20-color has-alpha-channel-opacity has-grey-20-background-color has-background is-style-wide"/>
<!-- /wp:separator -->
```

classes/patterns/class-action.php
```diff
@@ -0,0 +1,81 @@
...
+								</div>
+							</div>
+						<!-- /wp:media-text -->
+						<!-- wp:group {"backgroundColor":"grey-05","align":"full","className":"is-style-large-padding"} -->
+							<div class="wp-block-group alignfull has-grey-05-background-color has-background is-style-large-padding">
+								<!-- wp:group {"className":"container"} -->
+									<div class="wp-block-group container">
+										' . SideImageWithTextAndCta::get_config(
+										[
+											'media_position'    => 'right',
```
lithrel, line 57: I just realized there's a mistake in the `SideImageWithTextAndCta`, the media position class is used [instead of the mediaPosition attribute](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/patterns/class-sideimagewithtextandcta.php#L45), which confuses the media-text UI. You can add a tiny fix in this PR or we can open a different one

classes/patterns/class-action.php
mleray: Good catch! Your fix seems to work indeed, but since the text color is also defined I see this warning in the sidebar:
Screenshot 2022-08-04 at 10 55 16
But if I remove the text color from the attributes, then the background color gets unset too for some reason 🤷‍♀️ I guess it's not a big deal, maybe it's a core block issue 🤔

mleray: I still prefer your version since at least the background color is properly selected in the sidebar!

classes/patterns/class-action.php
mleray: Another good catch! [Pushed a fix](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/879/commits/b0688384075ca122da35a22b50ff8936e3952239)

lithrel (approved):
:+1:

PR #880, opened 2022-06-27
Fix deprecated wp.editor elements console warnings

mleray:
### Description

Some of these will be removed very soon so we need to stop using them:

```
wp.editor.RichText is deprecated since version 5.3 and will be removed in version 6.2. Please use wp.blockEditor.RichText instead.
```
```
wp.editor.InspectorControls is deprecated since version 5.3 and will be removed in version 6.2. Please use wp.blockEditor.InspectorControls instead.
```

### Testing

On local, you can add an Image block or a Counter block to any page and on `master` branch you'll see the console warnings, but not on this branch. Both blocks should still work as expected!


sagarsdeshmukh (approved):
👍

PR #877, opened 2022-06-23
PLANET-6523 Implement Get Informed page template

mleray:
### Description

See [PLANET-6523](https://jira.greenpeace.org/browse/PLANET-6523)

PR #883, opened 2022-07-01
Add title placeholder param to some patterns

mleray:
### Description

This is needed for some page templates, such as the [Action](https://jira.greenpeace.org/browse/PLANET-6529) one or the [Get Informed](https://jira.greenpeace.org/browse/PLANET-6523) one. We want to be able to add some placeholders when creating these patterns. You can see how it would be used for example [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/877/files#diff-59777031749d62dcf9c1abe1512e44340868b40c3004277c5357267891d3bfecR37).


GP-Dan-Tovbein (approved):
Code looks good to me!

PR #884, opened 2022-07-01
PLANET-6524: New homepage block pattern

GP-Dan-Tovbein:
Ref: [PLANET-6524](https://jira.greenpeace.org/browse/PLANET-6524)

Create a new homepage layout pattern.

[Demo page](https://www-dev.greenpeace.org/test-phobos/51534-2/) 

## Testing Instructions
1. Create a new page
2. Go to the Patterns > Pages and chose the Homepage.

### DIFF
--- a/classes/patterns/class-block-pattern.php
+++ b/classes/patterns/class-block-pattern.php
@@ -44,7 +44,6 @@ abstract class Block_Pattern {
 			DeepDive::class,
 			GetInformed::class,
 			HomePage::class,
-			SideImageWithTextAndCta::class,
 			HighlightedCta::class,
 			Issues::class,
 			QuickLinks::class,
--- a/classes/patterns/class-block-pattern.php
+++ b/classes/patterns/class-block-pattern.php
@@ -43,7 +43,7 @@ abstract class Block_Pattern {
 			BlankPage::class,
 			DeepDive::class,
 			GetInformed::class,
-			HomePage::class,
+			Homepage::class,
 			HighlightedCta::class,
 			Issues::class,
 			QuickLinks::class,
--- a/classes/patterns/class-block-pattern.php
+++ b/classes/patterns/class-block-pattern.php
@@ -43,6 +43,8 @@ abstract class Block_Pattern {
 			BlankPage::class,
 			DeepDive::class,
 			GetInformed::class,
+			HomePage::class,
+			SideImageWithTextAndCta::class,
 			HighlightedCta::class,
 			Issues::class,
 			QuickLinks::class,
--- a/classes/patterns/class-homepage.php
+++ b/classes/patterns/class-homepage.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * HomePage pattern class.
+ * Homepage pattern layout class.
  *
  * @package P4GBKS
  * @since 0.1
@@ -9,17 +9,17 @@
 namespace P4GBKS\Patterns;
 
 /**
- * This class is used for returning a home page template.
+ * This class is used for returning a homepage pattern layout template.
  *
  * @package P4GBKS\Patterns
  */
-class HomePage extends Block_Pattern {
+class Homepage extends Block_Pattern {
 
 	/**
 	 * @inheritDoc
 	 */
 	public static function get_name(): string {
-		return 'p4/home-page-pattern-layout';
+		return 'p4/homepage-pattern-layout';
 	}
 
 	/**
@@ -29,7 +29,7 @@ class HomePage extends Block_Pattern {
 	 */
 	public static function get_config( $params = [] ): array {
 		return [
-			'title'      => __( 'Homepage', 'planet4-blocks-backend' ),
+			'title'      => __( 'Homepage pattern layout', 'planet4-blocks-backend' ),
 			'blockTypes' => [ 'core/post-content' ],
 			'categories' => [ 'layouts' ],
 			'content'    => '
@@ -65,7 +65,7 @@ class HomePage extends Block_Pattern {
 				</div>
 				<!-- /wp:planet4-blocks/carousel-header -->
 
-				' . Issues::get_config()['content'] . '
+				' . Issues::get_config( [ 'title_placeholder' => __( 'The issues we work on', 'planet4-blocks-backend' ) ] )['content'] . '
 
 				<!-- wp:spacer {"height":"88px"} -->
 				<div style="height:88px" aria-hidden="true" class="wp-block-spacer"></div>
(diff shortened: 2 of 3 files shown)


planet-4:

:watch: 2022.07.01 18:30:26

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @package P4GBKS\Patterns
+ */
+class HomePage extends Block_Pattern {
+
+/**
+ * @inheritDoc
+ */
+public static function get_name(): string {
+	return 'p4/home-page';
```
comzeradd, line 22: let's align that as: `p4/home-page-pattern-layout`

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * @inheritDoc
+ *
+ * @param array $params Optional array of parameters for the config.
+ */
+public static function get_config( $params = [] ): array {
+	return [
+		'title'      => __( 'Homepage', 'planet4-blocks-backend' ),
+		'blockTypes' => [ 'core/post-content' ],
+		'categories' => [ 'pages' ],
```
comzeradd, line 34: `'categories' => [ 'layouts' ],`

classes/patterns/class-block-pattern.php
```diff
@@ -36,6 +36,7 @@ public static function register_all() {
 
 		$patterns = [
 			BlankPage::class,
+			HomePage::class,
```
Thendo-T, line 39: 👍🏾

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,117 @@
...
+	public static function get_config( $params = [] ): array {
+		return [
+			'title'      => __( 'Homepage', 'planet4-blocks-backend' ),
+			'blockTypes' => [ 'core/post-content' ],
+			'categories' => [ 'layouts' ],
+			'content'    => '
+			<!-- wp:group -->
+			<div class="wp-block-group">
+
+				<!-- wp:planet4-blocks/carousel-header -->
```
GP-Dan-Tovbein, line 39: TBH, I disagree to add the carousel as this way.
However, if we add it with somethig like  you'll get a block error whenever you are editing this pattern.

It does happen something similar with the Covers.

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,117 @@
...
+
+				' . SideImageWithTextAndCta::get_config( [] )['content'] . '
+
+				' . SideImageWithTextAndCta::get_config( [ 'media_position' => 'right' ] )['content'] . '
+
+				<!-- wp:spacer {"height":"104px"} -->
+				<div style="height:104px" aria-hidden="true" class="wp-block-spacer"></div>
+				<!-- /wp:spacer -->
+
+				<!-- wp:planet4-blocks/covers {"cover_type":"take-action","className":"is-style-take-action"} -->
```
GP-Dan-Tovbein, line 86: Similar to https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/884/files#r922201476

classes/patterns/class-homepage.php
comzeradd: Do you think is possible to at least import this markup from somewhere else in the codebase? Because we would need to use Carousel Header in other page layouts too.

GP-Dan-Tovbein: Yes, I thought the same and also we need to use the Covers too.

classes/patterns/class-block-pattern.php
```diff
@@ -43,6 +43,8 @@ public static function get_list(): array {
 			BlankPage::class,
 			DeepDive::class,
 			GetInformed::class,
+			HomePage::class,
+			SideImageWithTextAndCta::class,
```
mleray, line 47: Looks like you have this one declared twice 😬 Probably something went wrong while rebasing at some point!

GP-Dan-Tovbein: Good catch! I've just removed.

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,151 @@
...
+
+				<!-- wp:planet4-blocks/articles {
+					"article_heading":"' . __( 'Read our Stories', 'planet4-blocks' ) . '"
+				} /-->
+
+				<!-- wp:spacer {"height":"56px"} -->
+				<div style="height:56px" aria-hidden="true" class="wp-block-spacer"></div>
+				<!-- /wp:spacer -->
+
+				' . SideImageWithTextAndCta::get_config( [] )['content'] . '
```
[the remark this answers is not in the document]
mleray: So you can just write `get_config()`

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,151 @@
...
+
+				<!-- wp:spacer {"height":"30px"} -->
+				<div style="height:30px" aria-hidden="true" class="wp-block-spacer"></div>
+				<!-- /wp:spacer -->
+
+				' . SideImageWithTextAndCta::get_config( [ 'media_position' => 'right' ] )['content'] . '
+
+				<!-- wp:spacer {"height":"56px"} -->
```
mleray, line 90: Here you marked `56px`, but on the following line it's `54px`, so I get an error in the editor when adding the layout

GP-Dan-Tovbein: Good catch, thanks!
That happens because I've updated the code directly to the demo page instead of creating a new layout. And after the UAT is passed then I set all these values on my local file.

classes/patterns/class-homepage.php
```diff
@@ -0,0 +1,151 @@
...
+					<!-- wp:group {"className":"container"} -->
+					<div class="wp-block-group container">
+
+					<!-- wp:columns -->
+					<div class="wp-block-columns">
+
+						<!-- wp:column -->
+						<div class="wp-block-column">
+
+							<!-- wp:heading {"level":2,"style":{"typography":{"fontSize":"40px"}},"placeholder":"' . __( 'Enter title', 'planet4-blocks-backend' ) . '"} -->
```
mleray, line 117: I'm pretty sure that level 2 is the default one, so you can probably remove it from the attributes (not a big deal, but it simplifies the line a bit!)

classes/patterns/class-homepage.php
mleray: Maybe this CarouselHeader should be turned into a template too, similar to what we did with [Covers](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/classes/patterns/templates/class-covers.php)?

GP-Dan-Tovbein: Good point! I didn't add it since as far as I know this is the only pattern layout that uses this block. What do you think?

mleray: I think it would still make sense to do, because it would make the pattern more readable and we might have other pattern layouts in the future that use the CarouselHeader too... But I'll leave it up to you, it's not a blocker at all 🙂

GP-Dan-Tovbein: Let's move it then 💪

classes/patterns/class-block-pattern.php
```diff
@@ -43,6 +43,7 @@ public static function get_list(): array {
 			BlankPage::class,
 			DeepDive::class,
 			GetInformed::class,
+			HomePage::class,
```
mleray, line 46: This usage has a capital `P` but [the actual declaration doesn't](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/884/files#diff-25f0e4e3db0b77616a6f7b7662e425dc8343525cac2a6b8a478461a2ef08b376R20), even if it seems to work as expected we should probably be consistent in the naming?

classes/patterns/templates/class-carouselheader.php
```diff
@@ -0,0 +1,55 @@
...
+ * @package P4GBKS
+ * @since 0.1
+ */
+
+namespace P4GBKS\Patterns\Templates;
+
+use P4GBKS\Patterns\Templates\TemplatePattern;
+
+/**
+ * This class is used for returning a blank page with a default content.
```
mleray, line 14: This comment should be updated 😉

mleray (approved):
Looks good to me! 👏

PR #886, opened 2022-07-06
PLANET-6829 Fix block style label

sagarsdeshmukh:
The block style label no more accept tooltip component from wp6 upgrade, hence concat a label and help text

https://jira.greenpeace.org/browse/PLANET-6829

**Testing:**
- Test below blocks for block style on local or test instance.
   - Covers
   - Gallery
   - Submenu
   - Columns
   - Hubspot form

### DIFF
--- a/assets/src/blocks/Columns/ColumnsBlock.js
+++ b/assets/src/blocks/Columns/ColumnsBlock.js
@@ -1,21 +1,10 @@
 import { ColumnsEditor } from './ColumnsEditor.js';
-import { Tooltip } from '@wordpress/components';
 import { LAYOUT_NO_IMAGE, LAYOUT_IMAGES, LAYOUT_ICONS, LAYOUT_TASKS } from './ColumnConstants.js';
 import { example } from './example';
+import { getStyleLabel } from '../../functions/getStyleLabel';
 
 const { __ } = wp.i18n;
 const { registerBlockType } = wp.blocks;
-
-const getStyleLabel = (label, help) => {
-  if (help) {
-    return (
-      <Tooltip text={help}>
-        <span>{label}</span>
-      </Tooltip>
-    );
-  }
-  return label;
-};
 
 export const registerColumnsBlock = () =>
   registerBlockType('planet4-blocks/columns', {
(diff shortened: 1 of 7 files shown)


assets/src/blocks/Columns/ColumnsBlock.js
```diff
@@ -8,11 +7,7 @@ const { registerBlockType } = wp.blocks;
 
 const getStyleLabel = (label, help) => {
   if (help) {
-    return (
-      <Tooltip text={help}>
-        <span>{label}</span>
-      </Tooltip>
-    );
+    return label . ' - ' . help;
```
mleray, line 10: This notation doesn't work in JS, it crashes 🙃 I think you should write either 
```
label + ' - ' + help
```
or 
```
`${label} - ${help}`
```

sagarsdeshmukh: ops, my bad 😬 , I was just jump from PHP to JS very fast. Thanks 🙏

GP-Dan-Tovbein (commented):
Code looks to me!

However and thinking loud 🤔 what if we put the `getStyleLabel` function into the `functions` directory? Since we're currently duplicating this method in a bunch of modules .

sagarsdeshmukh:
> Code looks to me!
> 
> However and thinking loud 🤔 what if we put the `getStyleLabel` function into the `functions` directory? Since we're currently duplicating this method in a bunch of modules .

Definitely 💯 , Done ✅

mleray (approved):

Only thing I would suggest is maybe changing the [description of the HubspotFormBlock's style](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d5f8e818f6c9280e8f08de49eae2de20a20ede65/assets/src/blocks/HubspotForm/HubspotFormBlock.js#L73), right now it's a link but I'm not sure that makes sense to use since editors can't click on it 🤔 

Screenshot 2022-07-06 at 15 22 23

sagarsdeshmukh:
> Tested on local, looks good to me 👍
> 
> Only thing I would suggest is maybe changing the [description of the HubspotFormBlock's style](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/d5f8e818f6c9280e8f08de49eae2de20a20ede65/assets/src/blocks/HubspotForm/HubspotFormBlock.js#L73), right now it's a link but I'm not sure that makes sense to use since editors can't click on it 🤔
> 
> Screenshot 2022-07-06 at 15 22 23

Indeed the style description link is not very user-friendly, I added the same link as read more in sidebar at bottom.
image

lithrel (approved):
:+1:

PR #889, opened 2022-07-07
PLANET-6817 Add hover zoom effect to images in Stretched links

mleray:
### Description

See the [ticket](https://jira.greenpeace.org/browse/PLANET-6817) and the [example](https://www-dev.greenpeace.org/gutenberg/topics-link-animate-both-image-title-on-hover/) that Houssam implemented.

This is mostly for block patterns such as Deep Dive or Quick Links, but with this implementation it will be applied to all Stretched links that contain a rounded image.

### Testing

You can check out [this page](https://www-dev.greenpeace.org/test-oberon/patterns-with-image-zoom-on-hover/) that I prepared for UAT! Otherwise on local you can test it in any Quick Links or Deep Dive pattern.

### DIFF
--- a/assets/src/variations/stretched-link/index.scss
+++ b/assets/src/variations/stretched-link/index.scss
@@ -1,6 +1,21 @@
 .group-stretched-link {
   position: relative;
 
+  [class*="is-style-rounded-"].force-no-caption {
+    overflow: hidden;
+    border-radius: 50%;
+
+    img {
+      transition: all .2s ease-in-out;
+    }
+  }
+
+  &:hover {
+    [class*="is-style-rounded-"] img {
+      transform: scale(1.3);
+    }
+  }
+
   // Ensure it's only applying to the first link.
   a:first-of-type {
     &:before {


assets/src/styles/blocks/core-overrides/Image.scss
```diff
@@ -1,12 +1,10 @@
 @mixin rounded-image-size($size) {
-  figure,
+  border-radius: 50%;
```
GP-Dan-Tovbein, line 2: This change causes this [issue](https://www-dev.greenpeace.org/test-oberon/51128-2/)

GP-Dan-Tovbein (changes requested):

planet-4:

:watch: 2022.07.07 14:03:12

assets/src/styles/blocks/core-overrides/Image.scss
mleray: I've also added the `.force-no-caption` class to the selector, since the issues you noticed occur when there is a caption. This class is now present in all our patterns, so it should work where we expect it!

PR #897, opened 2022-07-19
PLANET-6848: Articles block - Manual override does not work on Post

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-6846

For `postType === 'post'`, we add an exclude_post_id query parameter to excclude the current one.
It forces the use of a more generic db query, even if specific postts IDs to include are specified.


GP-Dan-Tovbein (approved):
Code looks good! I've also tested through this [new demo page](https://www-dev.greenpeace.org/test-venus/?p=53111)

PR #898, opened 2022-07-19
PLANET-6527: New news & stories page layout pattern

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6527

- Add default heading


sagarsdeshmukh:
The name news & stories confuses with taxonomy names, can we use other naming like blog-page or just blog (cause it include all post type posts) ?

PR #900, opened 2022-07-21
Update carousel header styles

GP-Dan-Tovbein:
## Description
These styles are rendered an unnecessary white space to its bottom and fix the reported issue from [here](https://jira.greenpeace.org/browse/PLANET-6524).

From my point of view, we would need to keep the carousel header with its natural height without adding any extra space in its bottom. That space could be added for instance by a `Spacer` block.

Related PR: #884 

[Demo page of the current Homepage implementation](https://www-dev.greenpeace.org/test-phobos/51534-2/)

### Before
Screenshot 2022-07-21 at 12 15 54

### After
Screenshot 2022-07-21 at 12 16 39

## Demo pages
- [With margin to bottom](https://www-dev.greenpeace.org/test-neptune/)
- [Without margin to bottom](https://www-dev.greenpeace.org/test-neptune/33089-2/)

I'm open for suggestions :) What do you all think?

### DIFF
--- a/assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss
+++ b/assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss
@@ -78,6 +78,16 @@ $medium-image-height: 600px;
 .carousel-header {
   position: relative;
   width: 100vw;
+
+  &.is-style-fit-height-to-content {
+    &.block {
+      display: block;
+    }
+
+    &.block-header {
+      margin-bottom: 0;
+    }
+  }
 }
 
 .carousel-inner {
--- a/assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js
+++ b/assets/src/blocks/CarouselHeader/CarouselHeaderBlock.js
@@ -46,6 +46,13 @@ export const registerCarouselHeaderBlock = () =>
       html: false, // Disable "Edit as HTMl" block option.
     },
     attributes,
+    styles: [
+      {
+        name: 'fit-height-to-content',
+        label: __('Fit height to content', 'planet4-blocks-backend'),
+        isDefault: false,
+      },
+    ],
     edit: CarouselHeaderEditor,
     save: ({ attributes }) => {
       const markup = ReactDOMServer.renderToString(<div


assets/src/styles/blocks/CarouselHeader/CarouselHeaderStyle.scss
```diff
@@ -78,6 +78,14 @@ $medium-image-height: 600px;
 .carousel-header {
   position: relative;
   width: 100vw;
+
+  &.block {
+    display: block;
+  }
+
+  &.block-header {
+    margin-bottom: 0;
```
GP-Dan-Tovbein, line 87: This value reset the default one from the [block styles](https://github.com/greenpeace/planet4-master-theme/blob/master/assets/src/scss/layout/_blocks.scss#L15-L18) which is causing the next layout issue.

Screenshot 2022-07-21 at 12 15 54

planet-4:

:watch: 2022.07.21 16:56:18

comzeradd (approved):
> From my point of view, we would need to keep the carousel header with its natural height without adding any extra space in its bottom. That space could be added for instance by a `Spacer` block.

This space is inherited from `.block`, so it's not specific to Carousel Header. It's something we add on all blocks.

Which in this cases seems odd because the Issues pattern has a grey background and it comes right after a carousel. 

But if the design team is okay that this will remove the space from all occurrences of the Carousel Header, then it's fine.

GP-Dan-Tovbein:
> > From my point of view, we would need to keep the carousel header with its natural height without adding any extra space in its bottom. That space could be added for instance by a `Spacer` block.
> 
> This space is inherited from `.block`, so it's not specific to Carousel Header. It's something we add on all blocks.
> 
> Which in this case seems odd because the Issues pattern has a grey background and it comes right after a carousel.
> 
> But if the design team is okay that this will remove the space from all occurrences of the Carousel Header, then it's fine.

[Slack thread](https://greenpeace-gpi.slack.com/archives/G015K63081W/p1658487951119669) with the design team.

GP-Dan-Tovbein:
@comzeradd I've updated the PR and also de PR description for more info.

comzeradd (approved):
💯

PR #904, opened 2022-07-26
Add background color param to Quick Links pattern

mleray:
### Description

This pattern should have grey 5% as background color by default, but for some pattern layouts such as [Action](https://jira.greenpeace.org/browse/PLANET-6529) or [High-Level Topic](https://jira.greenpeace.org/browse/PLANET-6737), we need it to have white background

### Testing

On local, you can try manually adding the `background_color` attribute in the pattern's params (using a color from our [palette](https://github.com/greenpeace/planet4-master-theme/blob/master/theme.json#L12-L84)) and saving. After that, when adding the pattern to a page, you should see it with the chosen background color. It should also work as expected when there is no background color specified in the params, meaning it should have `grey-05` background color applied by default 🙂


lithrel (approved):
:+1:

PR #907, opened 2022-07-26
Fix Media & Text block padding

mleray:
### Description

Before this fix, the block looks weird when it has a background or when it has full alignment (which will be needed for the [High-Level Topic](https://jira.greenpeace.org/browse/PLANET-6737)), especially with the image on the right:

Screenshot 2022-07-27 at 09 58 41

When the block is full width, the content part should follow our grid.

### Testing

You can either test the changes on local, or check [this page](https://www-dev.greenpeace.org/test-titan/media-text-improvements/) and compare it with the "broken" version [here](https://www-dev.greenpeace.org/test-leda/1090-2/). Please make sure you check all screen sizes, and also the RTL direction 🙂

### DIFF
--- a/assets/src/styles/blocks/core-overrides/_media-text.scss
+++ b/assets/src/styles/blocks/core-overrides/_media-text.scss
@@ -1,12 +1,26 @@
+@mixin adapt-padding-inline-start($containerWidth) {
+  padding-inline-start: calc(100% - (#{$containerWidth} / 2) + 0.75rem);
+}
+
+@mixin adapt-padding-inline-end($containerWidth) {
+  padding-inline-end: calc(100% - (#{$containerWidth} / 2) + 0.75rem);
+}
+
 .wp-block-media-text {
   &:not(.force-no-lightbox) img {
     cursor: pointer;
   }
 
+  // 601px is when the block is no longer stacked
   @media (max-width: 600px) {
+    &.alignfull .wp-block-media-text__content {
+      max-width: 540px;
+      width: 100%;
+      margin: auto;
+    }
+
     .wp-block-media-text__content {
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.07.27 07:47:39

assets/src/styles/blocks/core-overrides/_media-text.scss
```diff
@@ -1,12 +1,26 @@
+@mixin adapt-padding-inline-start($containerWidth) {
+  padding-inline-start: calc(100% - (#{$containerWidth} / 2) + 0.75rem);
+}
+
+@mixin adapt-padding-inline-end($containerWidth) {
+  padding-inline-end: calc(100% - (#{$containerWidth} / 2) + 0.75rem);
```
lithrel, line 6: I'm wondering if, because of the alignfull declaration, the content should stick to the same vertical line as our logo, instead of giving it a little `+ 0.75rem` gap mirroring the content padding. Any thought on this ? I'm thinking about this because the [page header design](https://zeroheight.com/05f6e9516/v/latest/p/1285c9-page-header/b/88c480) has its content aligned on the logo (and it's using the media/text block too, so similar situation)

mleray: I'm not sure, that would need to be discussed with the design team 🤔 Mag is the one who asked for this one to follow our grid. I guess the difference is that the page header is at the top of the page (so right below the logo) whereas this block will in most cases be placed in the middle of the page, so it would be strange to have it aligned with the logo and not with the rest of the content?

lithrel (approved):
:+1:

PR #908, opened 2022-07-27
Override pagination native functionality

GP-Dan-Tovbein:
## Description

Because of the new listing pages, we should override the WP native functionality.
- Change "Previous Page" button text to "Prev"
- Change "Next Page" button text to "Next"
- Show always both buttons despite of being in the first or last page.

Related to [#1759]( https://github.com/greenpeace/planet4-master-theme/pull/1759)

**Demo page**: We'll set the same as [#1759]( https://github.com/greenpeace/planet4-master-theme/pull/1759)

### Before
Screenshot 2022-07-27 at 11 13 01

### After
Screenshot 2022-07-27 at 11 12 22

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -440,6 +440,44 @@ $breakpoints = [
 ];
 
 add_filter(
+	'render_block_core/query-pagination-previous',
+	function ( $content, $parsed, $block ) {
+		if ( empty( $block->attributes['label'] ) ) {
+			$block->attributes['label'] = __( 'Prev', 'planet4-blocks' );
+			return $block->render();
+		}
+
+		// Check if the button isn't rendered, then return it.
+		if ( empty( $content ) ) {
+			return '<a href="/" class="wp-block-query-pagination-previous disabled">' . __( 'Prev', 'planet4-blocks' ) . '</a>';
+		}
+
+		return $content;
+	},
+	10,
+	3
+);
+
+add_filter(
+	'render_block_core/query-pagination-next',
+	function ( $content, $parsed, $block ) {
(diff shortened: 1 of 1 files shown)


planet-4:

:watch: 2022.07.27 14:52:30

planet4-gutenberg-blocks.php
```diff
@@ -439,6 +439,44 @@ function ( $std_class_object ) {
 	],
 ];
 
+add_filter(
+	'render_block_core/query-pagination-previous',
+	function ( $content, $parsed, $block ) {
+		if ( empty( $block->attributes['label'] ) ) {
```
lithrel, line 445: I just realized this could lead to an infinite loop if somebody saves `Prev` translation as an empty string :grin: , so you could replace this condition with `if ( ! array_key_exists( 'label', $block->attributes ) ) {`

planet4-gutenberg-blocks.php
```diff
@@ -439,6 +439,44 @@ function ( $std_class_object ) {
...
+			$block->attributes['label'] = __( 'Prev', 'planet4-blocks' );
+			return $block->render();
+		}
+
+		// Check if the button isn't rendered, then return it.
+		if ( empty( $content ) ) {
+			return '<a href="/" class="wp-block-query-pagination-previous disabled">' . __( 'Prev', 'planet4-blocks' ) . '</a>';
```
lithrel, line 452: This link would lead to the homepage, if we want no link we could remove the href attribute for example ?

GP-Dan-Tovbein: Yep! In fact, I've added the disable class to disable the link but I've just noticed that I forgot to add that rule 🤦

lithrel (approved):
:+1:

PR #909, opened 2022-07-27
Add alignfull param to Side Image with Text & CTA

mleray:
### Description

This is needed for some pattern layouts such as the [High-Level Topic one](https://jira.greenpeace.org/browse/PLANET-6737).

### Testing

On local, you can manually set `alignfull` as true in the pattern's params and then save. After that, when adding the pattern to a page, you should see it full width as expected!


classes/patterns/class-sideimagewithtextandcta.php
```diff
@@ -34,13 +34,16 @@ public static function get_config( $params = [] ): array {
 		$background_color     = $params['background_color'] ?? null;
 		$background_attribute = $background_color ? ',"backgroundColor":"' . $background_color . '"' : '';
 		$background_classes   = $background_color ? 'has-' . $background_color . '-background-color has-background' : '';
+		$alignfull            = $params['alignfull'] ?? false;
+		$alignment_class      = $alignfull ? 'alignfull' : '';
```
mleray, line 38: I don't think it's necessary to set it to `alignwide` if it's not full width, it seems to be added by default!

lithrel (approved):
:+1:

PR #912, opened 2022-07-29
Use TwoColumnsGravityForms template in Get Informed

mleray:
### Description

According to the [mockups](https://www.figma.com/file/leueo1LMrlPabJkZYs2NgW/IA%26nav_Final-mockups?node-id=1812%3A4965), the Get Informed pattern layout should also have a Gravity Forms with the possibility to add title/description on the side. Since we now have a template for this we need to use it here!


GP-Dan-Tovbein (approved):
Code looks good to me!

PR #913, opened 2022-07-29
Test on JS declared patterns

lithrel:
Cf [README file](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/b70350407c4ed50fb59889d08c63efe28fd4f311/assets/src/block-patterns/readme.md)

Using static output to declare patterns makes a lot of sense for editors to be able to exchange those patterns using just a  copy/paste - the [Block Pattern Directory](https://wordpress.org/patterns/) is exactly the goal of this.

But for developers, the way we do it currently requires a double declaration; for every change we make to various attributes of blocks, we also have to report them to the HTML output. This has several disadvantages:
- more work for reporting changes in both places
- more work for adding parameters to be able to extend those patterns without repetition
- error prone, using the wrong attribute output or value can break the pattern
- the block output is static, it won't evolve if the blocks used get new features or modifies its output with deprecations

This PR proposes to declare patterns only in the simplest JS declaration of blocks `[ block name, block attributes, inner blocks ]`, and use Gutenberg functions to generate the pattern HTML output when needed. 

## Features
- Block patterns declaration in JS
- HTML content generated at runtime from Gutenberg editor functions

### DIFF
--- a/assets/src/block-patterns/translate.js
+++ b/assets/src/block-patterns/translate.js
@@ -1,0 +1,103 @@
+import { _x } from '@wordpress/i18n';
+import { isEmpty } from 'lodash';
+
+function isObject( object ) {
+  return object !== null && typeof object === 'object';
+}
+
+/**
+ * Copy/paste of https://github.com/WordPress/gutenberg/blob/58c78ec5200a38401281c14186095bced2f6f12a/packages/blocks/src/api/registration.js#L296
+ *
+ * Translates block settings provided with metadata using the i18n schema.
+ *
+ * @param {string|string[]|Object[]} i18nSchema   I18n schema for the block setting.
+ * @param {string|string[]|Object[]} settingValue Value for the block setting.
+ * @param {string}                   textdomain   Textdomain to use with translations.
+ *
+ * @return {string|string[]|Object[]} Translated setting.
+ */
+function translateBlockSettingUsingI18nSchema(
+  i18nSchema,
+  settingValue,
+  textdomain
+) {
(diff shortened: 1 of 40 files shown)


assets/src/block-patterns/translate.js
```diff
@@ -0,0 +1,103 @@
...
+  i18nSchema,
+  settingValue,
+  textdomain
+) {
+  if ( typeof i18nSchema === 'string' && typeof settingValue === 'string' ) {
+    return _x( settingValue, i18nSchema, textdomain );
+  }
+  if (
+    Array.isArray( i18nSchema ) &&
+    ! isEmpty( i18nSchema ) &&
```
GP-Dan-Tovbein, line 29: Just a small comment, instead of invoking a lodash function you can check `i18nSchema` content through `i18nSchema.length` because you've already know that `i18nSchema` it's an array. Does it make sense?

planet-4:

:watch: 2022.08.03 10:05:25

GP-Dan-Tovbein:
@lithrel How should we lead with deprecated versions of patterns?

lithrel:
> How should we lead with deprecated versions of patterns?

As discussed in Slack,  don't think a pattern deprecates, once you click on it it's just a list of blocks, so deprecation of existing content will follow individual block deprecation system.
The deprecation/issues we'll face are about the block attributes, if they change in the blocks used we'll probably have an error in our declaration, and we'll have to update them.

assets/src/block-patterns/translate.js
lithrel: Yes :+1:

PR #915, opened 2022-08-04
PLANET-6828 Fix Covers carousel layout in Safari

mleray:
### Description

See [PLANET-6828](https://jira.greenpeace.org/browse/PLANET-6828)
By hiding some of the covers we make sure that they are updated in Safari too, but unfortunately we lose the smooth scrolling animation... On the plus side, the code is a bit simplified with this solution!

### Testing

You can test the changes on [this page](https://www-dev.greenpeace.org/test-venus/custom-paddings-demo-page/) for example, it should work in all browsers and also in all screen sizes!

### DIFF
--- a/assets/src/blocks/Covers/CampaignCovers.js
+++ b/assets/src/blocks/Covers/CampaignCovers.js
@@ -5,19 +5,16 @@ const { __ } = wp.i18n;
 
 export const CampaignCovers = ({
   covers,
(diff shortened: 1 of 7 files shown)


mleray:
Closing in favor of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/923 where we use Bootstrap for the carousel

PR #929, opened 2022-08-17
Replacement of side-image-with-text-and-cta pattern by a block template

lithrel:
Using [block templates](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-templates/) to declare bits of code to help composing pattern layouts.

It allows to delegate the HTML output generation to Gutenberg as a lazy expansion in the editor, while keeping the flexibility of attributes and inner blocks inclusion.


planet-4:

:new_moon: [deimos](https://www-dev.greenpeace.org/test-deimos) | [admin](https://www-dev.greenpeace.org/test-deimos/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-deimos/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-deimos) | [composer-local.json](https://github.com/greenpeace/planet4-test-deimos/blob/main/composer-local.json)

:watch: 2022.08.17 12:20:54

mleray (commented):
I just noticed an issue, that I can't seem to properly remove the pattern/block template now 🤔 If I try, it actually still shows up in the code editor and then when switching back to the visual editor it appears again... The only way seems to be deleting it from the code editor which is not ideal 😬

assets/src/block-templates/save.js
```diff
@@ -0,0 +1,6 @@
+const el = wp.element.createElement;
```
GP-Dan-Tovbein, line 1: Same suggestion as [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/929/files#r1011696458). Again, this is not required to change it 🙏

assets/src/block-templates/side-image-with-text-and-cta/block.json
```diff
@@ -0,0 +1,19 @@
+{
```
GP-Dan-Tovbein, line 1: I love this new way🥇

assets/src/block-templates/side-image-with-text-and-cta/block.json
```diff
@@ -0,0 +1,19 @@
...
+  "$schema": "https://schemas.wp.org/trunk/block.json",
+  "apiVersion": 2,
+  "name": "planet4-block-templates/side-image-with-text-and-cta",
+  "title": "Side image with text and CTA",
```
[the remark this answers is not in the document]
mleray: Probably not, if I remember correctly we previously agreed that since in the Handbook we use the English names for the blocks/patterns/etc, we shouldn't translate them in the backend 🤔 Otherwise it might be confusing for editors

assets/src/block-templates/side-image-with-text-and-cta/block.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Side image with text and CTA",
+  "category": "planet4-block-templates",
+  "textdomain": "planet4-blocks-backend",
+  "attributes": {
+    "title": { "type": "string" },
+    "titlePlaceholder": { "type": "string" },
+    "backgroundColor": { "type": "string" },
+    "alignFull": { "type": "bool" },
+    "mediaPosition": { "type": "string" }
```
GP-Dan-Tovbein, line 14: Do you know if it's possible to define somethink like `""mediaPosition: { "type": "right" | "left" }` like in Typescript?

assets/src/block-templates/side-image-with-text-and-cta/template.js
```diff
@@ -0,0 +1,28 @@
+import mainThemeUrl from '../main-theme-url';
+
+const { __ } = wp.i18n;
+
+const template = ({
+  title = null,
```
GP-Dan-Tovbein, line 6: I think we should set an empty `string` instead of `null`. Why? 
First, because the `core/heading` has already defined the `content` attribute as a `string` ([see the WP heading doc](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/heading/block.json#L15)) and [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/929/files#diff-03eabc1163fecc06781194aaf556684fd1c6b304204ab2b21c638e0363469c5cR20) you are passing the content as `null`.
And finally to have more consistency with the attribute types.

assets/src/block-templates/template-list.js
```diff
@@ -0,0 +1,7 @@
+import * as sideImgTextCta from './side-image-with-text-and-cta';
+
+const templateList = [
+  sideImgTextCta,
+];
+
+export default templateList;
```
GP-Dan-Tovbein, line 7: `export default [ sideImgTextCta ]` it's also fine and you'll avoid to create the [templateList const](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/929/files#diff-fd7451822976d0285f46eb1c7105919fa991e213e25af06a137251a9515a70e2R3).

planet4-gutenberg-blocks.php
```diff
@@ -140,6 +140,8 @@
 	'planet4-blocks/guestbook',
 	'leadin/hubspot-form-block',
 	'gravityforms/form',
+	// Block templates.
```
GP-Dan-Tovbein, line 143: We could also define a new list of `BLOCK_TEMPLATES`

GP-Dan-Tovbein (commented):
This is a nice job @lithrel 🥳 Personally, I think this is a really good improvement for us, better to understand, better to maintain, all it's better 😄 

However, I've left some comments, suggestions, ideas and looking forward to your feedback. Of course feel free to ignore the ones you see don't necessary to follow up !

assets/src/block-templates/side-image-with-text-and-cta/block.json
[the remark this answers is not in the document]
GP-Dan-Tovbein: I agree to don't translate `name`, but what about the `title`?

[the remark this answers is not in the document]
mleray: Yeah, I think because we use the English titles in the Handbook it would be confusing to have them translated in the editor. I think @sagarsdeshmukh worked on this?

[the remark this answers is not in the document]
sagarsdeshmukh: Yes, it was my proposal to keep the block name/title, style name etc in base language, to minimise naming confusion in different languages while searching them in handbook for more details. Also the [native Gutenberg block](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/button/block.json#L5) names/style names are not offer translation.

assets/src/block-templates/side-image-with-text-and-cta/block.json
lithrel: Yes good point, [the schema includes an `enum` attribute](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-attributes/), I'm gonna switch to it :+1:

assets/src/block-templates/edit.js
```diff
@@ -0,0 +1,11 @@
+const el = wp.element.createElement;
```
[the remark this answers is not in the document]
lithrel: Yep that works too :+1:  my JS is very naive, I mostly followed WP doc and Gutenberg source code :)

assets/src/block-templates/side-image-with-text-and-cta/block.json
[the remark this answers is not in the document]
lithrel: I did not implement it in this PR, but Gutenberg (and WP) has a system to translate the metadata in `block.json`, using the `textdomain` included in the metadata, this function [translateBlockSettingUsingI18nSchema](https://github.com/WordPress/gutenberg/blob/ec588b48861ac86a8f96589d3da3c50cacb406e5/packages/blocks/src/api/registration.js#L306-L356) during metadata reading (the same exists server side if I remember), and a file [describing the translation schema (fields to translate)](https://github.com/WordPress/gutenberg/blob/5d2e14a5d475fc62a2b343481dd65bc2dc5d4cad/packages/blocks/src/api/i18n-block.json) for blocks.

lithrel:
>  However I'm wondering if it's not a bit confusing to have these patterns listed twice in the editor? It will also end up being a very long list of options once we convert all of them sweat_smile

Right, not sure how to deal with that, we have to choose which one we want to show (block or pattern), and based on that decision either remove the pattern declaration or hide the block in the list. Pinging @comzeradd on that, maybe we should talk with the design team next meeting.

assets/src/block-templates/main-theme-url.js
```diff
@@ -0,0 +1 @@
+export default `${window.p4bk_vars.themeUrl || ''}`;
```
GP-Dan-Tovbein, line 1: Do we need to use template literals here? Since`themeUrl` it's always an `string` https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/e312e0d4d1c6c00acf6a0955932d8d5cf2ef6223/classes/class-loader.php#L404
I think `export default window.p4bk_vars.themeUrl || ''` it's fine.  Am I correct?

lithrel:
> I just noticed an issue, that I can't seem to properly remove the pattern/block template now thinking If I try, it actually still shows up in the code editor and then when switching back to the visual editor it appears again... The only way seems to be deleting it from the code editor which is not ideal grimacing

@mleray  Good catch :+1:  Fixed it, now it displays a block selector with basic options; I also removed the additional class field that's not used
Screenshot from 2022-11-07 18-07-12

assets/src/block-templates/side-image-with-text-and-cta/template.js
```diff
@@ -0,0 +1,28 @@
+import mainThemeUrl from '../main-theme-url';
+
+const { __ } = wp.i18n;
+
+const template = ({
+  title = '',
+  titlePlaceholder = __('Enter title', 'planet4-blocks-backend'),
```
mleray, line 7: I'm not sure we need this attribute 🤔 I feel like the placeholder will always be the same

planet4-gutenberg-blocks.php
```diff
@@ -276,6 +278,7 @@ function set_allowed_block_types( $allowed_block_types, $context ) {
 		PAGE_BLOCK_TYPES,
 		! Features::is_active( 'beta_blocks' ) ? [] : BETA_PAGE_BLOCK_TYPES,
 		! $enform_active ? [] : [ 'planet4-blocks/enform' ],
+		BLOCK_TEMPLATES,
```
mleray, line 281: Do we want them only allowed in pages? I would think that we also want them in Actions at least 🤔

lithrel:
Added the [supported property](https://raw.githubusercontent.com/WordPress/gutenberg/trunk/schemas/json/block.json) `inserter: false` so that the block template does not appear in blocks inserter list. For follow-ups:
- to hide a block, use `supports.inserter: false` (done automatically for block templates here)
- to hide a pattern, remove its class from `Block_Pattern::get_list`

assets/src/block-templates/side-image-with-text-and-cta/block.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://schemas.wp.org/trunk/block.json",
+  "apiVersion": 2,
+  "name": "planet4-block-templates/side-image-with-text-and-cta",
+  "title": "Side image with text and CTA",
+  "description": "Side image with text and CTA",
```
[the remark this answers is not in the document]
lithrel: Nope, only name and title are required, I'll remove that :)

assets/src/editorIndex.js
```diff
@@ -66,3 +66,6 @@ registerBlockVariation('core/group', {
   },
   icon: 'admin-links',
 });
+
+import { registerBlockTemplates } from './block-templates/register';
```
mleray, line 70: Not a big deal, but maybe this `import` statement should be with all the others at the beginning of the file?

mleray (approved):
Looks good to me 🎉 Such an improvement for patterns' creation & maintenance, thanks @lithrel for all the investigation and hard work!

GP-Dan-Tovbein (approved):
Code looks good to me! 

Excellent job mate 🥳 This new way will be very useful for all!

PR #931, opened 2022-08-24
PLANET-6853 Hide PageHeader overflow for parallax effect

mleray:
### Description

See [PLANET-6853](https://jira.greenpeace.org/browse/PLANET-6853)
Without this small CSS change, the image keeps scrolling down the page


planet-4:

:watch: 2022.08.24 11:40:00

lithrel (approved):
:+1:

classes/class-loader.php
```diff
@@ -110,6 +110,18 @@ private function __construct() {
 
 		// Load block patterns.
 		Block_Pattern::register_all();
+
+		add_action(
```
mleray, line 114: This of course will only be needed if we end up using the Rellax library ([see PR](https://github.com/greenpeace/planet4-master-theme/pull/1783)). I'm not sure if there's a better place to add this though 🤔

classes/class-loader.php
```diff
@@ -110,6 +110,18 @@ private function __construct() {
...
+			'wp_enqueue_scripts',
+			static function () {
+				if ( has_block( 'core/media-text' ) ) {
+					wp_enqueue_script(
+						'rellax',
+						'https://cdn.jsdelivr.net/gh/dixonandmoe/rellax@master/rellax.min.js'
```
comzeradd, line 120: We could use the [version from cdnjs](https://cdnjs.com/libraries/rellax), since we already use it for some libraries here and in the theme (to save one more dns request).

mleray: [Updated](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/931/commits/30855649e9157ad6cf1f403cd3b4ef4159a3a81b) 🙂

classes/class-loader.php
lithrel: The other conditional enqueues we have are each in their own block class (carouselheader, gallery), so I guess this one is fine here :) We'll probably rework/regroup those if we try to systematically split and load css/js conditionally.
You can also use `BlockList::has_block()` instead of `has_block()` if you want to handle the reusable block issue.

assets/src/styles/blocks/core-overrides/_media-text.scss
```diff
@@ -11,6 +11,11 @@
     cursor: pointer;
   }
 
+  &.is-style-parallax .wp-block-media-text__media {
+    overflow: hidden;
+    z-index: -1;
+  }
```
lithrel, line 17: This makes the image disappear in the editor for me, do you have this too ? (the z-index specifically, on firefox)

mleray: Good catch! I think that [this commit](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/931/commits/511de74b689fb7d615262a01aa0fe61d3de0d7a4) solves the problem 🙂

lithrel (approved):
:+1:

classes/class-loader.php
```diff
@@ -110,6 +110,22 @@ private function __construct() {
 
 		// Load block patterns.
 		Block_Pattern::register_all();
+
+		// Load parallax library for Media & Text block.
+		add_action(
+			'wp_enqueue_scripts',
+			static function () {
```
GP-Dan-Tovbein, line 117: What is the reason to implement a `static` function?

assets/src/components/Parallax/setupParallax.js
```diff
@@ -0,0 +1,21 @@
+export const setupParallax = () => {
+  const parallaxImages = document.querySelectorAll('.is-style-parallax img');
+
+  if (parallaxImages.length === 0) {
```
GP-Dan-Tovbein, line 4: `if (!parallaxImages.length) {` is also working.

assets/src/styles/blocks/PageHeader.scss
```diff
@@ -28,6 +28,12 @@ $text-column-padding-in: 24px;
     margin: -$sp-5x $sp-3 $sp-3;
     padding-inline-start: 0;
     padding-inline-end: 0;
+    z-index: 1;
+
+    .wp-block-button,
+    .wp-block-button a {
```
GP-Dan-Tovbein, line 34: You can use `wp-block-button__link` instead.

GP-Dan-Tovbein (approved):
Code looks good to me!
I've left some comments but nothing to block the PR.

GP-Dan-Tovbein:
Good job @mleray ! This is an amazing feature 🥳

assets/src/styles/blocks/PageHeader.scss
mleray: We could indeed, tbh I just copy-pasted this from the [Media & Text styles](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/master/assets/src/styles/blocks/core-overrides/_media-text.scss#L26-L30) since this is what these lines override

classes/class-loader.php
mleray: Tbh I don't know 😬 I've removed it since it seems to work fine without it!

lithrel: [In theory](https://www.php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static) you avoid any binding issue with the container class (and it might have a tiny tiny perf advantage), in practice for this case both are fine :)

mleray: Okay, thanks for the explanation Flo 👍 removed the commit!

GP-Dan-Tovbein: Interesting @lithrel and sorry @mleray for this confusion 🤭

PR #936, opened 2022-08-30
PLANET-6862: Enable reusable blocks for editors

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6862

## Description
This PR enables Editors to navigate through the reusable blocks page into the Admin panel.
I've also left more info into comments, please validate them 🙏 .

## Testing
Login within an Editor user and check if that user can navigate to `Blocks > Reusable blocks`

### DIFF
--- a/classes/controller/menu/class-settings-controller.php
+++ b/classes/controller/menu/class-settings-controller.php
@@ -22,11 +22,12 @@ if ( ! class_exists( 'Settings_Controller' ) ) {
 
 			$current_user = wp_get_current_user();
 
-			if ( in_array( 'administrator', $current_user->roles, true ) && current_user_can( 'manage_options' ) ) {
+			if ( ( in_array( 'administrator', $current_user->roles, true ) || in_array( 'editor', $current_user->roles, true ) ) && current_user_can( 'edit_posts' ) ) {
+
 				add_menu_page(
 					__( 'Blocks', 'planet4-blocks-backend' ),
 					__( 'Blocks', 'planet4-blocks-backend' ),
-					'manage_options',
+					'edit_posts',
 					P4GBKS_PLUGIN_SLUG_NAME,
 					null,
 					'dashicons-layout'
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2022.08.30 19:53:05

classes/controller/menu/class-settings-controller.php
```diff
@@ -22,11 +22,12 @@ public function create_admin_menu() {
 
 			$current_user = wp_get_current_user();
 
-			if ( in_array( 'administrator', $current_user->roles, true ) && current_user_can( 'manage_options' ) ) {
+			if ( ( in_array( 'administrator', $current_user->roles, true ) || in_array( 'editor', $current_user->roles, true ) ) && current_user_can( 'edit_posts' ) ) {
+
 				add_menu_page(
 					__( 'Blocks', 'planet4-blocks-backend' ),
 					__( 'Blocks', 'planet4-blocks-backend' ),
-					'manage_options',
```
GP-Dan-Tovbein, line 29: I changed this capability to `edit_posts` since Editors hasn't set this capability by default.

We can add that `cap` by `get_role( 'editor' )->add_cap( 'manage_options' )` but in that case it would be affected to other parts of the code.

We are also using the same as [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/936/files#diff-5b286082494da95cebb1e07ad153d42ff3053c22bcbdc673c3d95f0075cd2055L35)

comzeradd: Good call! I don't think we should add editors to `manage_options`.

But can also make a small adjustment so that editors can also see the two Beta reports. But without allowing the old "Usage" reports, since these are going to be replaced in favor of the Betas at some point.

comzeradd:
Looks good. But maybe also allow the current "Blocks" report too after all, since it's the default menu entry. So it's already being displayed and it's currently leading to 404 for editors.

GP-Dan-Tovbein:
> Looks good. But maybe also allow the current "Blocks" report too after all, since it's the default menu entry. So it's already being displayed and it's currently leading to 404 for editors.

@comzeradd I've update the PR with your requested feature.

comzeradd (approved):

PR #960, opened 2022-10-20
Fix Carousel Header console error

mleray:
### Description

This seems to be because of the data-reactroot attribute


assets/src/blocks/CarouselHeader/deprecated/carouselHeaderV2.js
```diff
@@ -0,0 +1,14 @@
+import ReactDOMServer from 'react-dom/server';
+import { CarouselHeaderFrontend } from '../CarouselHeaderFrontend';
+
+export const carouselHeaderV2 = {
+  save: () => ({ attributes }) => {
```
GP-Dan-Tovbein, line 5: Don't you need to add `attributes` despite of not being modified?

PR #965, opened 2022-11-02
PLANET-6951: Implement new filter hook applied to query pagination

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6951

**Demo page**: You can test this functionality trough you local environment and test a listing pagination of campaigns, or any other listing page with only 1 page or wait after [#1812](https://github.com/greenpeace/planet4-master-theme/pull/1812) is being merged.

### DIFF
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -448,7 +448,7 @@ add_filter(
 			str_contains( $content, 'wp-block-query-pagination-previous disabled' )
 			&& str_contains( $content, 'wp-block-query-pagination-next disabled' )
 		) {
-			return '<div class="wp-block-query-pagination is-empty"></div>';
+			return null;
 		}
 
 		return $content;
--- a/planet4-gutenberg-blocks.php
+++ b/planet4-gutenberg-blocks.php
@@ -441,6 +441,23 @@ $breakpoints = [
 ];
 
 add_filter(
+	'render_block_core/query-pagination',
+	function ( $content ) {
+		// Check if prev and next buttons are disabled.
+		if (
+			str_contains( $content, 'wp-block-query-pagination-previous disabled' )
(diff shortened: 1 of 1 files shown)


planet4-gutenberg-blocks.php
```diff
@@ -440,6 +440,23 @@ function ( $std_class_object ) {
...
 
+add_filter(
+	'render_block_core/query-pagination',
+	function ( $content ) {
+		// Check if prev and next buttons are disabled.
+		if (
+			str_contains( $content, 'wp-block-query-pagination-previous disabled' )
+			&& str_contains( $content, 'wp-block-query-pagination-next disabled' )
+		) {
+			return '<div class="wp-block-query-pagination is-empty"></div>';
```
mleray, line 451: Maybe instead of creating this new `is-empty` class, you could use Boostrap's `d-none` that does what you want? Then [the other PR](https://github.com/greenpeace/planet4-master-theme/pull/1830) would no longer be necessary 🙂 
Or actually maybe you could just return `null` since the pagination is not needed at all in this case 🤔

GP-Dan-Tovbein: Yes! that make sense, good point! we can return an empty string too. 
But the main reason of why I do that, is because we can style directly through CSS and keep the element. A same situation happens with the Next and Prev button. We've added the `disabled` instead of don't return them.
I'll ask the design team of what do they prefer.

mleray: I think that was in case you are on the first or last page, the design team wanted the prev/next controls to still be shown but disabled. This is different since there's only one page 🤔

mleray (approved):
Looks good to me 👍 But maybe the commit's name could be a bit more clear? Maybe something like `Hide pagination on listing pages with only one page`

PR #973, opened 2022-11-17
Fix warning on block templates category missing

lithrel:
Adding a new block template in the editor generates a Warning message in the browser console, saying that the block category is invalid.

`The block "planet4-block-templates/side-image-with-text-and-cta" is registered with an invalid category "planet4-block-templates".`

## Fix 

- Declare missing block templates category


sagarsdeshmukh (approved):
Warning message vanish from console log. 👍

PR #983, opened 2022-12-12
Delete theme obsolete folder

comzeradd:
This seems like a remnant of the theme editor


lithrel (commented):
Removing only this file seems to break the Campaign editor sidebar, I think we need to also remove reference to it in [LocalThemeSettings](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/components/Sidebar/LocalThemeSettings.js#L14-L25) and/or use of that component in [CampaignThemeSidebar](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/components/Sidebar/CampaignThemeSidebar.js#L159)

PR #988, opened 2022-12-26
Convert Issues pattern to a block template

GP-Dan-Tovbein:
Ref: [PLANET-6942](https://jira.greenpeace.org/browse/PLANET-6942)

## Description
- Converted the issues pattern to a block template
- Replaced the `core/media-text` to `core/image` which works better with media sizes (`width`, `height`). The `core/media-text` works with `%` and it's difficult to follow exactly the UI. 

[Demo page](https://www-dev.greenpeace.org/test-rhea/convert-issues-pattern-to-a-block-template/)

### DIFF
--- a/assets/src/block-templates/issues/template.js
+++ b/assets/src/block-templates/issues/template.js
@@ -1,23 +1,22 @@
 import mainThemeUrl from '../main-theme-url';
 
 const item = ['core/group', {
-  className: 'd-flex',
   backgroundColor: 'white',
   style: {
-    border: { radius:'4px' },
+    border: { radius: '4px' },
     spacing: {
       padding: {
-        top:'32px',
-        right:'16px',
-        bottom:'32px',
-        left:'16px',
+        top: '32px',
+        right: '16px',
+        bottom: '32px',
+        left: '16px',
       }
     },
   },
   layout: {
     type: 'flex',
     flexWrap: 'nowrap',
-    justifyContent: "left",
+    justifyContent: 'left',
     orientation: 'horizontal',
   },
 }, [
@@ -30,7 +29,7 @@ const item = ['core/group', {
     level: 5,
     className: 'w-auto',
     style: {
-      typography: { fontSize: '16px' },
+      typography: { fontSize: '1rem' },
       spacing: {
         margin: { top: '0px', bottom: '0px', left: '16px' },
       },
@@ -82,14 +81,14 @@ const template = ({
           [ 'core/group', {
               className: 'is-style-space-evenly',
               layout: {
-                type: "flex",
+                type: 'flex',
                 allowOrientation: false,
               },
               style: {
                 spacing: {
                   padding: {
-                    top: "40px",
-                    bottom: "56px",
+                    top: '40px',
+                    bottom: '56px',
                   },
                 },
               },
@@ -98,8 +97,8 @@ const template = ({
           ],
           [ 'core/buttons', {
               layout: {
-                type: "flex",
-                justifyContent: "center",
+                type: 'flex',
+                justifyContent: 'center',
               },
             }, [
             [ 'core/button', { placeholder: __( 'Enter text', 'planet4-blocks-backend' ) } ]]],
--- a/assets/src/block-templates/issues/template.js
+++ b/assets/src/block-templates/issues/template.js
@@ -1,0 +1,115 @@
+import mainThemeUrl from '../main-theme-url';
+
+const item = ['core/group', {
+  className: 'd-flex',
+  backgroundColor: 'white',
+  style: {
+    border: { radius:'4px' },
+    spacing: {
+      padding: {
+        top:'32px',
+        right:'16px',
+        bottom:'32px',
+        left:'16px',
+      }
+    },
+  },
+  layout: {
+    type: 'flex',
+    flexWrap: 'nowrap',
+    justifyContent: "left",
+    orientation: 'horizontal',
+  },
+}, [
+  ['core/image', {
+    className: 'force-no-lightbox force-no-caption my-0 w-40',
+    url: `${mainThemeUrl}/images/placeholders/placeholder-40x40.jpg`,
+    alt: __('Enter text', 'planet4-blocks-backend'),
+    width: 40,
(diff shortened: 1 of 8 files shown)


planet-4:

:new_moon: [rhea](https://www-dev.greenpeace.org/test-rhea) | [admin](https://www-dev.greenpeace.org/test-rhea/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-rhea/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-rhea) | [composer-local.json](https://github.com/greenpeace/planet4-test-rhea/blob/main/composer-local.json)

:watch: 2022.12.26 17:35:57

assets/src/block-templates/issues/template.js
```diff
@@ -0,0 +1,115 @@
...
+  },
+  layout: {
+    type: 'flex',
+    flexWrap: 'nowrap',
+    justifyContent: "left",
+    orientation: 'horizontal',
+  },
+}, [
+  ['core/image', {
+    className: 'force-no-lightbox force-no-caption my-0 w-40',
```
GP-Dan-Tovbein, line 25: I've needed to add the `.w-40`, which is not necessary, but it force the width to `40px` whenever any of the images are changed. In some cases, those images are not `40x40` pixels.

mleray: I didn't know about these classes 🤯 but it's a really good solution for the image sizes! The only "problem" is that the image's height and width in the sidebar don't match with it, but I'm not sure we can fix that 😕

assets/src/block-templates/issues/template.js
```diff
@@ -0,0 +1,115 @@
...
+    url: `${mainThemeUrl}/images/placeholders/placeholder-40x40.jpg`,
+    alt: __('Enter text', 'planet4-blocks-backend'),
+    width: 40,
```
mleray, line 28: I think that we can remove these `width` and `height` attributes, because they are overridden as soon as the image is changed (and for the placeholder image they would be set to 40 anyway since it's the correct size)

assets/src/block-templates/issues/template.js
mleray: Or maybe we can using a similar technique to [what we used](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/components/Image/ImageBlockEdit.js#L27-L32) for the `is-style-rounded-` classes 🤔

assets/src/styles/blocks/core-overrides/Image.scss
```diff
@@ -34,6 +34,10 @@
     display: none !important;
   }
 
+  &.w-40 {
```
mleray, line 37: Wouldn't we need to also fix the height to 40px? 🤔

GP-Dan-Tovbein: I did not add the height since we assume that the pictures there will be all square. 
However, following up your good suggestion I've replaced the `.w-40` with `.square-40` which is more consistent.

mleray: Is there a downside to adding the height too? Especially if we're gonna rename it `square-40`, it seems it would make sense 🤔

GP-Dan-Tovbein: It's been added [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/5ba2ac61f944a9f42c399e1288d58ab1beb7951f/assets/src/styles/blocks/core-overrides/Image.scss#L17-L18). Does it make sense?

mleray: Ah, yes, sorry 🤦‍♀️

assets/src/block-templates/issues/template.js
```diff
@@ -0,0 +1,113 @@
+import mainThemeUrl from '../main-theme-url';
+
+const item = ['core/group', {
+  className: 'd-flex',
```
mleray, line 4: Do we need to set **both** the `d-flex` class and the layout type `flex`? 🤔

assets/src/block-templates/issues/template.js
```diff
@@ -0,0 +1,112 @@
...
+  ['core/image', {
+    className: 'force-no-lightbox force-no-caption my-0 square-40',
+    url: `${mainThemeUrl}/images/placeholders/placeholder-40x40.jpg`,
+    alt: __('Enter text', 'planet4-blocks-backend'),
+  }],
+  ['core/heading', {
+    level: 5,
+    className: 'w-auto',
+    style: {
+      typography: { fontSize: '16px' },
```
mleray, line 32: Maybe better to use `rem` unit for font sizes?

assets/src/block-templates/issues/template.js
```diff
@@ -0,0 +1,112 @@
+import mainThemeUrl from '../main-theme-url';
+
+const item = ['core/group', {
+  backgroundColor: 'white',
+  style: {
+    border: { radius:'4px' },
```
mleray, line 6: Not a big deal, but there's a space missing here between `radius:` and `4px`

mleray: Also in the padding settings below 🙂

assets/src/block-templates/issues/template.js
```diff
@@ -0,0 +1,112 @@
...
+    spacing: {
+      padding: {
+        top:'32px',
+        right:'16px',
+        bottom:'32px',
+        left:'16px',
+      }
+    },
+  },
+  layout: {
```
mleray, line 16: I don't see these layout settings in the sidebar after adding the pattern, do you know where they can be edited?

GP-Dan-Tovbein: Do you mean this?
Screenshot 2023-01-16 at 13 41 39

mleray: Aaaaah yes, ok it's only for Row blocks! Thanks 👍

mleray (approved):

PR #976, opened 2022-11-29
PLANET-3641 Implement a more strict ESLint config to achieve consistency

Osong-Michael:
### Description

See [PLANET-3641](https://jira.greenpeace.org/browse/PLANET-3641)

- I added this [package](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-eslint-plugin/) to our ESLint config file to make sure our standards match what we use for the PHP files already.

- This PR will contain a few commits as my aim is to group the folders and fix the errors gradually so it makes it easier for review.

_**The final suggestions based on this investigation can be found in this [document](https://docs.google.com/document/d/1q7spckzW2cNEOTaLtPZRM8aIjtz_1QkrFm4Iw2ebXQ0).**_

**This [PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1006) contains the fixes for this Investigation.**

PR #1006, opened 2023-02-09
This commit contains all the fixes necessary for ESLint

Osong-Michael:
### Reference PR: [976](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/976)

There are many changes in this commit. They mainly include:

- Indentation fixes
- Removal of space inside _parans_, _curly brackets_ and _square brackets_
- Refactoring some of the code(like changing Class components to Functional components)
- Changing import statements for some file that referred to the `react` package to `@wordpress/element`.

**Testing**

For testing you can pull this branch locally and run `npx eslint assets/src`.


planet-4:

:watch: 2023.02.09 16:08:46

GP-Dan-Tovbein:
I've tested it locally by installing again the new packages and I got these warnings but I'm not sure if are required to fix them.

```
/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/blocks/ENForm/ENForm.js
  12:3  warning  ServerSideRender not found in '@wordpress/components'  import/named

/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/blocks/ENForm/ENFormInPlaceEdit.js
  4:9  warning  ToolbarGroup not found in '@wordpress/components'  import/named

/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/blocks/Happypoint/HappypointEditor.js
  27:3  warning  ToolbarGroup not found in '@wordpress/components'  import/named

/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
  11:3  warning  ServerSideRender not found in '@wordpress/components'  import/named

/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/blocks/SubPages/SubPages.js
  3:3  warning  ServerSideRender not found in '@wordpress/components'  import/named

/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
  8:3  warning  ToolbarGroup not found in '@wordpress/components'  import/named

/Users/dantovbein/Documents/Clients/Greenpeace/International/Projects/planet4-docker-compose/persistence/app/public/wp-content/plugins/planet4-plugin-gutenberg-blocks/assets/src/components/TermSelector/TermSelector.js
  20:9  warning  useDebounce not found in '@wordpress/compose'  import/named
```

What do you think?

assets/src/ImageBlockExtension.js
```diff
@@ -92,25 +92,25 @@ const addExtraControls = function() {
               title={ __( 'Planet4 Image Options' ) }
               initialOpen={ true }
             >
-              <label>Caption alignment</label>
+              <label htmlFor>Caption alignment</label>
```
GP-Dan-Tovbein, line 95: It seems weird that `React` accepts `htmlFor` as this way.
What do you think?

assets/src/ImageBlockExtension.js
```diff
@@ -92,25 +92,25 @@ const addExtraControls = function() {
...
+                      key={key}
+                      value={option.value}
+                      onClick={() => {
+                        props.setAttributes({
+                          captionAlignment: option.value,
+                        });
+                        updateCaptionAlignment(option.value);
+                      }}
+                      isPrimary={ captionAlignment === option.value }
+                      isSecondary={ captionAlignment !== option.value }>
```
GP-Dan-Tovbein, line 109: Thinking loud, why we leave an empty space between `{` and its value, but in other places we don't use that rule? 
For example [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1006/files#diff-28439af7370575789c096bf0633de7a826ac204b2f6341ee7115939709452ca3R100)
`value={option.value}` but `isPrimary={ captionAlignment === option.value }`

assets/src/block-templates/reality-check/template.js
```diff
@@ -1,31 +1,33 @@
...
 ]];
 
 const template = () => ([
   ['core/columns', {
-    className: 'block'
+    className: 'block',
   }, [
-    column, column, column
-  ]]
+    column, column, column,
```
GP-Dan-Tovbein, line 29: This is not part of the scope but we can do `[...Array(3).keys()].map(() => column)]` also here!

assets/src/block-templates/register.js
```diff
@@ -22,10 +22,10 @@ export const registerBlockTemplates = ( blockTemplates ) => {
   const templates = blockTemplates || templateList;
   const postType = getCurrentPostType();
 
-  templates.map( ( blockTemplate ) => {
-    let { metadata, template, templateLock = false } = blockTemplate;
+  templates.forEach( ( blockTemplate ) => {
```
GP-Dan-Tovbein, line 25: We can also consider to replace to `for` instead. I don't think we'd need to `return` a value.

Osong-Michael:
> I've tested it locally by installing again the new packages and I got these warnings but I'm not sure if are required to fix them.
> 
> ```
> assets/src/blocks/ENForm/ENForm.js
>   12:3  warning  ServerSideRender not found in '@wordpress/components'  import/named
> 
> assets/src/blocks/ENForm/ENFormInPlaceEdit.js
>   4:9  warning  ToolbarGroup not found in '@wordpress/components'  import/named
> 
> assets/src/blocks/Happypoint/HappypointEditor.js
>   27:3  warning  ToolbarGroup not found in '@wordpress/components'  import/named
> 
> assets/src/blocks/SocialMediaCards/SocialMediaCards.js
>   11:3  warning  ServerSideRender not found in '@wordpress/components'  import/named
> 
> assets/src/blocks/SubPages/SubPages.js
>   3:3  warning  ServerSideRender not found in '@wordpress/components'  import/named
> 
> assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
>   8:3  warning  ToolbarGroup not found in '@wordpress/components'  import/named
> 
> assets/src/components/TermSelector/TermSelector.js
>   20:9  warning  useDebounce not found in '@wordpress/compose'  import/named
> ```
> 
> What do you think?

Yes I found it weird too, especially since the Documentation states that those files are imported from that folder.

assets/src/block-templates/register.js
Osong-Michael: That is why I am using the forEach, since we are not returning anything.

assets/src/ImageBlockExtension.js
Osong-Michael: Oh, nice catch, strange it didn't return an error 🤔

assets/src/blocks/CarouselHeader/deprecated/carouselHeaderV1.js
```diff
@@ -15,18 +15,18 @@ export const carouselHeaderV1 = {
...
           link_url_new_tab: false,
-        }
+        },
       ],
     },
   },
-  isEligible({ slides }) {
-    return slides?.some(slide => typeof slide.header_size !== 'undefined');
+  isEligible({slides}) {
+    return slides?.some((slide) => typeof slide.header_size !== 'undefined');
```
GP-Dan-Tovbein, line 23: From my point of view, I preferred to use `( )` only if the `function` receives more than 1 `param`.

assets/src/blocks/Columns/ColumnsEditor.js
```diff
@@ -108,76 +109,78 @@ export const ColumnsEditor = ({ isSelected, attributes, setAttributes }) => {
...
             allowedFormats={[]}
           />
         </header>
       }
       {!isExample &&
         <RichText
-          tagName='p'
-          className='page-section-description'
+          tagName="p"
+          className="page-section-description"
```
GP-Dan-Tovbein, line 168: I'm not sure which I understand the rule of when we need to use `"` and when `'`

assets/src/blocks/Columns/ColumnsTasks.js
```diff
@@ -15,47 +15,47 @@ export const ColumnsTasks = ({ isCampaign, columns, no_of_columns }) => (
...
                   }
                 </h5>
               }
               {description &&
-                <p dangerouslySetInnerHTML={{ __html: description }} />
+                <p dangerouslySetInnerHTML={{__html: description}} />
               }
               {hasImage &&
-                <img src={attachment} alt='' loading='lazy' />
+                <img src={attachment} alt="" loading="lazy" />
```
GP-Dan-Tovbein, line 49: Thinking about the A11Y, we would need to consider to a add an `eslint` rule (if it exists) to validate empty `alt`. This is a very important attribute to respect.

assets/src/blocks/Cookies/CookiesFrontend.js
```diff
@@ -70,46 +70,46 @@ export const CookiesFrontend = props => {
...
 
     gtag('consent', 'update', {
-      [key]: granted ? 'granted' : 'denied',
+      [ key ]: granted ? 'granted' : 'denied',
     });
     dataLayer.push({
-      'event': 'updateConsent',
-      [key]: granted ? 'granted' : 'denied',
+      event: 'updateConsent',
+      [ key ]: granted ? 'granted' : 'denied',
```
GP-Dan-Tovbein, line 85: What was the main motivation to change `[key]` to `[ key ]` but `{ key }` to `{key}`?

assets/src/blocks/Columns/ColumnsEditor.js
Osong-Michael: So we use double quotes("") for jsx attributes and single quotes('') for strings. So if you are declaring a variable that is a string you do something like `const myVar = 'This string'; `

For an attribute, it will be `<span className="some-class", data-attribute="some-value">My Span</span>;`

assets/src/blocks/Cookies/CookiesFrontend.js
Osong-Michael: Thank you for pointing this out, I will take a look at it.

assets/src/blocks/Covers/CoversEditor.js
```diff
@@ -154,11 +154,11 @@ const renderView = (attributes, toAttribute) => {
           />
         </>
       }
-      {!loading && !covers.length && !isExample ?
-        <div className='EmptyMessage'>
+      {!loading && !covers.length && !isExample
+        ? <div className="EmptyMessage">
```
GP-Dan-Tovbein, line 158: We should find an `eslint` rule to validate this kind of `classnames`. 
Do you think if it's possible?

assets/src/blocks/Covers/CoversScript.js
```diff
@@ -1,7 +1,7 @@
-import { CoversFrontend } from './CoversFrontend';
+import {CoversFrontend} from './CoversFrontend';
 
-document.querySelectorAll( `[data-render='planet4-blocks/covers']` ).forEach(
-  blockNode => {
+document.querySelectorAll( '[data-render=\'planet4-blocks/covers\']' ).forEach(
```
GP-Dan-Tovbein, line 3: Is it invalid to change to
`document.querySelectorAll( '[data-render="planet4-blocks/covers"]' )` or `document.querySelectorAll( "[data-render='planet4-blocks/covers']" )`. I always disagree to `escape` chars when it can be avoided 🤭 .

GP-Dan-Tovbein (commented):
Amazing job 👏 
I've left some comments but not all of them are required to make a change. Some of them are doubts and point of views. Let me know what do you think!

assets/src/ImageBlockExtension.js
Osong-Michael: Yea, it was a nice hack to avoid the errors generated by ESLint.

assets/src/blocks/Columns/ColumnsTasks.js
Osong-Michael: That is true, but also thinking about the [documentation](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/alt-text.md), they state that if the image is used as an icon, button or control element, the `alt` prop should be an empty string. So we could look into that.

assets/src/blocks/Covers/CoversScript.js
Osong-Michael: Thank you, fixed.

assets/src/blocks/Columns/ColumnsTasks.js
GP-Dan-Tovbein: Here is another interesting explanation of when it's recommended (or not necessary) to set an empty string.

According to the [A11y project](https://www.a11yproject.com/posts/alt-text/)
> Use an empty alt attribute for any image that is decorative or not necessary for understanding the content of the page. For example, an image providing decoration to an article's heading could be hidden by giving the image an alt="" or simply alt

assets/src/BlockEditorValidation.js
```diff
@@ -10,37 +10,38 @@ let canPublish = true;
...
     const currentMessages = [];
-    const invalidBlocks = blocks.reduce( (invalidBlocks, block ) => {
+    // eslint-disable-next-line no-shadow
+    const invalidBlocks = blocks.reduce((invalidBlocks, block) => {
       // Normally `blocks` contains a valid list of blocks, however it can happen that one of them is `null` in rare
       // cases. It happened to me once while running with WordPress 5.8.1 and undoing multiple edits. This made the
       // editor crash while it's trying to access `block.name`.
       if (!block) {
-        return;
+        return null;
```
mleray, line 22: Are you sure this doesn't cause a crash? 🤔 According to the comment above `it can happen that one of them is null` was the issue prevented here

assets/src/block-templates/highlighted-cta/template.js
```diff
@@ -1,12 +1,12 @@
 import mainThemeUrl from '../main-theme-url';
 
 const template = ({
-  titlePlaceholder = 'Enter text'
+  titlePlaceholder = 'Enter text',
```
mleray, line 4: Not related to ESLint, but our other templates seem to offer translation here

assets/src/blocks/Articles/ArticlesEditor.js
```diff
@@ -81,12 +79,12 @@ const renderEdit = (attributes, toAttribute) => {
           {attributes.tags.length === 0 && attributes.post_types.length === 0 &&
             <div>
               <hr />
-              <label>{__('Manual override', 'planet4-blocks-backend')}</label>
+              <label htmlFor>{__('Manual override', 'planet4-blocks-backend')}</label>
```
mleray, line 82: Maybe instead of adding empty `htmlFor` attributes, we shouldn't be using labels for this? A paragraph for example would probably be fine

assets/src/blocks/Articles/ArticlesFrontend.js
```diff
@@ -33,26 +31,26 @@ export const ArticlesFrontend = (props) => {
...
-        { read_more_link ?
-          <div className="col-md-12 col-lg-5 col-xl-5">
+        { read_more_link
+          ? <div className="col-md-12 col-lg-5 col-xl-5">
             <a
               className="btn btn-secondary btn-block article-load-more"
-              href={ read_more_link }
-              target={ button_link_new_tab ? '_blank' : '' }
+              href={read_more_link}
+              target={button_link_new_tab ? '_blank' : ''} rel="noreferrer"
```
mleray, line 44: Shouldn't this `rel="noreferrer"` be added only when target is blank?

assets/src/blocks/Counter/CounterFrontend.js
```diff
@@ -1,41 +1,45 @@
...
-    const { completed_api } = this.props;
+    const {completed_api} = this.props;
     // Calculate completed and remaining values depending on props
-    let counter = this;
+    const counter = this;
     counter.calculateRemaining();
     // Add an eventListener to the window to enable instantly updating counters with supported APIs
     if (completed_api && completed_api.startsWith('https://')) {
-      window.addEventListener('updateCounter', function () { counter.calculateRemaining(); }, false);
+      window.addEventListener('updateCounter', () => {
```
mleray, line 23: Not a blocker, but this can probably be shortened to 
```
window.addEventListener('updateCounter', counter.calculateRemaining, false);
```

assets/src/components/FrontendBlockNode/FrontendBlockNode.js
```diff
@@ -7,10 +6,12 @@ import { Component } from '@wordpress/element';
...
-	render() {
-		return <div className={ this.props.className }
-			data-render={ this.props.blockName }
-			data-attributes={ JSON.stringify( this.props.attributes ) }></div>
-	}
-}
+
+export const FrontendBlockNode = () => {
+  return (
+    <div className={this.props.className}
```
mleray, line 12: You cannot use `this` outside of a Component, this causes crashes... Instead you probably need to add `props` in the params

assets/src/blocks/ENForm/ENFormInPlaceEdit.js
```diff
@@ -17,78 +17,80 @@ export const ENFormInPlaceEdit = ({attributes, setAttributes}) => {
...
-      icon: "awards",
+      id: 'thankyou',
+      icon: 'awards',
       title: __('Thank you message', 'planet4-blocks-backend'),
-    }
+    },
   ];
   const [activeTplId, setActiveTplId] = useState('signup');
-  const activeTpl = templates.find((tpl) => {return tpl.id === activeTplId});
+  const activeTpl = templates.find((tpl) => {
```
mleray, line 31: Not a blocker, but this can probably be shortened to 
```
const activeTpl = templates.find((tpl) => tpl.id === activeTplId);
```

assets/src/blocks/ENForm/ENFormInPlaceEdit.js
```diff
@@ -17,78 +17,80 @@ export const ENFormInPlaceEdit = ({attributes, setAttributes}) => {
...
               icon: tpl.icon,
               title: tpl.title,
               isActive: activeTplId === tpl.id,
-              onClick: () => { setActiveTplId(tpl.id) }
-            }
-          })
-        }
-      />
-    </BlockControls>
+              onClick: () => {
```
mleray, line 61: The brackets `{}` can be removed here I think

assets/src/blocks/ENForm/ENFormInPlaceEdit.js
```diff
@@ -99,77 +101,78 @@ const BackgroundImage = ({attributes}) => {
...
+          controls={['h1', 'h2'].map((size) => {
             return {
               isActive: title_size === size,
-              icon: "heading",
+              icon: 'heading',
               title: size.toUpperCase(),
-              onClick: () => { setAttributes({content_title_size: size}) }
-            }
-          })
+              onClick: () => {
```
mleray, line 145: Same comment here as line 61

assets/src/blocks/ENForm/ENFormInPlaceEdit.js
```diff
@@ -354,8 +357,10 @@ const FormContent = ({attributes, setAttributes, fields}) => {
             tag="button"
             id="p4en_form_save_button"
             className={'btn btn-primary btn-block' + (fwbg ? ' w-auto' : '')}
-            value={ button_text || __( 'Sign', 'planet4-engagingnetworks' ) }
-            onChange={(text) => {setAttributes({button_text: text})}}
+            value={button_text || __('Sign', 'planet4-engagingnetworks')}
+            onChange={(text) => {
```
mleray, line 361: Same comment here as line 61

assets/src/blocks/ENForm/ENFormInPlaceEdit.js
```diff
@@ -365,23 +370,27 @@ const FormContent = ({attributes, setAttributes, fields}) => {
                 value={text_below_button}
                 placeholder={__('Text below button', 'planet4-blocks-backend')}
                 allowedFormats={[]}
-                onChange={(text) => {setAttributes({text_below_button: text})}}
+                onChange={(text) => {
```
mleray, line 373: Same comment here as line 61

assets/src/blocks/ENForm/ENFormInPlaceEdit.js
```diff
@@ -365,23 +370,27 @@ const FormContent = ({attributes, setAttributes, fields}) => {
...
-        {! fwbg &&
+        {!fwbg &&
           <div className="enform-legal">
             <RichText
               tagName="p"
               value={text_below_button}
               placeholder={__('Text below button', 'planet4-blocks-backend')}
               allowedFormats={[]}
-              onChange={(text) => {setAttributes({text_below_button: text})}}
+              onChange={(text) => {
```
mleray, line 387: Same comment here as line 61

assets/src/blocks/ENForm/ENFormSettings.js
```diff
@@ -61,48 +58,52 @@ export const ENFormSettings = ({attributes, setAttributes}) => {
...
             value={en_page_id}
             options={[
-              { label: 'No pages', value: 0 },
-              ...page_list
+              {label: 'No pages', value: 0},
+              ...page_list,
             ]}
             disabled={page_list.length <= 0}
-            onChange={(id) => {setAttributes({en_page_id: parseInt(id)})}}
+            onChange={(id) => {
```
mleray, line 75: Brackets `{}` can be removed here too

assets/src/blocks/ENForm/ENFormSettings.js
```diff
@@ -61,48 +58,52 @@ export const ENFormSettings = ({attributes, setAttributes}) => {
...
+            label={__('Planet 4 Engaging Networks form', 'planet4-engagingnetworks-backend')}
+            value={en_form_id}
             options={[
-              { label: 'No forms', value: 0 },
-              ...en_forms
+              {label: 'No forms', value: 0},
+              ...en_forms,
             ]}
-            onChange={(id) => {setAttributes({en_form_id: parseInt(id)})}}
+            onChange={(id) => {
```
mleray, line 101: Same comment as line 75

assets/src/blocks/ENForm/ENFormSettings.js
```diff
@@ -113,7 +114,9 @@ export const ENFormSettings = ({attributes, setAttributes}) => {
             >
               <ImageOrButton
                 title={__('Select background image', 'planet4-blocks-backend')}
-                onSelectImage={(image) => {onBackgroundChange(image)}}
+                onSelectImage={(image) => {
```
mleray, line 117: Same comment as line 75

assets/src/blocks/ENForm/ENFormSettings.js
```diff
@@ -126,7 +129,9 @@ export const ENFormSettings = ({attributes, setAttributes}) => {
                     url={background_image_src}
                     dimensions={focal_picker_dimensions}
                     value={focus_bg_image_obj}
-                    onChange={(focus) => {onFocalChange('background_image_focus', focus)}}
+                    onChange={(focus) => {
```
mleray, line 132: Same comment as line 75

assets/src/blocks/ENForm/ENFormSettings.js
```diff
@@ -184,26 +190,26 @@ const getPageListByType = () => {
   }
 
   let flattenedPages = [];
-  for (var i in pages) {
-    let pagesByType = pages[i].map(page => {
-      return { label: page.name, value: page.id };
+  for (const i in pages) {
```
mleray, line 193: Here as well I believe it should be `let` instead of `const` 🤔

assets/src/blocks/ENForm/FormGenerator.js
```diff
@@ -88,155 +87,153 @@ const TextInput = ({field, onInputChange, onBlur, errors, is_side_style}) => {
...
-          onBlur={ (e) => onBlur(field, e) }
-        />
-        <span
-          className="custom-control-description"
-          dangerouslySetInnerHTML={{__html: `${field.label}${field.required ? ' *' : ''}`}}
-        />
-        {has_error &&
-          <div className="invalid-feedback">{ errors[field.id] ?? errorMessage }</div>
-        }
+        <label className={`custom-checkbox ${field.name === dependent_field ? 'disable-checkbox' : ''}`} htmlFor>
```
mleray, line 140: In this case the input has an id, maybe we can use it for the `htmlFor`?

assets/src/blocks/ENForm/FormGenerator.js
```diff
@@ -88,155 +87,153 @@ const TextInput = ({field, onInputChange, onBlur, errors, is_side_style}) => {
...
-  const errorMessage = __( 'This field is required', 'planet4-engagingnetworks' );
+  const question_option = {};
+  const has_error = errors && errors[ field.id ];
+  const errorMessage = __('This field is required', 'planet4-engagingnetworks');
 
   return (
     <div className={`en__field en__field--check en__field--${field.id}`}>
       <div className="en__field__element en__field__element--check form-group form-check-label-block custom-control p4-custom-control-input">
-        <label className="custom-checkbox">
+        <label className="custom-checkbox" htmlFor>
```
mleray, line 178: Same comment as line 140

assets/src/blocks/ENForm/FormGenerator.js
```diff
@@ -88,155 +87,153 @@ const TextInput = ({field, onInputChange, onBlur, errors, is_side_style}) => {
...
-  const has_error = errors && errors[field.id];
+  const options = field.radio_options[ field.locale ] || [];
+  const has_error = errors && errors[ field.id ];
 
   const inputs = options.map((opt, index) => {
     return (
       <div key={index} className={`en__field en__field--check en__field--${field.id}`}>
         <div className="en__field__element en__field__element--check form-group form-check-label-block custom-control p4-custom-control-input">
-          <label className="custom-radio">
+          <label className="custom-radio" htmlFor>
```
mleray, line 217: Same comment as line 140

assets/src/blocks/Articles/ArticlesFrontend.js
Osong-Michael: It should, but we are conditionally adding target blank but ESLint does not know that so it throws an error because it sees the target attribute.

assets/src/blocks/ENForm/ENForm.js
```diff
@@ -9,41 +9,37 @@ import {
...
+    const {getCurrentPostType} = wp.data.select('core/editor');
+    const currentPostType = getCurrentPostType();
 
     let flattenedPages = [];
     let pagesByType;
 
-    for (var i in window.p4en_vars.pages) {
-      pagesByType = window.p4en_vars.pages[i].map(page => {
-        return { label: page.name, value: page.id };
+    for (const i in window.p4en_vars.pages) {
```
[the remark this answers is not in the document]
Osong-Michael: So because this is a `for-in-loop`, `i` is initialised for each iteration so it is no problem, if we were using an imperative for loop where `i` is incremented, then we will use `let`.

assets/src/blocks/ENForm/ENFormSettings.js
Osong-Michael: Same thing too as above, `for-in-loop` re-initialises the variable for each iteration

assets/src/ImageBlockExtension.js
mleray: Maybe we can use paragraphs instead of labels if we have nothing to write in the `htmlFor`? 🤔

mleray: That might require some styling fixes too though

assets/src/blocks/Counter/CounterFrontend.js
```diff
@@ -1,41 +1,41 @@
...
-    const { completed_api } = this.props;
+    const {completed_api} = this.props;
     // Calculate completed and remaining values depending on props
-    let counter = this;
+    const counter = this;
     counter.calculateRemaining();
     // Add an eventListener to the window to enable instantly updating counters with supported APIs
     if (completed_api && completed_api.startsWith('https://')) {
-      window.addEventListener('updateCounter', function () { counter.calculateRemaining(); }, false);
+      window.addEventListener('updateCounter', counter.calculateRemaining(), false);
```
mleray, line 23: I think you need to remove the parenthesis, `counter.calculateRemaining()` => `counter.calculateRemaining`

assets/src/theme/p4ServerThemes.js
```diff
@@ -1,4 +1,4 @@
-/* global wp */
+// global wp
```
mleray, line 1: I'm not sure we actually need to declare this globally here 🤔 We use `wp` in lots of places without it!

assets/src/components/enform/InlineFormFeedback/InlineFormFeedback.js
```diff
@@ -1,7 +1 @@
-import {Component} from '@wordpress/element';
-
-export class InlineFormFeedback extends Component {
-  render() {
-    return <div className='InlineFormFeedback'>{this.props.children}</div>;
-  }
-}
+export const InlineFormFeedback = () => (<div className="InlineFormFeedback">{this.props.children}</div>);
```
mleray, line 1: You cannot use `this` here, you'll need to pass `props` as params

assets/src/components/enform/FormSectionTitle/FormSectionTitle.js
```diff
@@ -1,7 +1 @@
-import {Component} from '@wordpress/element';
-
-export class FormSectionTitle extends Component {
-  render() {
-    return <div className='FormSectionTitle'>{this.props.children}</div>;
-  }
-}
+export const FormSectionTitle = () => (<div className="FormSectionTitle">{this.props.children}</div>);
```
mleray, line 1: You cannot use this here, you'll need to pass props as params

assets/src/components/enform/FormHelp/FormHelp.js
```diff
@@ -1,7 +1 @@
-import {Component} from '@wordpress/element';
-
-export class FormHelp extends Component {
-  render() {
-    return <div className='FormHelp'>{this.props.children}</div>;
-  }
-}
+export const FormHelp = () => (<div className="FormHelp">{this.props.children}</div>);
```
mleray, line 1: You cannot use `this` here, you'll need to pass `props` as params

assets/src/components/Lightbox/Lightbox.js
```diff
@@ -28,13 +28,13 @@ export const Lightbox = ({ index, isOpen, items, onClose = () => {} }) => {
 
     const photoSwipe = new PhotoSwipe(photoSwipeElement, PhotoSwipeUI_Default, items, photoSwipeOptions);
 
-    photoSwipe.listen('gettingData', function (index, galleryItem) {
+    photoSwipe.listen('gettingData', (galleryItem) => {
```
mleray, line 31: The Lightbox no longer works as expected if you remove the `index` param, even if it's unused... It's probably a function from the library

assets/src/blocks/Timeline/TimelineEditorScript.js
```diff
@@ -3,20 +3,20 @@ import {
...
+import {URLInput} from '../../components/URLInput/URLInput';
+import {useScript} from '../../components/useScript/useScript';
+import {useStyleSheet} from '../../components/useStyleSheet/useStyleSheet';
+import {Timeline} from './Timeline';
+import {languages} from './TimelineLanguages';
+import {URLDescriptionHelp} from './URLDescriptionHelp';
+import {debounce} from 'lodash';
+
+const {RichText} = wp.blockEditor;
+const {__, _} = wp.i18n;
```
mleray, line 19: Isn't `_` from `lodash` and not from `wp.i18n`?

assets/src/blocks/SubPages/SubPagesBlock.js
```diff
@@ -16,16 +16,8 @@ export class SubPagesBlock {
...
-      } )( ( {
-               isSelected,
-               attributes,
-               setAttributes
-             } ) => {
-        return <SubPages/>;
-      } ),
-    } );
-  };
+      edit: withSelect(() => {})(() => <SubPages />),
```
mleray, line 19: Cannot this be written simply as `edit: SubPages`?

assets/src/blocks/CarouselHeader/useSlides.js
```diff
@@ -8,8 +8,12 @@ import { useState, useEffect } from '@wordpress/element';
  * - Adds an *exit* transition class for the `active` element
  * - Adds a listener for `ontransitionend` to the `active` element
  *
- * @param {Array} slidesRef
+ * @param {Array}  slidesRef
+ * @param {*}      lastSlide
+ * @param {*}      containerRef
  * @param {Object} options
+ *
+ * @return {*} not exactly sure
```
mleray, line 16: Is this an official return type? 😁 Can't we just put `object`?

assets/src/blocks/Counter/CounterBlock.js
```diff
@@ -52,45 +52,46 @@ export class CounterBlock {
...
+      },
 
-      ]
+    ];
 
     if (window.p4ge_vars.features.feature_engaging_networks) {
       styles.push({
         name: 'en-forms-bar',
-        label: __('Progress Bar inside EN Form', 'planet4-blocks-backend')
+        label: __('Progress Bar inside EN Form', 'planet4-blocks-backend'),
```
mleray, line 91: I'm pretty sure that this should not be translatable, like the other styles 🤔

assets/src/blocks/Counter/CounterFrontend.js
```diff
@@ -123,32 +126,32 @@ export class CounterFrontend extends Component {
...
+                <div className={`progress-bar ${style === 'en-forms-bar' ? 'enform-progress-bar' : ''}`} style={{width: `calc(${percent}% + 20px)`}} />
               </div>
             }
             {style === 'arc' &&
               <svg className="progress-arc" xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 24 14">
                 <path className="background" d="M 2 12 A 1 1 0 1 1 22 12" />
                 <path className="foreground" d="M 2 12 A 1 1 0 1 1 22 12"
                   strokeDasharray={arcLength}
-                  strokeDashoffset={`${(1 - percent / 100) * arcLength}`} />
+                  strokeDashoffset={`${((1 - percent) / 100) * arcLength}`} />
```
mleray, line 142: Are you sure this isn't supposed to be the other way around (`1 - (percent / 100)`)?

assets/src/blocks/ENForm/FormGenerator.js
```diff
@@ -250,22 +247,22 @@ const RadioInput = ({field, onInputChange, onBlur, errors}) => {
       </span><br />
       { inputs }
       {has_error &&
-        <div className="invalid-feedback">{ errors[field.id] ?? errorMessage }</div>
+        <div className="invalid-feedback">{ errors[ field.id ] ?? errors.errorMessage }</div>
```
mleray, line 250: For the other field types we define an `errorMessage` string, we don't get it from the `errors` object so I think we should do the same here?

assets/src/blocks/Gallery/GalleryEditor.js
```diff
@@ -152,51 +150,52 @@ const renderView = (attributes, setAttributes) => {
...
+      image_id_array.forEach((img_id) => {
+        const img_details = select('core').getMedia(img_id);
         if (img_details) {
-          image_urls_array[img_id] = img_details.source_url;
+          imageUrlsArray[ img_id ] = img_details.source_url;
         }
       });
     }
-    return { image_urls_array };
+    return {imageUrlsArray};
```
mleray, line 169: This should be `return {image_urls_array: imageUrlsArray};`

assets/src/blocks/Media/MediaEditor.js
```diff
@@ -40,40 +38,41 @@ const MediaInspectorOptions = ({ attributes, setAttributes }) => {
...
-          allowedTypes={["image"]}
-        />
-      </PanelBody>
-    </InspectorControls>
+      <MediaPlaceholder
+        labels={{title: __('Video poster image [Optional]', 'planet4-blocks-backend'), instructions: __('Applicable for .mp4 video URLs only.', 'planet4-blocks-backend')}}
+        icon="format-image"
+        onSelect={onSelectImage}
+        // eslint-disable-next-line no-console
+        onError={console.log('Error Selecting Image')}
```
mleray, line 58: Shouldn't this be written like `onError={() => console.log('Error Selecting Image')}`?

assets/src/blocks/Counter/CounterFrontend.js
Osong-Michael: I have no idea, but I will change it.

assets/src/blocks/CarouselHeader/useSlides.js
Osong-Michael: So I was looking at the function and I could not figure out what the return type was, the `*` just means any

assets/src/blocks/SubPages/SubPagesBlock.js
```diff
@@ -16,16 +15,8 @@ export class SubPagesBlock {
...
-      } )( ( {
-               isSelected,
-               attributes,
-               setAttributes
-             } ) => {
-        return <SubPages/>;
-      } ),
-    } );
-  };
+      edit: <SubPages />,
```
mleray, line 18: I think you can simplify it even more here and write `edit: SubPages`

assets/src/blocks/Counter/CounterBlock.js
Osong-Michael: Should I remove it, because I have no idea too.

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
```diff
@@ -116,69 +117,68 @@ export const TakeActionBoxoutEditor = ({
   const takeActionPageSelected = take_action_page && parseInt(take_action_page) > 0;
 
   if (loading || !actPageList.length) {
-    return __("Populating block's fields...", 'planet4-blocks-backend');
+    return __('Populating block\'s fields…', 'planet4-blocks-backend');
```
mleray, line 120: I'm wondering how escaping characters will affect translations 🤔 Will they show as new strings? Will editors see the `\` in them and get confused?

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutFrontend.js
```diff
@@ -13,15 +11,16 @@ export const TakeActionBoxoutFrontend = ({
...
   >
+    {/* eslint-disable-next-line jsx-a11y/anchor-has-content */}
     <a
       data-ga-category="Take Action Boxout"
       data-ga-action="Image"
       data-ga-label="n/a"
       className="cover-card-overlay"
       href={link}
-      {...newTab && link && { target: "_blank" }}
+      {...newTab && link && {target: '_blank'}}
```
mleray, line 23: Shouldn't we add `rel=noreferrer` here? And for the following ones too?

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutEditor.js
comzeradd: Minor issues, but better use 3 dots here at the end of the string, as it was originially, instead of a special character

assets/src/components/Sidebar/CampaignThemeSidebar.js
```diff
@@ -109,59 +109,59 @@ export class CampaignThemeSidebar extends Component {
...
+        this.setState({options});
         if (themeName) {
           try {
-            const response = await fetch(`${ themeJsonUrl + themeName.replace('-new', '') }.json`);
+            const response = await fetch(`${themeJsonUrl + themeName.replace('-new', '')}.json`);
             const theme = await response.json();
             applyChangesToDom(theme, []);
           } catch (e) {
-            console.log(`Failed loading config for ${ themeName }`);
+            throw `Failed loading config for ${themeName}`;
```
mleray, line 125: Maybe we should just disable the ESLint warning for `console.log` here instead? Not sure how `throw` behaves in comparison 🤔

assets/src/blocks/CarouselHeader/useSlides.js
mleray: Ah okay 🙂 the description should still be improved I think, maybe you can just write "functions for the carousel header slides"

assets/src/blocks/Counter/CounterBlock.js
mleray: Yes I think you can remove the translation function there

mleray (commented):
I still see some changes in the master-theme submodule and the `package-lock.json` that probably shouldn't be there 🤔 
For the submodule it can probably be fixed by running `git submodule update`

Osong-Michael:
> I still see some changes in the master-theme submodule and the `package-lock.json` that probably shouldn't be there 🤔 For the submodule it can probably be fixed by running `git submodule update`

For the master-theme submodule I have no idea, for the `package-lock.json`, the new packages that were added should be responsible for updating it.

assets/src/ImageBlockExtension.js
mleray: It turns out writing just `htmlFor` like this results in the following console warning:
```
Warning: Received `true` for a non-boolean attribute `htmlFor`.

If you want to write it to the DOM, pass a string instead: htmlFor="true" or htmlFor={value.toString()}.
```
so we should probably find another solution!

assets/src/components/TagSelector/TagSelector.js
```diff
@@ -1,63 +1,63 @@
...
-    }, [] );
+    }, []);
 
-    return tags.map( tag => tag.name );
+    return tags.map((tag) => tag.name);
   }
 
   render() {
-    const { tagSuggestions, onChange, label, placeholder, value, ...ownProps } = this.props;
+    const {tagSuggestions, label, placeholder, ...ownProps} = this.props;
```
mleray, line 42: We actually need to keep `onChange` and `value` here, even if unused, otherwise they are in `ownProps` and when we later pass it down, they override the ones what we set (I hope that makes sense? 😅)

assets/src/components/PostTypeSelector/PostTypeSelector.js
```diff
@@ -1,44 +1,44 @@
...
+    const postTypeIds = value.map((postTypeName) => {
+      return this.props.postTypes.find((postType) => postType.name === postTypeName).id;
     });
-    this.props.onChange( postTypeIds );
+    this.props.onChange(postTypeIds);
   }
 
   render() {
-    const { postTypes, onChange, label, placeholder, value, ...ownProps } = this.props;
+    const {postTypes, label, placeholder, value, ...ownProps} = this.props;
```
mleray, line 23: Same issue here as for `TagSelector`, we need to exclude `onChange` from the `ownProps` object to avoid overriding the function that we pass down

assets/src/components/TagSelector/TagSelector.js
Osong-Michael: Oh my bad, got it 👍

assets/src/blocks/Timeline/TimelineEditorScript.js
```diff
@@ -3,20 +3,20 @@ import {
...
-const { __ } = wp.i18n;
+import {InspectorControls} from '@wordpress/block-editor';
+
+import {URLInput} from '../../components/URLInput/URLInput';
+import {useScript} from '../../components/useScript/useScript';
+import {useStyleSheet} from '../../components/useStyleSheet/useStyleSheet';
+import {Timeline} from './Timeline';
+import {languages} from './TimelineLanguages';
+import {URLDescriptionHelp} from './URLDescriptionHelp';
+import {debounce, _} from 'lodash';
```
mleray, line 16: I know I'm the one who told you `_` comes from lodash, but there's actually a console error now:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'noConflict')
```
so I don't know where `_` is supposed to come from 😬

mleray: Maybe we are supposed to write `import {debounce, noConflict} from 'lodash'` and use it directly, without `_.` 🤔

Osong-Michael: @lithrel can definitely help us out here, he wrote the code 😅

mleray: I've tested it and `import {debounce, noConflict} from 'lodash'` works 🙂

Osong-Michael: No need for the underscore(`_`)?

mleray: Nope!

assets/src/blocks/Gallery/GalleryEditor.js
```diff
@@ -1,62 +1,60 @@
...
-      image_data.push({ url: value[key].url, focalPoint: { x: 0.5, y: 0.5 }, id: img_id });
+      image_ids.push(value[ key ].id);
+      const img_id = value[ key ].id;
+      imageData.push({url: value[ key ].url, focalPoint: {x: 0.5, y: 0.5}, id: img_id});
     }
-    setAttributes({ multiple_image: image_ids.join(',') });
-    setAttributes({ image_data });
-  }
+    setAttributes({multiple_image: image_ids.join(',')});
+    setAttributes({imageData});
```
mleray, line 34: This should be `setAttributes({image_data: imageData})`

assets/src/blocks/ENForm/FormGenerator.js
```diff
@@ -88,155 +87,153 @@ const TextInput = ({field, onInputChange, onBlur, errors, is_side_style}) => {
     >
       <div
         className="en__field__element en__field__element--text form-group animated-label"
-        style={field.en_type === 'GEN' ? {display: "flex", "flex-direction": "row"} : {display: "block"} }
+        style={field.en_type === 'GEN' ? {display: 'flex', 'flex-direction': 'row'} : {display: 'block'}}
```
mleray, line 90: Here it should be `flexDirection` instead of `'flex-direction'`, currently there is an error: 
```
Warning: Unsupported style property flex-direction. Did you mean flexDirection?
```

assets/src/components/PostTypeSelector/PostTypeSelector.js
```diff
@@ -1,44 +1,43 @@
...
+    const postTypeIds = value.map((postTypeName) => {
+      return this.props.postTypes.find((postType) => postType.name === postTypeName).id;
     });
-    this.props.onChange( postTypeIds );
+    this.props.onChange(postTypeIds);
   }
 
   render() {
-    const { postTypes, onChange, label, placeholder, value, ...ownProps } = this.props;
+    // eslint-disable-next-line no-unused-vars
```
mleray, line 21: Maybe you could open a follow-up ticket to refactor this, since it seems it would be doable to just rename the props functions instead of disabling the linting 🤔 But I'm fine with this temporary solution 👍

assets/src/components/SidebarFields/ImageSidebarField.js
```diff
@@ -1,35 +1,36 @@
...
+
+export const ImageSidebarField = ({value, setValue, label}) => (
   <div className="components-base-control mb-3">
-    <label className="components-base-control__label mb-2 d-block">
+    <label className="components-base-control__label mb-2 d-block" htmlFor={value.id}>
       {label}
     </label>
     <MediaUploadCheck>
       <MediaUpload
+        id={value.id}
```
mleray, line 14: It looks like the MediaUpload component [doesn't accept `id` as a prop](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/media-upload/README.md), so I'm not sure how that works 🤔

mleray: Maybe the id should instead be set inside of the `render`

assets/src/components/PostTypeSelector/PostTypeSelector.js
Osong-Michael: Ok, noted.

assets/src/BlockEditorValidation.js
```diff
@@ -10,19 +10,22 @@ let canPublish = true;
 
 export const blockEditorValidation = () => {
   subscribe(() => {
-    const blocks = select( 'core/block-editor' ).getBlocks();
+    const blocks = select('core/block-editor').getBlocks();
 
     const currentMessages = [];
-    const invalidBlocks = blocks.reduce( (invalidBlocks, block ) => {
+    // eslint-disable-next-line no-shadow
+    const invalidBlocks = blocks.reduce((invalidBlocks, block) => {
```
GP-Dan-Tovbein, line 17: What if we find another name of `invalidBlocks` of the `accumulator`, which is the `shadow` var, and avoid adding `eslint-disable-next-line no-shadow`? I know that is  not part of the scope, but the idea of using `eslint` is also to improve the quality of the code.

What do you think?

assets/src/BlockEditorValidation.js
```diff
@@ -10,19 +10,22 @@ let canPublish = true;
...
       // editor crash while it's trying to access `block.name`.
       if (!block) {
+        // eslint-disable-next-line array-callback-return
         return;
       }
-      const validations = blockValidations[ block.name ] || {};
+      const validations = blockValidations[block.name] || {};
 
-      const results = Object.entries( validations ).reduce( (results, [attrName, validate]) => {
+      // eslint-disable-next-line no-shadow
```
GP-Dan-Tovbein, line 27: Same here!

assets/src/BlockEditorValidation.js
```diff
@@ -10,19 +10,22 @@ let canPublish = true;
...
     const currentMessages = [];
-    const invalidBlocks = blocks.reduce( (invalidBlocks, block ) => {
+    // eslint-disable-next-line no-shadow
+    const invalidBlocks = blocks.reduce((invalidBlocks, block) => {
       // Normally `blocks` contains a valid list of blocks, however it can happen that one of them is `null` in rare
       // cases. It happened to me once while running with WordPress 5.8.1 and undoing multiple edits. This made the
       // editor crash while it's trying to access `block.name`.
       if (!block) {
+        // eslint-disable-next-line array-callback-return
         return;
```
GP-Dan-Tovbein, line 23: Just thinking loud and not part of the scope.. In this case if you return the `accumulator` value will fix the `eslint` issue.

assets/src/BlockEditorValidation.js
```diff
@@ -49,60 +51,60 @@ export const blockEditorValidation = () => {
...
+  registerPlugin('pre-publish-checklist', {render: PrePublishCheckList});
   wp.hooks.addFilter(
     'blocks.registerBlockType',
     'planet4-plugin-gutenberg-blocks',
     registerAttributeValidations
   );
 };
 
-document.addEventListener('change', e=> {
+document.addEventListener('change', (e) => {
```
GP-Dan-Tovbein, line 68: Have we decided to use `e` instead of `evt`? Or we haven't decided yet?

assets/src/ImageBlockExtension.js
```diff
@@ -12,112 +12,113 @@ const targetBlocks = [
 
 const captionStyleOptions = [
   {
-  label: __( 'Medium' ),
```
GP-Dan-Tovbein, line 15: Correct me if I'm wrong, is it not missed the source here since you're translating here?

assets/src/blocks/Accordion/AccordionEditor.js
```diff
@@ -125,8 +124,9 @@ const renderEdit = ({ tabs }, setAttributes, updateTabAttribute) => {
     <InspectorControls>
       <PanelBody title={__('Settings', 'planet4-blocks-backend')}>
         {tabs.map((tab, index) => {
-          const { button } = tab;
+          const {button} = tab;
           if (!button) {
+            // eslint-disable-next-line array-callback-return
             return;
```
GP-Dan-Tovbein, line 130: Same as before, maybe if you return `null` you'll be avoided to add `array-callback-return`.

assets/src/blocks/Accordion/AccordionEditorScript.js
```diff
@@ -39,14 +38,15 @@ registerBlockType(BLOCK_NAME, {
...
-    'collapsible'
+    'collapsible',
   ],
   supports: {
     html: false, // Disable "Edit as HTMl" block option.
   },
   attributes,
   edit: AccordionEditor,
-  save: ({ attributes }) => {
+  // eslint-disable-next-line no-shadow
```
GP-Dan-Tovbein, line 48: As less as we add this skips, we'll have a cleaner code.

GP-Dan-Tovbein (commented):
You're doing a great job! I've just left some comments.

Still reviewing.. 🤣

assets/src/BlockEditorValidation.js
Osong-Michael: That is a great idea, I remember not fully understanding the logic here so I didn't want to change the var name and break things, but it is a good point.

assets/src/BlockEditorValidation.js
Osong-Michael: I am not 100% sure, but I think this is like a hack to prevent the editor crashing. It is not to really return a value.

assets/src/BlockEditorValidation.js
Osong-Michael: I could not find a rule to enforce this variable name, but we can decide as a team what to do

assets/src/ImageBlockExtension.js
Osong-Michael: I think just the indentation was changed on this line, the translation is still there.

assets/src/blocks/Accordion/AccordionEditor.js
Osong-Michael: Yea, same to as with the code above, this is just a workaround to prevent the editor from crashing. Returning `null` may actually cause that.

assets/src/blocks/Accordion/AccordionEditorScript.js
Osong-Michael: Ok, thanks for the feedback

assets/src/blocks/Columns/ColumnsTasks.js
```diff
@@ -43,19 +43,19 @@ export const ColumnsTasks = ({ isCampaign, columns, no_of_columns }) => (
...
                 <a
                   className={`btn btn-small btn-${isCampaign ? 'primary' : 'secondary'}`}
                   href={cta_link}
                   target={link_new_tab ? '_blank' : ''}
-                  data-ga-category='Columns Block'
-                  data-ga-action='Call to Action'
-                  data-ga-label={cta_link}
+                  data-ga-category="Columns Block"
+                  data-ga-action="Call to Action"
+                  data-ga-label={cta_link} rel="noreferrer"
```
mleray, line 58: From a quick research, it seems that we shouldn't add `rel="noreferrer"` to internal links (i.e. when `target` is not `_blank`) because it could mess with analytics... So maybe instead of adding it all the time we could use the same condition as for the target? Something like:
```
{...link_new_tab && { rel: 'noreferrer', target: '_blank' } }
```

assets/src/blocks/TakeActionBoxout/TakeActionBoxoutFrontend.js
```diff
@@ -13,15 +11,17 @@ export const TakeActionBoxoutFrontend = ({
...
   >
+    {/* eslint-disable-next-line jsx-a11y/anchor-has-content */}
     <a
       data-ga-category="Take Action Boxout"
       data-ga-action="Image"
       data-ga-label="n/a"
       className="cover-card-overlay"
       href={link}
-      {...newTab && link && { target: "_blank" }}
+      target={(newTab && link) ? '_blank' : ''}
```
mleray, line 23: Same comment for this file as earlier, we should set `rel="noreferrer"` only when `target` is `_blank`

assets/src/components/withCharacterCounter/withCharacterCounter.js
```diff
@@ -52,7 +52,7 @@ const withCharacterCounter = ( WrappedComponent ) => {
     }
 
     render() {
-      const { characterLimit, warningThreshold, onChange, ...passThroughProps } = this.props;
+      const {characterLimit, ...passThroughProps} = this.props;
```
mleray, line 55: I think that here again we need to keep excluding these 2 props from `passThroughProps`, since we override these functionalities

mleray (commented):
I see new submodule updates that should not be needed 🤔 Also the `package-lock` file still seems to have way too many changes 😬

assets/src/blocks/CarouselHeader/StaticCaption.js
```diff
@@ -1,22 +1,21 @@
...
-        <div className='col-xs-12 col-sm-8 col-md-4 action-button'>
+        <div className="col-xs-12 col-sm-8 col-md-4 action-button">
           <a
             href={slide.link_url}
-            target={slide.link_url_new_tab ? '_blank' : '_self'}
-            className='btn btn-primary btn-block'
-            data-ga-category='Carousel Header'
-            data-ga-action='Call to Action'
-            rel='noopener noreferrer'
+            {...slide.link_url_new_tab && {rel: 'noreferrer', target: '_blank'}}
```
mleray, line 15: We usually put these spreadable props (`{...}`) at the end, I don't know if it actually matters but I would do it anyways, at least for consistency

assets/src/blocks/Columns/Columns.js
```diff
@@ -25,62 +25,66 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample =
...
+
+                { (!isExample && cta_link) ?
                   <a
                     href={cta_link}
-                    data-ga-category='Columns Block'
+                    data-ga-category="Columns Block"
                     data-ga-action={columns_block_style === LAYOUT_ICONS ? 'Icon' : 'Image'}
                     data-ga-label={cta_link}
-                    { ...link_new_tab && { target: '_blank' } }
+                    {...link_new_tab && {target: '_blank'}}
```
mleray, line 53: We should add `rel: 'noreferrer'` prop here as well

assets/src/blocks/Columns/ColumnsTasks.js
```diff
@@ -31,9 +31,9 @@ export const ColumnsTasks = ({ isCampaign, columns, no_of_columns }) => (
                   {cta_link ?
                     <a
                       href={cta_link}
-                      target={link_new_tab ? '_blank' : ''}
-                      data-ga-category='Columns Block'
-                      data-ga-action='Title'
+                      {...link_new_tab && {rel: 'noreferrer', target: '_blank'}}
```
mleray, line 34: Same comment as earlier, I would move this spreadable prop to the end

assets/src/ImageBlockExtension.js
GP-Dan-Tovbein: yeah, I know but my point is if it is not also necessary to add the source here? 
Of course this is not related to your changes, but just thinking loud !

GP-Dan-Tovbein: And this is not blocking the approval!

assets/src/ImageBlockExtension.js
```diff
@@ -12,112 +12,113 @@ const targetBlocks = [
...
-          <BlockEdit { ...props } />
+          <BlockEdit {...props} />
           <InspectorControls>
             <PanelBody
-              title={ __( 'Planet4 Image Options' ) }
-              initialOpen={ true }
+              title={__('Planet4 Image Options')}
+              initialOpen={true}
             >
+              {/* eslint-disable-next-line jsx-a11y/label-has-for, jsx-a11y/label-has-associated-control */}
```
GP-Dan-Tovbein, line 95: Couldn't we add the `a11y` labels? Or that is not possible?

assets/src/blocks/Columns/ColumnsTasks.js
mleray: @Osong-Michael could you please make this change?

assets/src/blocks/Columns/Columns.js
```diff
@@ -25,62 +25,66 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample =
...
               {cta_link && !isCampaign ?
                 <a
                   href={cta_link}
-                  data-ga-category='Columns Block'
-                  data-ga-action='Title'
+                  data-ga-category="Columns Block"
+                  data-ga-action="Title"
                   data-ga-label={cta_link}
-                  { ...link_new_tab && { target: '_blank' } }
+                  {...link_new_tab && {target: '_blank'}}
```
mleray, line 68: We should add `rel: noreferrer` prop here too

assets/src/blocks/Columns/Columns.js
```diff
@@ -25,62 +25,66 @@ export const Columns = ({ columns, columns_block_style, isCampaign, isExample =
...
-                  'standalone-link'
-                }
-                data-ga-category='Columns Block'
-                data-ga-action='Call to Action'
+                  'standalone-link'}
+                data-ga-category="Columns Block"
+                data-ga-action="Call to Action"
                 data-ga-label={cta_link}
-                { ...link_new_tab && { target: '_blank' } }
+                {...link_new_tab && {target: '_blank'}}
```
mleray, line 87: We should add `rel: noreferrer` prop here too

PR #1015, opened 2023-02-27
PLANET-7094 Fix broken rounded images animation in Safari

mleray:
### Description

See [PLANET-7094](https://jira.greenpeace.org/browse/PLANET-7094)
It seems it's not recommended to use `all` in CSS transitions, since browsers might interpret it in different ways. For Safari specifically, with a transition on the `transform` property, you need to add an additional `transformZ(0)` to the parent for some reason 😬 

### Testing

In Safari you can test the broken version with [this page](https://www.greenpeace.org/finland/metsat/) and the fixed version with [this page](https://www-dev.greenpeace.org/test-leda/52387-2/) for example. Make sure it also works as expected in other browsers.

### DIFF
--- a/assets/src/variations/stretched-link/index.scss
+++ b/assets/src/variations/stretched-link/index.scss
@@ -4,9 +4,10 @@
   [class*="is-style-rounded-"].force-no-caption {
     overflow: hidden;
     border-radius: 50%;
+    transform: translateZ(0); // This is needed for Safari, otherwise the animation looks off.
 
     img {
-      transition: all .2s ease-in-out;
+      transition: transform .2s ease-in-out;
     }
   }
 


planet-4:

:watch: 2023.02.27 14:39:06

assets/src/variations/stretched-link/index.scss
```diff
@@ -4,9 +4,10 @@
   [class*="is-style-rounded-"].force-no-caption {
     overflow: hidden;
     border-radius: 50%;
```
GP-Dan-Tovbein, line 6: Not part of the scope, but do we also need this property if it's already defined [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/styles/blocks/core-overrides/Image.scss#L4)?

GP-Dan-Tovbein:
I've tested on Safari but I noticed it's missing the `pointer`. Do you also notice that?

mleray:
> I've tested on Safari but I noticed that it's missing the `pointer`. Do you also notice that?

Yes, it's normal, it's because there's no link in this example (so nothing happens if you click) but if you add one, the cursor will be `pointer` as expected 🙂

GP-Dan-Tovbein (commented):
I've left only [this](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1015/files#r1120076115) comment

assets/src/variations/stretched-link/index.scss
mleray: Yes, this one is for the parent (the element with `is-style-rounded` class) whereas the other one is for the child (either an `img` or a `figure` element)
We need both for the animation on hover, otherwise it looks off:
Screenshot 2023-02-28 at 15 20 14

GP-Dan-Tovbein (approved):
Code looks good to me!

PR #1029, opened 2023-04-05
PLANET-7016 Apply the new colors to the Table block

Osong-Michael:
Description: [PLANET-7016](https://jira.greenpeace.org/browse/PLANET-7016)

**Testing:**
You can use this [page](https://www-dev.greenpeace.org/test-umbriel/test-new-identity-colors/) on the Umbriel instance to test the new changes for the Table block. Switching between the new identity styles and old ones should be reflected on the table.

Ref PR: https://github.com/greenpeace/planet4-master-theme/pull/1970


planet-4:

:watch: 2023.04.05 09:25:50

assets/src/styles/blocks/core-overrides/Table.scss
```diff
@@ -4,44 +4,59 @@
...
+        background-color: var(--table-even-row-blue--background, $table-even-row-blue);
       }
 
       tfoot td {
-        background-color: $table-footer-blue;
+        background-color: var(--table-footer-blue--background, $table-footer-blue);
+      }
+    }
+
+    // GP-Green background
```
mleray, line 48: Could you please add this new style in the `new-identity/style` file instead of here (like for the [Spreadsheet block](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/new-identity/style.scss#L57-L74))? It should only be added for the new identity, and these CSS variables won't be defined if the new identity styles are disabled so I'm not sure what the behaviour will be like

Osong-Michael: 😅 my bad, I forgot to move the code after testing. Thanks

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -1,6 +1,5 @@
 import {Component, Fragment} from '@wordpress/element';
 import {ArrowIcon} from './ArrowIcon';
-import {toDeclarations} from '../toDeclarations';
```
mleray, line 3: Good catch! We probably can even remove that file?

Osong-Michael: Sure

assets/src/blocks/Spreadsheet/SpreadsheetEditorScript.js
```diff
@@ -22,9 +22,13 @@ const colors = [
 ];
 
 if (isNewIdentity) {
+  const newColorValues = ['#167f82', '#198700', '#45494c'];
```
mleray, line 25: Maybe here it would be simpler if we just override the `colors`? A bit longer to write maybe, but I think it would make it easier to understand

assets/src/blocks/Spreadsheet/SpreadsheetEditorScript.js
```diff
@@ -52,9 +56,9 @@ export class SpreadsheetEditor extends Component {
   renderEdit() {
     const {attributes, setAttributes} = this.props;
 
-    const toColorName = code => colors.find(color => color.color === code).name;
+    const toColorName = code => colors.find(color => color.color === code)?.name || 'grey';
 
-    const toColorCode = name => colors.find(color => color.name === name).color;
+    const toColorCode = name => colors.find(color => color.name === name)?.color || '#ececec';
```
mleray, line 61: If the new identity styles are enabled, it should be a different color code

mleray: Yeah I know, it's just that if for some reason we pass there a name that is not in the colors, the default color code will be `#ececec` which doesn't exist in the new identity color options... But that should never happen, I guess we can leave it as is

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
mleray: Can you please delete this file then?

assets/src/styles/blocks/Spreadsheet/SpreadsheetStyle.scss
```diff
@@ -38,16 +38,9 @@ table.spreadsheet-table {
...
-  th --block-spreadsheet-grey-header-- {
-    background: $table-header-grey;
-  }
-
   tr {
     td {
-      --block-spreadsheet-- {
-        color: $table-font-color;
-      }
+      color: var(--block-spreadsheet--color, $table-font-color);
```
mleray, line 43: Maybe here we could just use `var(--body--color)` and it will be directly updated, same thing for the Table block [here](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/layout/_tables.scss#L45) (I wonder why we have the styles for that block split between 2 files?!). That would also allow us to get rid of this [variable](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/base/_colors.scss#L57). What do you think?

assets/src/blocks/Spreadsheet/SpreadsheetEditorScript.js
```diff
@@ -15,17 +15,19 @@ import {
...
 if (isNewIdentity) {
-  colors.push({
-    name: 'dark-green',
-    color: '#1f4912'
-  });
+  colors = [
+    {name: 'blue', color: '#167f82'},
+    {name: 'green', color: '#198700'},
+    {name: 'grey', color: '#45494c'},
+    {name: 'dark-green',color: '#1f4912'},
```
mleray, line 29: Since this is a new color, we could rename it `gp-green` instead of `dark-green` to be kinda consistent with the Table block and the CSS variables

assets/src/styles/blocks/core-overrides/Table.scss
```diff
@@ -4,44 +4,44 @@
...
     }
 
     tr {
-      background-color: $table-even-row-grey;
+      background-color: var(--table-even-row-grey--background, $table-even-row-grey);
     }
 
     tfoot td {
-      background-color: $table-footer-grey;
+      background-color: var(--table-footer-grey--background, $table-footer-grey);
```
mleray, line 15: This is more of a design question than a dev question, so maybe something to check with Houssam, I find it weird that the footer color is the same as some rows if we have stripes, because they might end up next to each other:
Screenshot 2023-04-07 at 09 23 05

assets/src/blocks/Spreadsheet/SpreadsheetEditorScript.js
Osong-Michael: So, the `gp-green` color code is actually what replaced the green, the `dark-green` is the new color added and matches the color code for the Table block

mleray: well that shouldn't be the case, `green` should remain `green`, and for the Table block [the new color is named `gp-green`](https://github.com/greenpeace/planet4-master-theme/pull/1970/files#diff-01c9525afc2c87cfcb03124117c9d0ebb067029f6c955a607fa0fd4274aca556R614) so I think we should use the same here

mleray: The colors should correspond between the blocks... So `#1f4912` should correspond with `green` and we can use `gp-green` for `#198700`... It's also what is described in the ticket with `#198700` being the new color

assets/src/styles/blocks/Spreadsheet/SpreadsheetStyle.scss
Osong-Michael: If i am not mistaken, `var(--body--color)` exist only when the new identity styles is enabled right? So we may want to do something like `color: var(--body--color, $table-font-color)`, so we still maintain the old styles when we switch back.

assets/src/styles/blocks/core-overrides/Table.scss
Osong-Michael: Houssam suggested to make the text for the Footer bold.

Osong-Michael: Well ignore my last comment 😅.
After talking with him again, we figured out some color changes, pushing those now.

assets/src/blocks/Spreadsheet/SpreadsheetEditorScript.js
```diff
@@ -15,17 +15,19 @@ import {
...
 if (isNewIdentity) {
-  colors.push({
-    name: 'dark-green',
-    color: '#1f4912'
-  });
+  colors = [
+    {name: 'blue', color: '#167f82'},
+    {name: 'green', color: '#1f4912'},
+    {name: 'grey', color: '#45494c'},
+    {name: 'gp-green',color: '#198700'},
```
mleray, line 29: There's a missing space here between the comma and `color`

assets/src/styles/blocks/core-overrides/Table.scss
```diff
@@ -4,44 +4,45 @@
...
+      font-weight: bold;
```
mleray, line 16: If we change the footer color, do we still want the bold font weight too?

Osong-Michael: He approved it, so I think so.

PR #1035, opened 2023-04-12
PLANET-6995 Apply the new dark grey to title links

Osong-Michael:
**Description: [PLANET-6995](https://jira.greenpeace.org/browse/PLANET-6995)**

**Testing**

### DIFF
--- a/assets/src/styles/blocks/Articles.scss
+++ b/assets/src/styles/blocks/Articles.scss
@@ -74,7 +74,7 @@
 .article-list-item-headline {
   --block-articles--article--headline-- {
     font-weight: 500;
-    color: var(--body--color);
+    color: var(--link--color);
   }
 
   @include mobile-only {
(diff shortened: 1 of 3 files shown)


planet-4:

:watch: 2023.04.12 10:37:56

Osong-Michael:
> The problem with making this change is that it will break the current styles (when the new identity styles are disabled) 😕 So actually maybe it's best to keep `--body--color`

But `--link--color` is also declared in the css variables file, we are just overriding it, or am I wrong?

Osong-Michael:
> > But `--link--color` is also declared in the css variables file, we are just overriding it, or am I wrong?
> 
> Indeed, this change will work with the new identity, but with the current colors the `--link--color` variable is set to blue and not grey, so titles will become blue which is not what we want! If you disable the new identity styles on the [test instance](https://www-dev.greenpeace.org/test-neptune/) or on local you should see what I mean. So I think actually for this ticket no changes are required 🙂

Oh I see your point, in that case, this PR is not necessary then.

PR #1039, opened 2023-04-18
PLANET-7129: Change the import of the ServerSideRender function

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7129

## Description
After Wordpress 6.2, this function is exported as default in a separate module.

Taken from the [Wordpress's official document](https://make.wordpress.org/core/2023/03/06/add-new-prop-to-serversiderender-component/)
> The ServerSideRender component has moved to its own package/script in WordPress 5.3 and accessing it using wp.components.ServerSideRender has been deprecated since. Starting WordPress 6.2, you should be able to access it directly using wp.serverSideRender.

And here is their [changed code](https://github.com/WordPress/gutenberg/pull/46106/files#diff-77df04898762524cd09f00a391d4977694eeed141eba84bcde4ca8eb7473a40bL51-L56).

__Please make sure to test and merge this new functionality after the next PRs__
- https://github.com/greenpeace/planet4-base/pull/244
- https://github.com/greenpeace/planet4-develop/pull/11

### DIFF
--- a/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
+++ b/assets/src/blocks/SocialMediaCards/SocialMediaCards.js
@@ -8,10 +8,10 @@ import {
 import {
   TextControl,
   TextareaControl,
-  ServerSideRender,
   Button,
   Tooltip,
 } from '@wordpress/components';
+import ServerSideRender from '@wordpress/server-side-render';
 
 export class SocialMediaCards extends Component {
   renderEdit() {
--- a/assets/src/blocks/SubPages/SubPages.js
+++ b/assets/src/blocks/SubPages/SubPages.js
@@ -1,7 +1,5 @@
 import {Fragment} from '@wordpress/element';
-import {
-  ServerSideRender,
-} from '@wordpress/components';
+import ServerSideRender from '@wordpress/server-side-render';
 import {Preview} from '../../components/Preview';
 
 export const SubPages = props => {


mleray (commented):
I was testing this PR and I saw the following warning in the console: 

Screenshot 2023-04-21 at 10 53 24

Probably something that needs to be updated too?

mleray (commented):
Probably not due to these changes, but isn't the SocialMediaCards block broken? On local if I try to test it I see the following notice in the frontend:
```
Notice: Function rest_validate_value_from_schema was called incorrectly. The "type" schema keyword for cards[0][image_id] can only be one of the built-in types: array, object, string, number, integer, boolean, and null. Please see [Debugging in WordPress](https://wordpress.org/documentation/article/debugging-in-wordpress/) for more information. (This message was added in version 5.5.0.) in /var/www/html/wp-includes/functions.php on line 5865
```
I've tried on a test instance and then nothing shows in the frontend.

GP-Dan-Tovbein:
Hey @mleray, which pages are you testing? 

FYI, I'm currently running the new dev environment and with Wordpress 6.2 and I don't see any of your mentioned warnings. I'm curious if you running it the same as me 🤔

mleray:
> Hey @mleray, which pages are you testing?

For example on [this page](https://www-dev.greenpeace.org/test-rhea/social-media-cards/) nothing is rendered in the frontend!
On local I still see the following instead of the block:

Screenshot 2023-04-24 at 09 42 34

> FYI, I'm currently running the new dev environment and with Wordpress 6.2 and I don't see any of your mentioned warnings. I'm curious if you running it the same as me 🤔

I'm also running the new dev env with WordPress 6.2... Now I also see the following warning when adding a Carousel Header to a page:

Screenshot 2023-04-24 at 09 25 34

For the warning regarding the `MediaUpload` and `MediaUploadCheck` components, I see it when adding a Social Media Cards component which makes sense because indeed there [we import it](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/SocialMediaCards/SocialMediaCards.js#L3-L6) from `editor` instead of `block-editor`. Looking a bit more into this, it seems we are not very consistent with how we import it BTW: 
- [GalleryEditor](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/SocialMediaCards/SocialMediaCards.js#L3-L6) uses `@wordpress/blockEditor` 
- [EditableBackground](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/CarouselHeader/EditableBackground.js#L1) and some others use `@wordpress/block-editor`
- [HappyPointEditor](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/Happypoint/HappypointEditor.js#L9-L14) and some others use `wp.blockEditor`

All seem to work, but it would be nice to agree on one way to declare this and stick to it 😬 But of course this should probably be done in another PR

GP-Dan-Tovbein:
Hey @mleray, about the below comment 👇I've created [this PR](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1043) to follow up it. 

> All seem to work, but it would be nice to agree on one way to declare this and stick to it 😬 But of course this should probably be done in another PR

mleray (approved):
Code looks good to me 👍

PR #1041, opened 2023-04-24
PLANET-7063 Re-write the Editor Selector test in Playwright

Osong-Michael:
**Description: [PLANET-7063](https://jira.greenpeace.org/browse/PLANET-7063)**

Testing: 
run `npx playwright test tests/e2e/editor-selector.spec.js` in the plugin directory and the test will run.

Ref PR: https://github.com/greenpeace/planet4-master-theme/pull/2003

### DIFF
--- a/tests/e2e/editor-selector.spec.js
+++ b/tests/e2e/editor-selector.spec.js
@@ -1,0 +1,38 @@
+const {test, expect} = require('@playwright/test');
+import {login} from './tools/lib/login';
+
+test('Test Editor basic functionalities', async ({page, context}) => {
+  test.setTimeout(240 * 1000);
+  await page.goto('./');
+  await login(page, context);
+
+  // Creating and navigating to new post
+  await page.goto('./wp-admin/post-new.php');
+
+  // On dev instance, need to close modal so test can continue
+  const closeButton = page.locator('.components-modal__header button');
+  if (await closeButton.isVisible()) {
+    await closeButton.click();
+  }
+
+  await page.locator('.editor-post-title__input').click();
+  await page.locator('h1.editor-post-title').fill('Test Post');
+  await page.locator('.block-editor-block-list__layout').click();
+  await page.locator('p.is-selected.wp-block-paragraph').fill('This is a test Post.');
+  await page.keyboard.press('Enter');
+  await page.locator('p.is-selected.wp-block-paragraph').fill('/youtube');
+  await page.keyboard.press('Enter');
+  await page.locator('[aria-label="YouTube URL"]').fill('https://youtu.be/3gPvDDHU41E');
+  await page.getByRole('button', {name: 'Embed'}).click();
+  await page.getByRole('button', {name: 'Publish', exact: true}).click();
+  await page.getByRole('region', {name: 'Editor publish'}).getByRole('button', {name: 'Publish', exact: true}).click();
+  await page.getByRole('link', {name: 'View Post', exact: true}).first().click();
+
+  // Asserting new Post contains video, paragraph and title
+  const h1 = await page.innerHTML('h1.page-header-title');
+  const paragraph = await page.innerHTML('.post-details p');
+  const video = page.locator('figure.is-provider-youtube');
+  expect(h1).toBe('Test Post');
+  expect(paragraph).toBe('This is a test Post.');
+  expect(video).toBeVisible();
+});


planet-4:

:new_moon: [deimos](https://www-dev.greenpeace.org/test-deimos) | [admin](https://www-dev.greenpeace.org/test-deimos/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-deimos/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-deimos) | [composer-local.json](https://github.com/greenpeace/planet4-test-deimos/blob/main/composer-local.json)

:watch: 2023.04.24 09:47:22

mleray (changes requested):
I think that this ticket is about testing the editor functionalities, so in this case we shouldn't use the `rest` function to create the post 🤔

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
+const {test, expect} = require('@playwright/test');
+import {login} from './tools/lib/login';
+
+test('Test the creation of a new Post with an embedded video and paragraph and check if they are present when published', async ({page, context}) => {
```
mleray, line 4: This is a really long test name, could we shorten it? Maybe `Test the creation of a new Post` is enough 🤔

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  test.setTimeout(240 * 1000);
+  await page.goto('./');
+  await login(page, context);
+
+  // Creating and navigating to new post
+  await page.goto('./wp-admin/post-new.php');
+
+  // On dev instance, need to close modal so test can continue
+  const closeButton = page.locator('.components-modal__header button');
+  if (await closeButton.isVisible()) {
```
mleray, line 14: This is just personal preference, but it seems a bit strange to me to have an `await` statement inside an `if`, maybe we could write this instead:
```
const isCloseButtonVisible = await page.locator('.components-modal__header button').isVisible();
if (isCloseButtonVisible) { ... }
```
What do you think?

Osong-Michael: Actually reading the documentation and seeing other code examples, it is better to have the await in the if statement, seems to validate better. I had it like you suggested but it kept failing.

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
...
+test('Test the creation of a new Post', async ({page, context}) => {
+  test.setTimeout(240 * 1000);
+  await page.goto('./');
+  await login(page, context);
+
+  // Creating and navigating to new post
+  await page.goto('./wp-admin/post-new.php');
+
+  // On dev instance, need to close modal so test can continue
+  const closeButton = page.locator('.components-modal__header button');
```
mleray, line 13: Does this mean that the page creation modal no longer shows on local? Because it should 🤔 Could this be due to the WordPress version maybe?

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  await page.locator('.editor-post-title__input').click();
+  await page.locator('h1.editor-post-title').fill('Test Post');
+  await page.locator('.block-editor-block-list__layout').click();
+  await page.locator('p.is-selected.wp-block-paragraph').fill('This is a test Post.');
+  await page.keyboard.press('Enter');
+  await page.locator('p.is-selected.wp-block-paragraph').fill('/youtube');
+  await page.keyboard.press('Enter');
+  await page.locator('[aria-label="YouTube URL"]').fill('https://youtu.be/3gPvDDHU41E');
+  await page.getByRole('button', {name: 'Embed'}).click();
+  await page.getByRole('button', {name: 'Publish', exact: true}).click();
```
mleray, line 27: Is there a way to select the buttons without using the name? 🤔 Because in test instances sometimes we use a different language in the backend, in which case these might fail

tests/e2e/editor-selector.spec.js
Osong-Michael: I have no idea, but on my local it does not show up, also the flow is a little different, the work around I had earlier made things slow so I refactored it.

Osong-Michael: If you use the VSCode extension and run the test in debug mode, it opens a webpage and shows you the tests running visually, that way you can see the difference.

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
...
+  await page.getByRole('region', {name: 'Editor publish'}).getByRole('button', {name: 'Publish', exact: true}).click();
+  await page.getByRole('link', {name: 'View Post', exact: true}).first().click();
```
Osong-Michael, line 29: @mleray, this is one of the reasons I use the names of the elements to locate them. In this case to minimise the amount of code written and make sure the tests run both on _local_ and _dev-instances_, I use this `selector` to catch any instances of the `link` _"View Post"_, on _local_ it appears once but on the _dev-instance_ it appears twice, so just the first instance is clicked. Also I realised using these `selectors` with names, seem to work faster when running tests.

mleray: I guess it's okay to keep as is then. Just something to keep in mind for certain test instances where it might cause issues due to a different language!

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
...
+
+  // Asserting new Post contains video, paragraph and title
+  const h1 = await page.innerHTML('h1.page-header-title');
+  const paragraph = await page.innerHTML('.post-details p');
+  const video = page.locator('figure.is-provider-youtube');
+  expect(h1).toBe('Test Post');
+  expect(paragraph).toBe('This is a test Post.');
+  expect(video).toBeVisible();
```
mleray, line 37: Maybe we should also check that the src is the right link?

mleray (commented):
Could you open another PR to remove the [corresponding acceptance test](https://github.com/greenpeace/planet4-master-theme/blob/main/tests/acceptance/editor.feature)?

tests/e2e/editor-selector.spec.js
```diff
@@ -0,0 +1,38 @@
+const {test, expect} = require('@playwright/test');
+import {login} from './tools/lib/login';
+
+test('Test the creation of a new Post', async ({page, context}) => {
```
mleray, line 4: I see actually that in the initial test the [description](https://github.com/greenpeace/planet4-master-theme/blob/main/tests/acceptance/editor.feature#LL1C1-L1C1) is `editor basics` so maybe we could rename this one to `test editor basic functionalities` or something like this?

tests/e2e/editor-selector.spec.js
Osong-Michael: The description did say we should test if the video is visible, as for testing the `src`, I am not sure how to do that because this is what I get when I inspect the page.
Screenshot 2023-05-02 at 22 04 38

comzeradd:
Why is this test on the plugin and not in the theme?

Osong-Michael:
> Why is this test on the plugin and not on the theme?

This was my decision seeing we are testing the editor's functionality and the Modal test was in the plugin repo.

PR #1045, opened 2023-05-03
Update Split Two Columns Block

GP-Dan-Tovbein:
## Description
This PR comes from [this comment](https://github.com/greenpeace/planet4-master-theme/pull/1996#discussion_r1183775081).
- https://github.com/greenpeace/planet4-master-theme/pull/1996
 
No ticket needed.


mleray (commented):
The whole point of my suggestion was to simplify the code, this doesn't really do that 🤔 In the current code we don't need to specify the `test-decoration` at all here since it follows normal link styles

PR #1042, opened 2023-04-24
PLANET-7153: Follow up Wordpress 6.2 deprecations

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-7153

## Description
Since we're about to upgrade to [WordPress 6.2 “Dolphy”](https://wordpress.org/news/2023/03/dolphy/) we would need to make sure our code is up to date with the new requirements. Some of them has been deprecated since [WordPress 6.1 “Misha”](https://wordpress.org/news/2022/11/misha/).

**The current PR should be merged after**
- https://github.com/greenpeace/planet4-base/pull/244

## Testing
Make sure to run `Wordpress 6.2` on your local.

PR #1050, opened 2023-05-08
PLANET-6958: Remove deprecated multiline prop

GP-Dan-Tovbein:
## Description
[PLANET-6958](https://jira.greenpeace.org/browse/PLANET-6958).

### Related PRs
- https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1042

### External Resources
- [Migrate to InnerBlocks](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/nested-blocks-inner-blocks/)

### DIFF
--- a/assets/src/blocks/Accordion/AccordionEditor.js
+++ b/assets/src/blocks/Accordion/AccordionEditor.js
@@ -43,7 +43,6 @@ const renderView = ({title, description, tabs, className}, setAttributes, isSele
           value={title}
           onChange={toAttribute('title')}
           withoutInteractiveFormatting
-          multiline="false"
           allowedFormats={[]}
         />
       </header>
@@ -65,7 +64,6 @@ const renderView = ({title, description, tabs, className}, setAttributes, isSele
             value={tab.headline}
             onChange={updateTabAttribute('headline', index)}
             withoutInteractiveFormatting
-            multiline="false"
             allowedFormats={[]}
           />
           <div className={`panel ${isSelected ? '' : 'panel-hidden'}`}>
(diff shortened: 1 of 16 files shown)


comzeradd:
So following up @mleray's comment from the other PR, now in these fields editors will be able to add multiple lines with this change. But looking at where this is currently being used, I see it's just for headers, where it should be ok to remove because I don't expect anyone to start adding multiple lines of `h2` text there.

Have you seen any other use case for this?

GP-Dan-Tovbein:
> So following up @mleray's comment from the other PR, now in these fields editors will be able to add multiple lines with this change. But looking at where this is currently being used, I see it's just for headers, where it should be ok to remove because I don't expect anyone to start adding multiple lines of `h2` text there.
> 
> Have you seen any other use case for this?

Yes [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/ENForm/ENFormInPlaceEdit.js#L277), which is the only place, in our code, that is using the `multiline` prop.

mleray:
> Yes [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/ENForm/ENFormInPlaceEdit.js#L277), which is the only place, in our code, that is using the `multiline` prop.

We should also remove these ones right? Not just the places where we set `multiline` to `false`.

mleray:
What about [this one](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/ENForm/ENFormInPlaceEdit.js#L277)?

mleray (approved):
👍

GP-Dan-Tovbein:
Unfortunately the `multiline` prop doesn't work even you set to `true` or `false`. Then, in favor of that, we'll proceed to remove it at all and proceed to the refactor of some blocks (as `<InnerBlocks/>`).
Here is an example of the [Accordion refactoring ticket](https://jira.greenpeace.org/browse/PLANET-7272).

PR #1064, opened 2023-06-07
PLANET-7166 Re-write Covers Take Action test in Playwright

sagarsdeshmukh:
Ref.: https://jira.greenpeace.org/browse/PLANET-7166

- Added test for Covers Take action block
- Removed old codeception test

**Testing:**
Run  `npx playwright test tests/e2e/covers-take-action.spec.js ` on branch locally.


tests/e2e/tools/lib/covers.js
```diff
@@ -0,0 +1,63 @@
+const {expect} = require('@playwright/test');
+
+const TAG_NAMES = ['Climate', 'Oceans','renewables', 'Consumption'];
+const PAGE_NAMES = ['Vestibulum leo libero', 'Consectetur adipiscing elit', 'Vestibulum placerat'];
+
+async function addCoversBlock(page, style) {
+  // Add Covers block.
+  await page.locator('.block-editor-block-list__layout').click();
+  await page.locator('p.is-selected.wp-block-paragraph').fill('/planet-4-covers');
```
lithrel, line 9: can you use [`type`](https://playwright.dev/docs/api/class-locator#locator-type) instead of `fill` in that case ? it seems more reliable to make the popover appear

lithrel (approved):
:+1:

PR #1066, opened 2023-06-15
PLANET-7083 Convert Homepage pattern layout to Block Templates

Osong-Michael:
**Description:** https://jira.greenpeace.org/browse/PLANET-7083

**Testing:** 

Open the editor on the assigned instance or on your local on this branch and the layout for the Homepage pattern should still look the same.

### DIFF
--- a/assets/src/block-templates/homepage/template.js
+++ b/assets/src/block-templates/homepage/template.js
@@ -1,0 +1,34 @@
+import gravityFormWithText from '../templates/gravity-form-with-text';
+const {__} = wp.i18n;
+const isNewIdentity = window.p4ge_vars.planet4_options.new_identity_styles ?? false;
+const backgroundColor = isNewIdentity ? 'beige-100' : 'grey-05';
+
+const template = () => (
+  [
+    ['core/group', {}, [
+      ['planet4-blocks/carousel-header'],
+      ['planet4-block-templates/issues', {
+        titlePlaceholder: __('The issues we work on', 'planet4-blocks'),
+      }],
+      ['core/spacer', {height: '88px'}],
+      ['planet4-blocks/articles', {
+        article_heading: __('Read our Stories', 'planet4-blocks'),
+      }],
+      ['core/spacer', {height: '56px'}],
+      ['planet4-block-templates/side-image-with-text-and-cta', {
+        title: __('Get to know us', 'planet4-blocks'),
+      }],
+      ['core/spacer', {height: '30px'}],
+      ['planet4-block-templates/side-image-with-text-and-cta', {
+        title: __('We win campaigns', 'planet4-blocks'),
+        mediaPosition: 'right',
+      }],
+      ['core/spacer', {height: '56px'}],
+      ['planet4-blocks/covers'],
+      ['core/spacer', {height: '72px'}],
+      gravityFormWithText({backgroundColor}),
+    ]],
+  ]
+);
+
+export default template;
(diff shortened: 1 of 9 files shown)


planet-4:

:watch: 2023.06.15 08:53:53

assets/src/block-templates/homepage/template.js
```diff
@@ -0,0 +1,31 @@
...
+      }],
+      ['core/spacer', {height: '30px'}],
+      ['planet4-block-templates/side-image-with-text-and-cta', {
+        title: __('We win campaigns', 'planet4-blocks'),
+        mediaPosition: 'right',
+      }],
+      ['core/spacer', {height: '56px'}],
+      ['planet4-blocks/covers'],
+      ['core/spacer', {height: '72px'}],
+      ['gravityforms/form'],
```
mleray, line 26: According to the previous version of the template, this shouldn't be just a Gravity Forms block, but [it should have some text on the side](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1066/files#diff-25f0e4e3db0b77616a6f7b7662e425dc8343525cac2a6b8a478461a2ef08b376L85)
Unless you agreed on this change with the design team?

assets/src/block-templates/template-list.js
```diff
@@ -5,6 +5,7 @@ import * as realityCheck from './reality-check';
 import * as issues from './issues';
 import * as pageHeader from './page-header';
 import * as highlightedCta from './highlighted-cta';
+import * as homepage from './homepage';
```
mleray, line 8: Just an idea but maybe we could rename the `homepage` folder to `homepage-layout`, to differentiate between the layouts and other templates? 🤔

assets/src/block-templates/homepage/template.js
```diff
@@ -0,0 +1,34 @@
...
+const {__} = wp.i18n;
+const isNewIdentity = window.p4ge_vars.planet4_options.new_identity_styles ?? false;
+const backgroundColor = isNewIdentity ? 'beige-100' : 'grey-05';
+
+const template = () => (
+  [
+    ['core/group', {}, [
+      ['planet4-blocks/carousel-header'],
+      ['planet4-block-templates/issues', {
+        titlePlaceholder: __('The issues we work on', 'planet4-blocks-backend'),
```
mleray, line 11: I see some inconsistencies in the translation domains, I think `planet4-blocks` makes more sense since these might be visible in the frontend too if editors don't change them? 🤔

mleray: Although some are obvious placeholders like `Enter title` so I'm not sure

comzeradd: All the strings that appear in the frontend should be just `planet4-blocks`. But I think input placeholders don't.

Osong-Michael: Yea I think so too, because in the [php](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/classes/patterns/class-homepage.php#L45) file it was `planet4-blocks-backend`

mleray: Yeah, this case is a bit tricky because technically it's not really a placeholder (not as generic as `Enter title` for example), editors might not change it and in that case it will appear in the frontend, that's why I wasn't sure

mleray (approved):

PR #1086, opened 2023-07-13
PLANET-7081 Convert Get Informed pattern layout to block template

mleray:
### Description

See [PLANET-7081](https://jira.greenpeace.org/browse/PLANET-7081)
This is easier to write and maintain

### Testing

Either on local or on the test instance, you should still be able to add the `Get Informed` pattern layout, and it should look the same as before.

### DIFF
--- a/assets/src/block-templates/issues/template.js
+++ b/assets/src/block-templates/issues/template.js
@@ -45,7 +45,7 @@ const item = ['core/group', {
 
 const template = ({
   backgroundColor = isNewIdentity ? 'beige-100' : 'grey-05',
-  titlePlaceholder = __('Enter title', 'planet4-blocks-backend'),
+  title = '',
 }) => ([
   ['core/group', {
     align: 'full',
@@ -66,7 +66,8 @@ const template = ({
     }, [
       ['core/heading', {
         level: 2,
-        placeholder: titlePlaceholder,
+        content: title,
+        placeholder: __('Enter title', 'planet4-blocks-backend'),
         style: {
           spacing: {
             margin: {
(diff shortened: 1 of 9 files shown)


assets/src/block-templates/issues/template.js
```diff
@@ -66,7 +66,7 @@ const template = ({
     }, [
       ['core/heading', {
         level: 2,
-        placeholder: titlePlaceholder,
+        content: titlePlaceholder,
```
mleray, line 69: This is to be consistent with other templates, also the idea is that this "placeholder" could potentially be used as the final title

assets/src/block-templates/issues/template.js
```diff
@@ -45,7 +45,7 @@ const item = ['core/group', {
 
 const template = ({
   backgroundColor = isNewIdentity ? 'beige-100' : 'grey-05',
-  titlePlaceholder = __('Enter title', 'planet4-blocks-backend'),
+  title = '',
```
mleray, line 48: This is to be consistent with other templates, the idea is that this could potentially be used as the final title so it's not a placeholder

planet-4:

:new_moon: [telesto](https://www-dev.greenpeace.org/test-telesto) | [admin](https://www-dev.greenpeace.org/test-telesto/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-telesto/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-telesto) | [composer-local.json](https://github.com/greenpeace/planet4-test-telesto/blob/main/composer-local.json)

:watch: 2023.07.13 09:11:24

PR #1087, opened 2023-07-14
PLANET-7081 Convert Get Informed pattern layout to block template

mleray:
### Description

See [PLANET-7081](https://jira.greenpeace.org/browse/PLANET-7081)
This is easier to write and maintain

### Testing

Either on local or on the test instance, you should still be able to add the `Get Informed` pattern layout, and it should look the same as before.


sagarsdeshmukh (approved):
PR looks good to me 👍

PR #1091, opened 2023-07-19
Fix act_page notices

lithrel:
Fix notices when act_page is not defined


mleray (approved):
Tested on local, no more warnings 👍

PR #1102, opened 2023-08-21
Refactor Spreadsheet block to use hydration

mleray:
### Description

See [PLANET-6917](https://jira.greenpeace.org/browse/PLANET-6917)
This is to stop using the deprecated `frontendRendered` function

### Testing

You can make sure the block still works as expected either on local or on the iocaste test instance (for example I've created [this page](https://www-dev.greenpeace.org/test-iocaste/the-title/) for tests)


planet-4:

:new_moon: [iocaste](https://www-dev.greenpeace.org/test-iocaste) | [admin](https://www-dev.greenpeace.org/test-iocaste/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-iocaste/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-iocaste) | [composer-local.json](https://github.com/greenpeace/planet4-test-iocaste/blob/main/composer-local.json)

:watch: 2023.08.21 10:42:55

assets/src/blocks/Spreadsheet/SpreadsheetBlock.js
```diff
@@ -1,26 +1,46 @@
-import {SpreadsheetEditor} from './SpreadsheetEditorScript';
+import {SpreadsheetEditor} from './SpreadsheetEditor';
 import {CSS_VARIABLES_ATTRIBUTE} from '../CssVariablesAttribute';
 import {frontendRendered} from '../frontendRendered';
+import {SpreadsheetFrontend} from './SpreadsheetFrontend';
+import {renderToString} from 'react-dom/server';
 
 const BLOCK_NAME = 'planet4-blocks/spreadsheet';
 
+const attributes = {
```
GP-Dan-Tovbein, line 9: What do you think if we define the next attribute by default?
```javascript
{
      url: {
    type: 'string',
    default: '',
  },
}
```
and then override it.

We'll avoid to repeat the `url` field in both versions. For instance,
```javascript
{
   ...attributes,
    color: {
         type: 'string',
         default: 'grey',
     },
}
```
OR
```javascript
```javascript
{
   ...attributes,
    css_variables: CSS_VARIABLES_ATTRIBUTE,
}
```
```

File: assets/src/blocks/Spreadsheet/SpreadsheetEditor.js [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+            onChange={newUrl => {
+              setUrl(newUrl);
+              debounceUrl(newUrl);
+            }}
+          />
+          <div className="sidebar-blocks-help">
+            <ul>
+              <li>
+                {/* eslint-disable-next-line @wordpress/i18n-no-collapsible-whitespace, no-restricted-syntax */}
+                { __(`From Your Google Spreadsheet Table choose File -> Publish on web.
```
GP-Dan-Tovbein, line 68: Can we use single quote here?

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
```diff
@@ -0,0 +1,117 @@
...
+  {name: 'green', color: '#d0fac9'},
+  {name: 'grey', color: '#ececec'},
+];
+
+if (isNewIdentity) {
+  colors = [
+    {name: 'blue', color: '#167f82'},
+    {name: 'green', color: '#1f4912'},
+    {name: 'grey', color: '#45494c'},
+    {name: 'gp-green',color: '#198700'},
```
GP-Dan-Tovbein, line 23: Missing empty space after `'gp-green',`

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
```diff
@@ -0,0 +1,117 @@
+import {useState} from '@wordpress/element';
+import {InspectorControls} from '@wordpress/block-editor';
+import ColorPaletteControl from '../../components/ColorPaletteControl/ColorPaletteControl';
+import {SpreadsheetFrontend} from './SpreadsheetFrontend';
+import {debounce} from 'lodash';
+import {TextControl, PanelBody} from '@wordpress/components';
```
GP-Dan-Tovbein, line 6: Could we move all the `@wordpress` deps at the top of the file, then other dependency, and finally the modules?

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -16,88 +16,75 @@ const placeholderData = {
...
+  const [loading, setLoading] = useState(false);
+  const [spreadsheetData, setSpreadsheetData] = useState(null);
+  const [searchText, setSearchText] = useState('');
+  const [sortDirection, setSortDirection] = useState('asc');
+  const [sortColumnIndex, setSortColumnIndex] = useState(null);
+
+  const onHeaderClick = index => {
+    if (index === sortColumnIndex) {
+      const newDirection = sortDirection === 'asc' ? 'desc' : 'asc';
+      setSortDirection(newDirection);
```
GP-Dan-Tovbein, line 34: I think it's not necessary to define a `const` since it's only used once.

Maybe `setSortDirection(sortDirection === 'asc' ? 'desc' : 'asc');` is better?

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -16,88 +16,75 @@ const placeholderData = {
...
+    const matches = urlParam.match(googleSheetsPattern);
     if (matches !== null) {
       return matches[1];
     }
     return false;
-  }
+  };
+
+  useEffect(() => {
+    const fetchSheetData = async () => {
```
GP-Dan-Tovbein, line 51: We can use [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) instead
```javascript
(async () => { ... })()
```

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -145,74 +132,69 @@ export class SpreadsheetFrontend extends Component {
             row.map((cell, cellIndex) => (
               <td key={cellIndex}>
                 {
-                  this.state.searchText.length > 0 ?
+                  searchText.length > 0 ?
```
GP-Dan-Tovbein, line 135: In this case, just evaluating `searchText.length` should works fine too.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -145,74 +132,69 @@ export class SpreadsheetFrontend extends Component {
...
-        </section>
-      </Fragment>
-    );
-  }
-}
+              </tr>
+            </thead>
+            <tbody>
+              {
+                loading === true ?
```
GP-Dan-Tovbein, line 185: I don't see necessary to evaluate if the value is `=== true`.

assets/src/blocks/Spreadsheet/SpreadsheetBlock.js
mleray: If we do it this way then this version: 
```
{ ...attributes, color: { type: 'string', default: 'grey', }, }
```
will be duplicated (once for the actual attributes, and once in the `deprecated` object) so I'm not sure it's better than the current implementation?

assets/src/blocks/Spreadsheet/SpreadsheetEditor.js
mleray: Yes we could, but is it really important? If it is then we should open a ticket to change that in other blocks too, because we don't respect that everywhere.

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -145,74 +132,69 @@ export class SpreadsheetFrontend extends Component {
...
-                  }
-                </tr>
-              </thead>
-              <tbody>
+  };
+
+  const headers = spreadsheetData ? spreadsheetData.header : placeholderData.header;
+
+  return (
+    <>
```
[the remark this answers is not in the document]
mleray: Nope 😁

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
mleray: I just kept the previous implementation, but I can change it

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
mleray: Updated!

assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js
```diff
@@ -145,74 +129,66 @@ export class SpreadsheetFrontend extends Component {
...
+            <tr>
+              {
+                headers.map((cell, index) => (
+                  <th
+                    className={(
+                      index === sortColumnIndex ?
+                        `spreadsheet-sorted-by sort-${sortDirection}` :
+                        ''
+                    )}
+                    onClick={() => onHeaderClick(index)}
```
GP-Dan-Tovbein, line 166: What do you think about the below suggestion?

Pass the `index` value through the dataset
```
...
data-index={index}
onClick={onHeaderClick}
...
```
Then from [here](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/adac09d072079130e1bb7c68c5c12331ddbe2a92/assets/src/blocks/Spreadsheet/SpreadsheetFrontend.js#L31-L38) we would do something like this:
```js
  const onHeaderClick = (evt) => {
   const index = evt.currentTarget.dataset.index;
    if (index === sortColumnIndex) {
      setSortDirection(sortDirection === 'asc' ? 'desc' : 'asc');
    } else {
      setSortColumnIndex(index);
      setSortDirection('asc');
    }
  };
```

What will be gain? To avoid create multiple `anonymus` function through the render method and use only one.

GP-Dan-Tovbein (approved):
Code looks good to me! Good job 👏 
I've only left [this suggestion](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1102/files#r1365445772) but it's up to you to apply it! It doesn't block the merge.

PR #1105, opened 2023-08-25
Refactor GuestBook block

mleray:
### Description

See [PLANET-6917](https://jira.greenpeace.org/browse/PLANET-6917)
This is to stop using the deprecated `frontendRendered` function

### DIFF
--- a/assets/src/blocks/GuestBook/GuestBookEditorScript.js
+++ b/assets/src/blocks/GuestBook/GuestBookEditorScript.js
@@ -1,0 +1,3 @@
+import {registerGuestBookBlock} from './GuestBookBlock';
+
+registerGuestBookBlock();
--- a/assets/src/blocks/GuestBook/GuestBookFrontend.js
+++ b/assets/src/blocks/GuestBook/GuestBookFrontend.js
@@ -9,7 +9,7 @@ export const GuestBookFrontend = () => {
 
   return (
     <p>
-      <iframe src={buildURL()} width="100%" height={700} frameBorder="0" title="GuestBook" />
+      <iframe src={buildURL()} width="100%" height={700} title="GuestBook" />
     </p>
   );
 };
--- a/assets/src/blocks/GuestBook/GuestBookScript.js
+++ b/assets/src/blocks/GuestBook/GuestBookScript.js
@@ -1,0 +1,11 @@
+import {createRoot} from 'react-dom/client';
+import {BLOCK_NAME} from './GuestBookBlock';
+import {GuestBookFrontend} from './GuestBookFrontend';
+
+// Fallback for non migrated content. Remove after migration.
+document.querySelectorAll(`[data-render="${BLOCK_NAME}"]`).forEach(
+  blockNode => {
+    const rootElement = createRoot(blockNode);
+    rootElement.render(<GuestBookFrontend />);
+  }
+);
--- a/classes/blocks/class-guestbook.php
+++ b/classes/blocks/class-guestbook.php
@@ -29,6 +29,9 @@ class GuestBook extends Base_Block {
 		register_block_type(
 			self::get_full_block_name(),
 		);
+
+		add_action( 'enqueue_block_editor_assets', [ self::class, 'enqueue_editor_assets' ] );
+		add_action( 'wp_enqueue_scripts', [ self::class, 'enqueue_frontend_assets' ] );
 	}
 
 	/**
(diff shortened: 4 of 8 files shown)


assets/src/blocks/GuestBook/GuestBookFrontend.js
```diff
@@ -9,7 +9,7 @@ export const GuestBookFrontend = () => {
 
   return (
     <p>
-      <iframe src={buildURL()} width="100%" height={700} frameBorder="0" title="GuestBook" />
```
mleray, line 12: This was marked as deprecated

planet-4:

:watch: 2023.08.25 08:43:28

classes/blocks/class-guestbook.php
```diff
@@ -29,6 +29,9 @@ public function __construct() {
 		register_block_type(
 			self::get_full_block_name(),
 		);
+
+		add_action( 'enqueue_block_editor_assets', [ self::class, 'enqueue_editor_assets' ] );
```
GP-Dan-Tovbein, line 33: What if you directly invoke `self::enqueue_editor_assets` instead of passing the class and method name?

assets/src/blocks/GuestBook/GuestBookEditorScript.js
```diff
@@ -0,0 +1,25 @@
+import {frontendRendered} from '../frontendRendered';
```
GP-Dan-Tovbein, line 1: I'm not sure if I understand this 🤔 
Why do we move the block logic to an script file? Shouldn't respect the same structure as the other blocks? I mean, the script to one side, and the block logic to another side

assets/src/blocks/GuestBook/GuestBookScript.js
```diff
@@ -0,0 +1,10 @@
+import {createRoot} from 'react-dom/client';
+import {GuestBookFrontend} from './GuestBookFrontend';
+
+// Fallback for non migrated content. Remove after migration.
+document.querySelectorAll('[data-render="planet4-blocks/guestbook"]').forEach(
```
GP-Dan-Tovbein, line 5: Don't we need to `hydrate` it afterwards?

mleray: This block doesn't need hydration to work properly

assets/src/blocks/GuestBook/GuestBookEditorScript.js
mleray: A lot of our blocks have this same structure 🤔 One script for the frontend and one script for the registration

classes/blocks/class-guestbook.php
mleray: Then you get an error, something like `"enqueue_editor_assets" is not a constant`, seems something like `function () { self::enqueue_editor_assets(); }` would work too but then it's longer so I think the current notation is better?

assets/src/blocks/GuestBook/GuestBookEditorScript.js
GP-Dan-Tovbein: Yeah I know, but my point is, shouldn't the registration be into the `GuestBookBlock.js` instead of `GuestBookScript.js`file?

mleray: It's in `GuestBookEditorScript.js`, I did that to be consistent with other blocks from our webpack config

mleray: Another possibility would have been to declare the registration as a function in `GuestBookBlock.js` and then call it in `GuestBookEditorScript.js` (similar to the [CarouselHeader block](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/blocks/CarouselHeader/CarouselHeaderEditorScript.js)) but I personally don't really see the point in creating 2 separate files for this

GP-Dan-Tovbein:
I've tested locally and I got this:

From the `main` branch I created a new page and insterted the `GuestBookBlock`

Screenshot 2023-09-01 at 08 45 51

Then I moved to the `refactor-guestbook-block`

Screenshot 2023-09-01 at 08 46 07

assets/src/blocks/GuestBook/GuestBookEditorScript.js
mleray: Okay, following our conversation on Slack I understand that this is easier to review so I've updated the code

GP-Dan-Tovbein (approved):
Code looks good to me as the first step for refactoring! 👏 
Good job 🥇

PR #1108, opened 2023-08-29
PLANET-6795: Fix top spacing of Media and Text

GP-Dan-Tovbein:
Ref: https://jira.greenpeace.org/browse/PLANET-6795

# Description
We should avoid to leave an empty space through within the `.media-and-text` block, only if it is rendered as the `first type of` into the `.page-content`.

## Testing
I've imported the same reported page for testing.


assets/src/styles/blocks/core-overrides/BlockSpacing.scss
```diff
@@ -23,6 +22,20 @@
...
+.page-content > .wp-block-media-text {
+  margin-top: $sp-1;
+  margin-bottom: $sp-2;
+
+  @include large-and-up {
+    margin-top: $sp-2;
+    margin-bottom: $sp-4;
+  }
+
+  &:first-of-type {
```
mleray, line 34: Shouldn't it be `first-child` instead of `first-of-type`?

GP-Dan-Tovbein: Yep! Makes more sense !

assets/src/styles/blocks/core-overrides/BlockSpacing.scss
```diff
@@ -23,6 +22,20 @@
   }
 }
 
+.page-content > .wp-block-media-text {
```
mleray, line 25: Maybe we should also apply this change to `.page-content > .wp-block-group`? A lot of our patterns and pattern layouts are within a Group block 🤔

mleray: Shouldn't it be just `.page-content > .wp-block-group`?

GP-Dan-Tovbein: I applied [this code](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1108/files#diff-1791ed999879035510ea6eb81b4e9f3997399ef07c9a3919594d80ed35dd78e2R25) to fix cases like this [demo page](https://www-dev.greenpeace.org/test-neptune/3100-2/) and avoid having double empty space.

But I think if we apply `.page-content > .wp-block-group`, we will lose all the empty space between some blocks into pages, for example into a layout patterns.

GP-Dan-Tovbein: Also the new [fix](https://www-dev.greenpeace.org/test-neptune/3100-2/) fit the [UI](https://www.figma.com/file/qmQBzfSlU1TgsUPT37M0eX/Page-pattern-layouts_new?type=design&node-id=10-831&mode=dev)

mleray: Actually, this is only an issue when there's no page title, right? Maybe we should add these overrides [here](https://github.com/greenpeace/planet4-master-theme/blob/main/assets/src/scss/pages/_page.scss#L12) then?

mleray: @dantovbein what do you think about my last comment?

GP-Dan-Tovbein: Hey @mleray Here is [another PR ](https://github.com/greenpeace/planet4-master-theme/pull/2147)

assets/src/styles/blocks/core-overrides/BlockSpacing.scss
```diff
@@ -23,6 +21,21 @@
...
+  margin-top: $sp-1;
+  margin-bottom: $sp-2;
+
+  @include large-and-up {
+    margin-top: $sp-2;
+    margin-bottom: $sp-4;
+  }
+
+  &:first-child {
```
mleray, line 34: Do we still need this, even with the new PR? If not then we probably need no changes in this file 🤔

GP-Dan-Tovbein: I think we should keep this change. Here is [another demopage ](https://www-dev.greenpeace.org/test-saturn/1429-2/)(which is similar of the reported). See what happen if you remove the margin top. 
What do you think?

GP-Dan-Tovbein:
Close in favor of https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/pull/1108

PR #1109, opened 2023-08-30
Fix Submenu links font family

mleray:
### Description

Reported by Quentin on Slack, and by Houssam on Jira: [PLANET-7241](https://jira.greenpeace.org/browse/PLANET-7241)
It shouldn't be the headings font, especially with the new identity

### Testing

Make sure it still looks good both with and without the new identity styles!


lithrel (approved):
:+1:

PR #1114, opened 2023-09-07
Implement new planet4 block pattern class names

GP-Dan-Tovbein:
# Description
Modify the way of rendering block templates. This will allow us to get the proper class name assigned to any `planet4 block pattern` and finally then style it instead of making tricky implementations to `groups`.

This new implementation will help us to solve more easier these different tickets (and among others!):
- [PLANET-7226](https://jira.greenpeace.org/browse/PLANET-7226)
- [PLANET-7228](https://jira.greenpeace.org/browse/PLANET-7228)
- [PLANET-7230](https://jira.greenpeace.org/browse/PLANET-7230)

## Testing
Just navigate to [this demo page](https://www-dev.greenpeace.org/test-mars/campaign-demo-page/), open the console and run `document.querySelector(".planet4-layout-pattern")` OR `document.querySelector(".page-headers-pattern")` OR `document.querySelector(".planet4-pattern")`.

### DIFF
--- a/assets/src/block-templates/register.js
+++ b/assets/src/block-templates/register.js
@@ -19,7 +19,13 @@ const setSupport = metadata => {
 };
 
 export const registerBlockTemplates = blockTemplates => {
-  const templates = blockTemplates || templateList;
+  const templates = blockTemplates || Object.keys(templateList).map(type => templateList[type].map(template => ({
+    ...template,
+    metadata: {
+      ...template.metadata,
+      type,
+    },
+  }))).flat();
   const postType = getCurrentPostType();
 
   templates.forEach(blockTemplate => {
@@ -34,7 +40,29 @@ export const registerBlockTemplates = blockTemplates => {
 
     registerBlockType(metadata, {
       edit: edit(template, templateLock),
-      save,
+      save: props => {
+        if(props.innerBlocks.length) {
+          const className = `${props.innerBlocks[0].attributes.className || ''}`;
+          let patternClassName = '';
+
+          switch (metadata.type) {
+          case 'planet4':
+            patternClassName = 'planet4-pattern';
+            break;
+          case 'pageHeaders':
+            patternClassName = 'page-headers-pattern';
+            break;
+          case 'layouts':
+            patternClassName = 'planet4-layout-pattern';
+            break;
+          }
+
+          if(!className.includes(patternClassName)) {
+            props.innerBlocks[0].attributes.className = `${className} ${patternClassName}`;
+          }
+        }
+        return save();
+      },
     });
   });
 };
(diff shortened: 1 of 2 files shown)


planet-4:

:watch: 2023.09.07 15:15:19

assets/src/block-templates/register.js
```diff
@@ -34,7 +40,29 @@ export const registerBlockTemplates = blockTemplates => {
 
     registerBlockType(metadata, {
       edit: edit(template, templateLock),
-      save,
+      save: props => {
+        if(props.innerBlocks.length) {
+          const className = `${props.innerBlocks[0].attributes.className || ''}`;
+          let patternClassName = '';
+
+          switch (metadata.type) {
```
lithrel, line 48: Instead of a switch on type your could declare/fix the `category` property in `block.json` and use that category in a classname.

assets/src/block-templates/register.js
```diff
@@ -34,7 +40,29 @@ export const registerBlockTemplates = blockTemplates => {
...
+          case 'pageHeaders':
+            patternClassName = 'page-headers-pattern';
+            break;
+          case 'layouts':
+            patternClassName = 'planet4-layout-pattern';
+            break;
+          }
+
+          if(!className.includes(patternClassName)) {
+            props.innerBlocks[0].attributes.className = `${className} ${patternClassName}`;
```
lithrel, line 61: a little trim on this string to avoid spaces ?

lithrel (commented):
A few remarks:
- one issue we might have with fixed class name is that templates are statically generated on usage, and existing templates usage are not modified afterwards even if we update their declaration; we would end up with potentially different structures and the same classname.
- a hardcoded classname in the template would not have that issue as we could change the classname without impacting existing usage of those templates.
- Layouts don't actually need a group container to exist, they are only delimited and recognized in the editor by their html tag comment (ie `<!-- wp:planet4-block-templates/campaign -->`), they can be composed by multiple blocks without container and so could have no place to add a class name. I don't think layout should need a class :thinking:

assets/src/block-templates/register.js
GP-Dan-Tovbein: Yeah! That was the initial idea, but in that case we'll need to update all of `block.json` files, not only the one related to campaigns.

assets/src/block-templates/register.js
GP-Dan-Tovbein: I don't see necessary in this case. Why? Since the `${className}` is the default class (as `block`) and you we are re-assigning that. In this case the new `className` would be `block planet4-layout-pattern` or  would be `block planet4-pattern`.

assets/src/block-templates/register.js
lithrel: right, but I don't think it would be an issue since those declarations don't appear in the editor block list (cf l.16), nothing should change visually in the editor ?

GP-Dan-Tovbein:
Closed as a part of a proof of concept implementation

PR #1140, opened 2023-10-18
Add missing attribute to SlideBackground

mleray:
### Description

This fixes some console warnings. The new `decoding` attribute [was added with WordPress 6.1](https://www.wpexplorer.com/remove-async-decoding-wordpress-images/). Hopefully this also solves some hydration errors that we see on live sites.

### Testing

~For some reason, it seems the `decoding` attribute is not added on test instances...~ This attribute is only added when we use hydration, so with old blocks that still use `data-render` it won't be present. Code has been updated to reflect that! 

On the test instance, here are the 2 examples you can use for testing:
- [new block using hydration](https://www-dev.greenpeace.org/test-phoebe/test-hydrated-carouselheader/), should have the new attribute
- [old block not using hydration](https://www-dev.greenpeace.org/test-phoebe/), should **not** have the new attribute

You can also test this on local, on `main` branch on a page with a CarouselHeader you should see the following console warning:

```
Warning: Extra attributes from the server: decoding
```

But it should be gone once you switch to this branch.

### DIFF
--- a/assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
+++ b/assets/src/blocks/CarouselHeader/CarouselHeaderFrontend.js
@@ -19,7 +19,7 @@ const usePageLoaded = () => {
   return pageLoaded;
 };
 
-export const CarouselHeaderFrontend = ({slides, carousel_autoplay, className}) => {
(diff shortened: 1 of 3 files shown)


lithrel (dismissed):
:+1:

GP-Dan-Tovbein (approved):
Code looks good to me! 
Interesting feature

PR #1143, opened 2023-10-25
 Fix hydration errors and duplicated blocks

mleray:
### Description

We need to make sure that blocks without the `data-hydrate` attribute are rendered correctly. Before this fix we would sometimes render blocks with `data-render` inside of a `data-hydrate` attribute which would result in duplicated blocks.

I've added a commit for some old blocks not rendering in the frontend, which might solve [PLANET-7202](https://jira.greenpeace.org/browse/PLANET-7202) for the Gallery block for example

### Testing

All hydration-related console errors and warnings should be gone. If you have a page with previously duplicated blocks (for example on local I had it on the `Explore` page) it should no longer be happening. All blocks (both with `data-hydrate` and `data-render`) should be rendered properly.

### DIFF
--- a/assets/src/blocks/Covers/CoversScript.js
+++ b/assets/src/blocks/Covers/CoversScript.js
@@ -1,6 +1,16 @@
+import {createRoot} from 'react-dom/client';
 import {CoversFrontend} from './CoversFrontend';
 import {hydrateBlock} from '../../functions/hydrateBlock';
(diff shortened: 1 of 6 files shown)


planet-4:

:watch: 2023.10.25 10:03:38

comzeradd (approved):
That did fix the duplication issue for me and cleared console errors

PR #1155, opened 2023-11-14
PLANET-7326: Detect Carousel header in reusable block

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7326

> When a Carousel Header is used in a Reusable block, the dependency library Hammer is not loaded. It is because we use the `has_block()` native function that can't read into reusable blocks.

Same issue for Gallery block.

## Test

Cf. [demo page on test instance](https://www-dev.greenpeace.org/test-neptune/reusable-block-test-page/)

- create a reusable block with a Carousel Header in it
- use this reusable block in a page
- frontend should show the Carousel and not crash with a console message `Hammer is not defined`

### DIFF
--- a/classes/blocks/class-carouselheader.php
+++ b/classes/blocks/class-carouselheader.php
@@ -36,7 +36,7 @@ class CarouselHeader extends Base_Block {
 	 */
 	public function register_carouselheader_block() {
 		register_block_type(
-			'planet4-blocks/carousel-header',
+			self::get_full_block_name(),
 			[
 				'render_callback' => [ $this, 'front_end_rendered_fallback' ],
 				'attributes'      => [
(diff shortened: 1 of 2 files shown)


classes/blocks/class-carouselheader.php
```diff
@@ -115,6 +101,22 @@ public function prepare_data( $fields ): array {
 		return [];
 	}
 
+	/**
+	 * Load additional frontend assets
+	 */
+	public static function enqueue_frontend_assets() {
+		if ( BlockList::has_block( 'planet4-blocks/carousel-header' ) ) {
```
mleray, line 108: The Gallery block has the same issue

lithrel: Thanks :) fixed too  :wrench:

lithrel: Ah yes good point :eyes:

mleray: I see for the Carousel Header we should also use it to register the block, the Gallery block already does that :)

mleray (approved):
Looks good to me 👍

PR #1163, opened 2023-11-28
PLANET-7334: Fix HappyPoint height on mobile with Recaptcha

lithrel:
Ref: https://jira.greenpeace.org/browse/PLANET-7334

A Happy Point containing an embed Hubspot form with Recaptcha doesn't always fit the entire form, the submit button is pushed down when Recaptcha appears after page loading; the wrapper div can't resize for that.
Passing its height to `auto` on mobile allows it to resize according to the change of the content.

### DIFF
--- a/assets/src/styles/blocks/Happypoint.scss
+++ b/assets/src/styles/blocks/Happypoint.scss
@@ -13,7 +13,8 @@
   }
 
   @include mobile-only {
-    height: 544px;
+    height: auto;
+    min-height: 544px;
   }
 
   @include medium-and-up {


planet-4:

:new_moon: [atlas](https://www-dev.greenpeace.org/test-atlas) | [admin](https://www-dev.greenpeace.org/test-atlas/wp-admin/) | [blocks report](https://www-dev.greenpeace.org/test-atlas/wp-admin/?page=plugin_blocks_report) | [CircleCI](https://app.circleci.com/pipelines/github/greenpeace/planet4-test-atlas) | [composer-local.json](https://github.com/greenpeace/planet4-test-atlas/blob/main/composer-local.json)

:watch: 2023.11.29 09:32:51

mleray (changes requested):
I think that we should add a min-height, because right no if there's no content apart from the image it has a height of 0 on mobile 😬 See [test instance homepage](https://www-dev.greenpeace.org/test-atlas/) for example

lithrel:
@mleray  Good point, added and redeployed :+1:

mleray (approved):
👍

PR #1166, opened 2023-12-01
Fix double declaration of pre-publish checklist

lithrel:
Remove duplicate call of block validation system
`blockEditorValidation` is already called l.40, triggering the console warning `Plugin "pre-publish-checklist" is already registered.`.


sagarsdeshmukh (approved):
👍

PR #1168, opened 2023-12-07
Split Cookies Block files

mleray:
### Description

This is for better performance since now the files will only be loaded when they are needed. This should be merged before we move the block to the theme for [PLANET-7302](https://jira.greenpeace.org/browse/PLANET-7302)

### Testing

Both new and existing Cookies blocks should behave as expected in the editor and in the frontend. There should be no errors or warnings in the console.


sagarsdeshmukh (approved):
👍

PR #1181, opened 2024-01-11
fix: change gsutil into gcloud

foppepieters:


foppepieters:
@comzeradd I rebased, but the linting is still whining about my commit message:

✖ header must be sentence-case [header-case]
✖ body may not be empty [body-empty]

comzeradd (approved):

PR #1194, opened 2024-03-11
PLANET-7444 Fix left/right aligned images in all post types

mleray:
### Description

See [PLANET-7444](https://jira.greenpeace.org/browse/PLANET-7444)
This was only working in posts.

**Related PR:** https://github.com/greenpeace/planet4-master-theme/pull/2226

### Testing

I've added the various Image alignments on the following:
- [Page](https://www-dev.greenpeace.org/test-pandora/page-with-images/)
- [Post](https://www-dev.greenpeace.org/test-pandora/story/1273/post-with-images/)
- [Action](https://www-dev.greenpeace.org/test-pandora/petitions/action-with-images/)

### DIFF
--- a/assets/src/styles/blocks/core-overrides/Image.scss
+++ b/assets/src/styles/blocks/core-overrides/Image.scss
@@ -63,14 +63,21 @@
     position: relative;
   }
 
-  .alignleft {
-    margin: 0 $sp-1 $sp-1 0;
-    padding-top: 6px;
+  @include small-and-up {
+    &.alignleft {
+      float: left;
+      margin-inline-end: $sp-1;
+    }
+
+    &.alignright {
+      float: right;
+      margin-inline-start: $sp-1;
+    }
   }
 
-  .alignright {
-    margin: 0 0 $sp-1 $sp-1;
-    padding-top: 6px;
+  &.aligncenter {
+    margin-left: auto;
+    margin-right: auto;
   }
 
   @include medium-and-up {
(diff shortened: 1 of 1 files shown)


assets/src/styles/blocks/core-overrides/Image.scss
```diff
@@ -63,14 +63,21 @@
...
-  .alignleft {
-    margin: 0 $sp-1 $sp-1 0;
-    padding-top: 6px;
+  @include small-and-up {
+    &.alignleft {
+      float: left;
+      margin-inline-end: $sp-1;
+    }
+
+    &.alignright {
```
lithrel, line 72: do you know why we do this only for small-and-up ? the alignright ends up on the left on smaller viewport
Screenshot from 2024-03-11 16-25-12

mleray: I'm not sure actually, I guessed it was because on mobile it didn't matter since the viewport is smaller... But now that you ask I realise that on posts the viewport is smaller than for other post types, so maybe it makes sense to add it to all screen sizes?

lithrel: According to blame, there was an issue with image captions and float on mobile maybe https://github.com/greenpeace/planet4-master-theme/pull/537 :thinking:

mleray: I've updated the code so that it gets applied to all screen sizes, seems to work fine to me 🤔

lithrel (approved):
:+1:

PR #1197, opened 2024-04-04
PLANET-7304 Move ENForm block to master theme

mleray:
### Description

See [PLANET-7304](https://jira.greenpeace.org/browse/PLANET-7304)
We need this check to make sure we don't load assets twice, see comment [here](https://github.com/greenpeace/planet4-master-theme/pull/2190#pullrequestreview-1979216923)

**Related PR:** https://github.com/greenpeace/planet4-master-theme/pull/2247

### DIFF
--- a/classes/class-loader.php
+++ b/classes/class-loader.php
@@ -121,12 +121,12 @@ final class Loader {
 			Controllers\Menu\Reusable_Blocks_Controller::class,
 			Controllers\Menu\Archive_Import::class,
 			Controllers\Menu\Postmeta_Check_Controller::class,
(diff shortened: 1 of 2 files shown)


classes/class-loader.php
```diff
@@ -9,8 +9,6 @@
 namespace P4GBKS;
 
 use P4\MasterTheme\Features;
-use P4\MasterTheme\Features\Dev\ThemeEditor;
```
mleray, line 12: These were not used

lithrel (approved):
:+1:

PR #1235, opened 2024-09-04
PLANET-7534 Allow Take Action Boxout in all post types

mleray:
### Description

See [PLANET-7534](https://jira.greenpeace.org/browse/PLANET-7534)

More specifically, it is now allowed in Pages, Posts and Actions.

### Testing

Make sure to test both sticky and non-sticky versions of the block!
- [Post with TAB](https://www-dev.greenpeace.org/test-umbriel/story/1231/post-with-tab/)
- [Page with TAB](https://www-dev.greenpeace.org/test-umbriel/page-with-tab/)
- [Action with TAB](https://www-dev.greenpeace.org/test-umbriel/petitions/action-with-tab/)


mardelnet (approved):
Good job, @mleray 
I see the block was not added to the Campaign post type, but that is not requested in the ticket requirements, so I suppose it's fine as it is.

mleray:
> Good job, @mleray I see the block was not added to the Campaign post type, but that is not requested in the ticket requirements, so I suppose it's fine as it is.

Thanks for the review @mardelnet! Indeed it is not mentioned in the ticket, because we are planning to get rid of Campaign pages and use Actions instead 🙂 However I'm waiting for Houssam to give me his feedback, because the boxout looks really wide in larger screens Pages/Actions so I might need to make additional changes.

PR #1252, opened 2024-10-09
Remove Stretched Link, Quick Links, and Deep Dive templates

mardelnet:
**DESCRIPTION:**
This PR removes the Stretched Link block variation, the Quick Links template, and the Deep Dive template as part of our effort to migrate the P4 plugin codebase into the master theme, since that code already exists in the theme. Additionally, the PHP files within the `languages` folder are ignored from PHPCS analysis.
<hr>

**RELATED PR:**
- https://github.com/greenpeace/planet4-master-theme/pull/2400


planet-4:

:watch: 2024.10.09 08:26:05

mleray:
This fix is fine, but what I'm wondering is why the parent div as the `group-stretched-link` class? It shouldn't be there and that is what causes problems 🤔 If I remove it it works as expected. Could it be the editor added it for some reason? It's not there in our template (I've added another one of our templates on your test page for you to see the difference)

Screenshot 2024-10-09 at 11 09 28

mleray (commented):
I'm wondering if we should also remove the [style registration](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/editorIndex.js#L34-L44) from the plugin 🤔 
It is used in two templates (Quick Links and Deep Dive), both of which have been moved to the theme already but are still in the plugin too.

mardelnet:
> I'm wondering if we should also remove the [style registration](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/editorIndex.js#L34-L44) from the plugin 🤔 It is used in two templates (`Quick Links` and `Deep Dive`), both of which have been moved to the theme already but are still in the plugin too.

If so, why don't we remove the templates completely?

mardelnet:
> I'm wondering if we should also remove the [style registration](https://github.com/greenpeace/planet4-plugin-gutenberg-blocks/blob/main/assets/src/editorIndex.js#L34-L44) from the plugin 🤔 It is used in two templates (`Quick Links` and `Deep Dive`), both of which have been moved to the theme already but are still in the plugin too.

Done! Would you mind rechecking it, @mleray?

assets/src/block-templates/deep-dive/template.js
```diff
@@ -3,7 +3,7 @@ import mainThemeUrl from '../main-theme-url';
 const {__} = wp.i18n;
 
 const topic = ['core/column', {}, [
-  ['core/group', {className: 'group-stretched-link'}, [
```
mleray, line 6: This isn't what I meant, I think that we can fully remove the templates from the plugin since they have been moved to the theme. 
If we don't want to remove them just yet, this class should stay.

mardelnet: Change reverted!

mleray: @mardelnet you don't think we can remove them from the plugin even though they are in the theme already?

mardelnet: Just to be clear, are you suggesting completely removing from the plugin the templates Quick Links and Deep Dive?

mleray: Yes, exactly 🙂 But it can also be done in a separate PR since it's not really related to the issue, as you prefer!

mardelnet: The templates were removed, @mleray

mleray:
> If so, why don't we remove the templates completely?

@mardelnet sorry I completely missed this message, but yeah indeed, we ended up understanding each other 😉

mleray (approved):
Looks good to me 👍 All seems to work as expected on the test instance.

